PR #117, opened 2026-04-24
Observability fixes + raise WMS/Maps dimension cap to 8000

mrauhala:
## Summary

Three small, independent-but-related fixes that came out of debugging 400s on the production WMS endpoint.

- **`fix(observability)`** — The Alloy pipeline used `slice .status 0 1` to derive `status_class`, but `tracing-subscriber` emits `status` as a JSON number and Alloy keeps numeric values as `float64`. `slice` silently failed, `status_class` was never set, and the error-rate-from-logs panel was permanently empty. Switched to `div (int .status) 100`. Also added a "Cache Hit Rate Over Time" time-series panel below the existing gauges, and removed a stale `docker-compose.yml` (the observability stack lives in `compose.yml`).

- **`feat(server)` — log error reason on 4xx/5xx** — Request logs only carried the status code; the reason was stuck in the response body (XML for WMS, JSON for Maps/Tiles) and couldn't be grepped in Loki. Added `ds_core::error::ErrorReason` as a lightweight carrier attached to response extensions. Each API's `IntoResponse` impl inserts one; `request_logging_middleware` reads it and emits an `error` field alongside the status. `ds-core` stays framework-free. Scoped to WMS/Maps/Tiles for now (api-edr/api-features do error mapping inline across many handlers and would need a broader refactor).

- **`feat(api-wms,api-maps)` — raise `MAX_MAP_DIMENSION` to 8192** — `engine-geotiff` already caps at 64M source pixels, but the API caps were 4096/dim / 16M total, which rejected legitimate desktop-sized radar mosaic requests (e.g. 5160×1881) with a 400. Aligned API caps with the engine so anything passing API validation renders successfully.

## Test plan

- [ ] `cargo test -p api-wms -p api-maps -p api-tiles -p server` passes (done locally — 42 tests in api-maps integration pass after bumping `width=5000` → `width=9000` in the "too large" test)
- [ ] `cargo clippy -p ds-core -p api-wms -p api-maps -p api-tiles -p server --tests -- -D warnings` clean (done locally)
- [ ] After deploy: issue a GetMap with an unknown layer and verify the request log line contains `"error":"LayerNotDefined: Layer '...' does not exist"`
- [ ] After deploy: issue a GetMap with 5160×1881 and verify it returns a PNG (previously 400)
- [ ] After deploy (+ Alloy restart): Grafana → "Error Rate by Status Class (from logs)" panel shows non-zero 4xx/5xx when a 400 occurs


mrauhala:
Closing manually: GitHub's PR backend got into a stuck state when this was merged, leaving `state=OPEN, mergeCommit=null, mergeable=CONFLICTING` even though branch HEAD `525016d` is an ancestor of main HEAD `dc673ca` (merge commit dated 2026-05-04). All four commits from this PR are on `main`, and the Docker image was rebuilt via the weekly scheduled CI run on 2026-05-10 (`ghcr.io/mrauhala/meteocore:main` is current).

Issue #127, opened 2026-05-11
Vector tiles: remaining simplification and engine pushdown work


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** MVT routes and encoder already exist. Any remaining simplification/property/pushdown work must be separated from the obsolete no-MVT premise.

**Next step / remaining scope:** Audit the remaining simplification, property-selection and engine-pushdown requirements against the existing MVT routes and encoder.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Motivation

\`api-features\` currently only emits GeoJSON. At city/country zoom levels, GeoJSON responses for \`engine-geojson\` and \`engine-postgis\` get huge — unusable in browsers. Mapbox Vector Tiles (MVT) is the standard fix and the universal input format for MapLibre/Mapbox/OpenLayers/deck.gl/Leaflet.VectorGrid.

This is the obvious counterpart to the existing raster Tiles API, and pairs perfectly with the built-in preview UI.

## Scope

- MVT output for \`api-features\` (and/or a new \`/tiles/{collection}/{tms}/{z}/{x}/{y}.mvt\` route consistent with the raster tiles pattern)
- Server-side simplification by zoom level (Douglas-Peucker)
- Attribute filtering — only emit the properties relevant to styling, not every column
- Reuse existing TileMatrixSet machinery from \`api-tiles\`
- WebMercator (\`WebMercatorQuad\`) and CRS84 (\`WorldCRS84Quad\`) at minimum

## Considerations

- Rust crates: \`mvt\` and \`geozero\` are both viable — survey before choosing
- PostGIS already has \`ST_AsMVT\` server-side, which is dramatically faster than client-side encoding — for \`engine-postgis\` we should delegate to the database
- GeoJSON engine: in-memory tile generation with quad-tree style spatial indexing for fast bbox queries
- Caching: MVT bytes are highly cacheable; reuse the existing render-cache pattern from \`engine-geotiff\`
- Tile coordinate → bbox conversion already exists for raster tiles — reuse it

## Open questions

- New trait method on \`FeatureEngine\` (\`get_vector_tile\`), or compose existing \`get_features(bbox=...)\` + an MVT encoder?
[...]

Issue #126, opened 2026-05-11
Built-in MapLibre preview: remaining basemap and UX work


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** Bundled MapLibre preview exists. Any unfinished basemap/UX phase needs its own current acceptance criteria.

**Next step / remaining scope:** Identify concrete unfinished basemap and preview UX acceptance criteria against the bundled viewer.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Motivation

Right now demonstrating MeteoCore requires the audience to bring a GIS client (QGIS, ArcGIS, an existing webmap). A built-in viewer would 10x demo-ability and onboarding ease — the server becomes self-explanatory the moment you hit its root URL.

This is the highest-leverage adoption multiplier in the backlog: small effort, outsized impact on first impressions and internal demos.

## Scope

- New route \`GET /preview\` (and optionally \`GET /preview/{collection_id}\`)
- Single-page MapLibre GL JS viewer, served as static assets embedded in the binary (\`include_str!\`/\`include_bytes!\`)
- Auto-discovers collections from the existing registry:
  - For \`maps\`/\`tiles\` collections → render as raster tiles
  - For \`wms\` collections → render as WMS source
  - For \`features\` collections → render as GeoJSON source
  - For \`edr\` collections without map APIs → show a marker per location (CSV) or a click-to-query handler
- Sidebar with layer toggles, opacity sliders, parameter switcher (for multi-parameter collections), time slider (for time-aware collections)
- Optional URL-state encoding (\`?layers=a,b&t=...&bbox=...\`) for shareable views
- Style switcher for WMS collections that expose multiple styles via \`style_bundle\`

## Non-goals (v1)

- Drawing tools / editing
- EDR area/trajectory queries from UI
- User authentication
- Saved views / bookmarks

## Considerations

- Bundle size: MapLibre GL JS is ~800 KB minified — acceptable. Vendor it; don't pull from CDN at runtime (offline deployments).
- The viewer should work *zero-config* — read \`/collections\` and decide what to show based on declared \`apis\` and engine capabilities.
- Time-aware collections expose temporal extent via \`/collections/{id}\` — viewer should auto-detect and render a time slider only when applicable.
[...]

PR #132, opened 2026-05-11
feat(preview): manifest.json — unified discovery for the UI (#126 Phase 1)

mrauhala:
## Summary

Phase 1 of #126. Stacked on top of #131 (#127 Phase 3).

Adds `GET /preview/manifest.json` — one aggregated discovery endpoint that the upcoming MapLibre preview SPA (Phase 2+) consumes as its only source of truth. Solves the discovery problem flagged in the plan's devil's advocate section: without this, the browser would have to fan out to five `/{api}/collections` probes per page load and reconcile per-API JSON drift client-side.

## Output shape

```json
{
  "collections": [
    {
      "id": "fmi-air",
      "title": "...",
      "description": "...",
      "apis": ["edr", "features", "tiles"],
      "spatial_extent": [west, south, east, north],
      "temporal_extent": {
        "start": "2024-01-01T00:00:00+00:00",
        "end": "2024-01-02T00:00:00+00:00",
        "values": ["..."],
        "truncated": false,
        "total_values": 24
      },
      "tiles": {
        "vector": { "tile_matrix_sets": ["..."], "url_template": "...{?f=mvt}" },
        "raster": { "tile_matrix_sets": ["..."], "url_template": "...", "default_style": "default" }
      },
      "styles": [{ "id": "default", "title": "Default" }]
    }
  ],
  "pagination": { "offset": 0, "limit": 100, "total": 5, "returned": 5, "next": null }
}
```

## Design choices

- **Temporal extent normalization** — EDR's `values[]` and Maps/Tiles' `interval[]` collapse server-side into the same shape, so the UI's time slider never has to care which API the collection came from.
- **`values[]` cap at 1000** + `truncated=true` flag — keeps an obs collection with 500k timestamps from blowing the manifest size.
- **Tile descriptors emitted only when wired** — no dead layer toggles for collections that don't actually have raster or vector tile sources registered.
- **Pagination** — `?offset` / `?limit` (default 100, max 1000); pagination response carries `total`, `returned`, `next` for client-side concatenation. A 1000-collection deployment doesn't OOM the browser.
- **Soft 1 MB body warning** logs without rejecting — operator-facing signal, never a 4xx.

## Test plan

- [x] 7 unit tests in `crates/server/src/preview.rs`: empty state, EDR aggregation, value truncation at 1000, vector+raster tile descriptor emission, pagination skip+limit+next, and that a collection registered in two APIs deduplicates to one entry.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean.
- [x] `cargo test --workspace` green.
- [ ] Manual: `cargo run -p server` → `curl -s http://localhost:8080/preview/manifest.json | jq` — verify in Phase 2 once asset embedding lands and end-to-end consumers exist.

## What's next

- **Phase 2** — static-asset embedding (`rust-embed`, vendored MapLibre, `/preview` HTML).
- **Phase 3+** — JS that consumes this manifest and renders layers.

Refs #126.

### DIFF
--- a/crates/server/src/preview.rs
+++ b/crates/server/src/preview.rs
@@ -64,7 +64,16 @@ pub async fn manifest_handler(
 
     // Size guard — log only, never reject; pagination defaults already prevent
     // pathological responses, and the soft warn lets operators see drift early.
-    let body = serde_json::to_vec(&manifest).unwrap_or_else(|_| b"{}".to_vec());
+    //
+    // `serde_json::to_vec` on a `serde_json::Value` essentially never fails
+    // (the input is already validated JSON), but if it does we log and fall
+    // back to `{}` rather than 500-ing the request — preview UIs can keep
+    // rendering whatever cached snapshot they hold while operators see the
+    // error in logs.
+    let body = serde_json::to_vec(&manifest).unwrap_or_else(|e| {
+        tracing::error!(error = %e, "preview manifest JSON serialisation failed");
+        b"{}".to_vec()
+    });
     if body.len() > MANIFEST_WARN_BYTES {
         tracing::warn!(
             "preview manifest body is {} bytes ({} collections, offset={}, limit={}); \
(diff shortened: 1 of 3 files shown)


crates/server/src/preview.rs
```diff
@@ -0,0 +1,1015 @@
...
+            parameter: "reflectivity".into(),
+            unit: "dBZ".into(),
+        });
+        let feature_engine: Arc<dyn FeatureEngine> = Arc::new(PointFeatureMock {
+            extent: Some([20.0, 60.0, 30.0, 70.0]),
+        });
+        tiles.map_engines.insert("radar".into(), raster);
+        tiles
+            .collections
+            .insert("radar".into(), config("radar", &["tiles"]));
```
mrauhala, line 749: Verified against `chrono = "0.4"` (currently 0.4.44):

- `DateTime::<Utc>::from_timestamp(secs, nsecs)` compiles **clean with `-D warnings`** — not deprecated.
- `DateTime::from_timestamp_opt` does **not exist** on `DateTime`: `rustc` itself suggests `from_timestamp` as the correct associated function name.

The deprecation introduced in chrono 0.4.30 applies to `NaiveDateTime::from_timestamp` (and the panicking `TimeZone::timestamp`), not to `DateTime::from_timestamp` — that one was *added as* the non-panicking replacement and returns `Option`.

Leaving the call as-is.

PR #136, opened 2026-05-11
fix(mvt): content-derived ETag + tile-buffer polygon clipping

mrauhala:
## Summary

Two distinct bugs surfaced from the same symptom (preview shows nothing
at low zoom, or warns about \"Geometry exceeds allowed extent\" at high
zoom):

- **ETag was key-derived.** \`VectorTileKey::etag()\` hashed the cache key,
  not the encoded bytes. A browser that cached a stale tile (e.g. after
  a server-side bug produced empty tiles) would keep getting 304
  forever — the etag would never change, even after the server fix.
  Moved into a content-derived \`CachedTile { bytes, etag }\` so different
  bytes produce different etags.
- **Polygon geometry exceeded MapLibre's allowed extent.** The encoder
  projected every vertex of an arbitrarily large polygon into a small
  tile, producing coordinates millions of units beyond the 4096 grid.
  MapLibre rejects features whose post-scale coords exceed \`i16\` range.
  Added Sutherland-Hodgman clipping against a buffered tile rectangle
  (256 MVT units of slop, matching MapLibre's source-layer buffer
  convention). Polygons that fall fully outside are dropped; polygons
  that cross the boundary are trimmed.

Verified end-to-end in the browser: at z=14 over Helsinki, 16/16 tiles
parse, 16/16 features render, no \"exceeds extent\" warnings.

## Test plan
- [x] \`cargo test -p ds-mvt\` (24 passed — includes new clipping tests)
- [x] \`cargo test -p api-tiles\` (53 passed)
- [x] \`cargo clippy -p ds-mvt -p api-tiles --all-targets -- -D warnings\`
- [x] Manual: hard-refresh browser, zoom in on municipalities to z=14

### DIFF
--- a/crates/api-tiles/src/handlers.rs
+++ b/crates/api-tiles/src/handlers.rs
@@ -785,10 +785,10 @@ async fn render_vector_tile(
     // would let stale browser caches survive a server fix indefinitely.
     if let Some(cached) = state.vector_tile_cache.get(&cache_key) {
         if let Some(ref inm) = if_none_match {
-            if ds_render::etag_matches(inm, &cached.etag) {
+            if ds_render::etag_matches(inm, cached.etag()) {
                 return Ok(axum::response::Response::builder()
                     .status(StatusCode::NOT_MODIFIED)
-                    .header(header::ETAG, &cached.etag)
+                    .header(header::ETAG, cached.etag())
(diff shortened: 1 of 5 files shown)


mrauhala:
Raster ETag parity follow-up filed as #145.

PR #146, opened 2026-05-12
ci: speed up Docker build (~248s → ~60-90s on cache-hit)

mrauhala:
## Summary

Two cleanly-scoped Dockerfile changes that together remove a real correctness bug and modestly speed up Docker builds.

## Changes

1. **Replace the hand-enumerated stub-lib trick with `cargo-chef`.**
   The old `Dockerfile` hand-enumerated 13 workspace crates in a `for dir in core engine-csv ...` loop and silently omitted `engine-postgis` after it was added. `cargo chef prepare` now derives a stable `recipe.json` from `Cargo.toml` + `Cargo.lock`; `cargo chef cook` compiles the dep tree into a Docker layer that GHA's `type=gha,mode=max` cache exports correctly. The layer only invalidates on actual dependency changes — adding a new workspace member can't silently break the cache.

2. **Expand `.dockerignore` to shrink the build context.**
   Added `testdata/`, `fuzz/`, `varnish/`, `.github/`, `*.md`, `README*`, and `recipe.json` to the existing `target/` + `.git/` exclusions. The planner's `COPY . .` and the builder's source `COPY . .` both run noticeably faster.

3. **Add `recipe.json` to `.gitignore`.** Belt-and-braces with the `.dockerignore` exclusion so a locally-generated planner output can't accidentally be committed and override the in-container output.

## What I tried that didn't work (documented in the Dockerfile)

- **BuildKit `--mount=type=cache` for `target/` and the Cargo registry.** Looked great on paper but `type=gha` cache backend doesn't export cache-mount contents — only image layers — so on ephemeral hosted runners every build hit an empty mount. Dropped.
- **cmake from bookworm-backports (3.31).** Backports has cmake 3.31 (libaec-sys needs ≥3.26, bookworm main ships 3.25), but apt's solver couldn't resolve `libjsoncpp25` across the main/backports boundary on the current `rust:1.94-slim`. Reverted to the original `pip install --break-system-packages cmake` approach. Documented inline so a future contributor knows the backports route was tried and why it didn't work.

## Realistic timing impact

Modest — the cache-mount path that would have produced the dramatic speedup is unavailable on hosted runners. The genuine wins on this PR:

- **Correctness fix**: `engine-postgis` deps now get pre-compiled instead of being rebuilt every time (silently slow before).
- **Smaller build context**: the `COPY . .` operations are faster.
- **Robustness**: adding a new workspace member can't break the dep-cache silently.

## Test plan

- [ ] `Docker Build (PR)` runs successfully (CI verifies).
- [ ] Resulting image starts and serves `/health` (smoke test on merge).


mrauhala:
Good catches. Pushed `58060f8` addressing all three:

1. **Build failure (cmake/libjsoncpp25)**: dropped `--no-install-recommends` from the backports `cmake` install only — apt's solver needs the Recommends headroom to resolve transitive deps across the main/backports boundary.

2. **Cache mounts were no-op on hosted runners** — you're right, `type=gha` only exports image layers, not `--mount=type=cache` contents. Dropped them.

3. **Stub-removal regressed dep caching** — also correct. Replaced the hand-enumerated stub trick with `cargo-chef`, which derives a stable `recipe.json` from `Cargo.toml`/`Cargo.lock` only. The cook layer caches across source-only changes and picks up new workspace members automatically (so adding a new crate can't silently break the cache the way the old loop did when `engine-postgis` was added).

PR body's "60–90s" was wrong about the hosted-CI path. Realistic target now: source-only changes ~3 min → ~1 min (workspace crate rebuild on top of cached deps); manifest changes stay at ~3 min.

PR #169, opened 2026-05-14
fix(wms,maps,tiles): emit image/png Content-Type for empty-tile fallback

### LINKED ISSUE #162: WMS/Maps/Tiles: empty-tile response has Content-Type mismatch when format != PNG
## Bug

When a GetMap / map / tile request hits an all-nodata tile, the handlers short-circuit and return a transparent PNG without going through the normal encode path. But the response \`Content-Type\` header has already been set from the *requested* format. If the client asked for \`image/jpeg\` or \`image/webp\` and the tile happens to be empty, they get a PNG body with a JPEG/WebP Content-Type header — breaks decoders that trust the Content-Type and exposes a clear server bug.

## Affected call sites

- **WMS** — \`crates/api-wms/src/handlers.rs:243-251\`
  \`\`\`rust
  Ok(None) => {
      // Empty tile: return transparent PNG without caching
      let rgba = vec![0u8; (params.width * params.height * 4) as usize];
      let png = ds_render::encode_png(&rgba, params.width, params.height).map_err(...)?;
      (bytes::Bytes::from(png), \"EMPTY\")
  }
  \`\`\`
  \`content_type\` is computed from \`params.format.content_type()\` *before* the match and is used unconditionally in the final response builder.

- **Maps** — \`crates/api-maps/src/handlers.rs:831-837\` — same shape.

- **Tiles** — \`crates/api-tiles/src/handlers.rs:1171-1178\` uses the pre-generated \`EMPTY_TILE_PNG\` but \`content_type\` is still \`validated.format.content_type()\`.

## Fix

The cleanest options:

1. **Encode in the requested format**: drop the PNG short-circuit, encode an all-transparent RGBA buffer through the same path the populated tile takes. WebP has good RLE behaviour on uniform buffers; JPEG has no alpha (would need to fall back to PNG). Adds work to the empty-tile fast path.

2. **Switch the Content-Type when emitting a PNG**: when falling back to the pre-generated PNG, set the Content-Type to \`image/png\` instead of the requested type. Cheapest; tells the client the truth.

3. **Reject the format conversion on empty tiles**: return 415 / error. Worst UX.

I'd go with (2) — it's the smallest change and consistent with what the server actually shipped. Document the behaviour (\`X-Cache: EMPTY\` always implies \`Content-Type: image/png\`).

## Related

Discovered while auditing the README WMS/Maps/Tiles sections (#159).

mrauhala:
## Summary

Closes #162.

When a GetMap / map / tile request hit an all-nodata tile, the three render handlers short-circuited and returned a transparent PNG without going through the format-aware encoder — but the \`Content-Type\` header had already been computed from the *requested* \`FORMAT\`/\`f=\`. A client asking for \`image/jpeg\` or \`image/webp\` who landed on an empty tile got PNG bytes with \`Content-Type: image/jpeg\` (or \`.../webp\`), breaking decoders that trust the header.

### Fix shape — same in all three handlers

Track \`Content-Type\` alongside the body in the match arm so the fast paths can emit \`image/png\` truthfully:

\`\`\`rust
let (image_bytes, x_cache, response_content_type) = match ... {
    Ok(Some(bytes)) => (..., \"MISS\",  content_type /* requested */),
    Ok(None)        => (..., \"EMPTY\", \"image/png\"),
    Err(e)          => (..., \"ERROR\", \"image/png\"),  // WMS only
};
\`\`\`

### Why option 2 (issue #162 enumerated three)

This is the issue author's recommended option: smallest change, truthful header, fast path preserved. Option 1 (re-encode into the requested format) would erase the empty-tile fast-path benefit, and JPEG can't carry alpha anyway so its \"empty\" output is degenerate either way.

### Affected files

- \`crates/api-wms/src/handlers.rs\` — covers both the \`Ok(None)\` empty branch and the \`Err(e)\` error-tile branch. \`render_error_tile\` also emits PNG, so it was lying about Content-Type via the same code path.
- \`crates/api-maps/src/handlers.rs\` — \`Ok(None)\` branch only (errors return JSON 500 from \`MapsError::Internal\`).
- \`crates/api-tiles/src/handlers.rs\` — uses the pre-generated \`EMPTY_TILE_PNG\` global. Same fix shape.
- \`README.md\` — amend the Tiles \"Raster Tiles\" step 4 to mention \`Content-Type: image/png\` on EMPTY and explain the rationale.

### Trade-off documented

The README now spells out the consequence: clients with strict JPEG/WebP pipelines should check \`X-Cache: EMPTY\` and refetch only if needed. The alternative (re-encoding a known-uniform RGBA buffer on every empty tile) burns CPU for negligible benefit, and JPEG-with-alpha doesn't exist anyway.

## Test plan

- [x] \`cargo test -p api-wms -p api-maps -p api-tiles\` — all existing tests still pass
- [x] \`cargo clippy -p api-wms -p api-maps -p api-tiles\` clean
- [x] \`cargo fmt --check\` clean
- [ ] Reviewer eyeballs the three match-arm sites for symmetry
- [ ] Manual: \`curl '...REQUEST=GetMap&FORMAT=image/jpeg&BBOX=<empty-area>'\` returns \`Content-Type: image/png\` with \`X-Cache: EMPTY\`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,9 +137,11 @@ dependencies = [
  "chrono",
  "ds-core",
  "ds-render",
+ "http-body-util",
  "quick-xml",
  "serde",
  "tokio",
+ "tower",
  "tracing",
 ]
 
--- a/README.md
+++ b/README.md
@@ -960,7 +960,7 @@ Raster requests resolve a style (the `default` style for the unstyled route, or 
 1. TMS → bbox conversion (Mercator math for `WebMercatorQuad`, linear lon/lat for `WorldCRS84Quad`).
 2. ETag check / rendered-cache check.
 3. On miss: acquire render permit (timed out → 503), call `MapEngine::get_raster_tile()` with the TMS-derived `output_crs`, colorize, encode.
-4. Empty (all-nodata) tiles return a pre-generated transparent PNG with `X-Cache: EMPTY` regardless of the requested format — no engine call, no cache insert.
+4. Empty (all-nodata) tiles return a pre-generated transparent PNG with `X-Cache: EMPTY` and `Content-Type: image/png` regardless of the requested format — no engine call, no cache insert. Clients that strictly need JPEG/WebP output should check `X-Cache: EMPTY` and refetch only if needed; the alternative would be re-encoding a known-uniform buffer on every empty tile.
 
 Successful responses include `Content-Type`, `Content-Crs` (OGC URI for the TMS's CRS), `ETag`, `Cache-Control` (`max-age=86400, immutable` when `datetime` is set, `max-age=60, must-revalidate` otherwise), and `X-Cache: HIT|MISS|EMPTY`.
 
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -822,18 +822,23 @@ async fn render_map(
     .await
     .map_err(|e| MapsError::Internal(format!("Render task failed: {e}")))?;
 
-    let (image_bytes, x_cache) = match render_result {
+    // The EMPTY fast path skips the format-aware encoder and emits PNG
+    // bytes directly. Track the actual Content-Type per branch so the
+    // header never lies about the payload — previously an `f=image/jpeg`
+    // request that hit an empty tile got PNG bytes with
+    // `Content-Type: image/jpeg`, breaking decoders that trust the type.
+    let (image_bytes, x_cache, response_content_type) = match render_result {
         Ok(Some(bytes)) => {
             let image_bytes = bytes::Bytes::from(bytes);
             rendered_cache.insert(cache_key, image_bytes.clone());
-            (image_bytes, "MISS")
+            (image_bytes, "MISS", content_type)
         }
         Ok(None) => {
-            // Empty tile: return transparent PNG without caching
+            // Empty tile: return transparent PNG without caching.
             let rgba = vec![0u8; (width * height * 4) as usize];
             let png = ds_render::encode_png(&rgba, width, height)
                 .map_err(|e| MapsError::Internal(format!("Failed to encode empty tile: {e}")))?;
-            (bytes::Bytes::from(png), "EMPTY")
+            (bytes::Bytes::from(png), "EMPTY", "image/png")
         }
         Err(e) => {
             tracing::warn!("Maps render error for collection '{}': {e}", collection_id);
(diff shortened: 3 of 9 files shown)


mrauhala:
Regression tests added in 1167597 — addresses the blocking finding:

- **api-maps**: `mod get_map` gets `empty_tile_forces_png_content_type_even_when_jpeg_requested` and `...webp_requested`. New `EmptyMockMapEngine` returns `RasterTile { values: vec![None; n] }` and a `build_empty_router` helper wires it through `MapsState`.
- **api-tiles**: same pair of tests in `mod get_tile`, same `EmptyMockMapEngine` shape adapted for `TilesState`.
- **api-wms**: there was no integration test file at all — created `tests/integration.rs` from scratch with the same regression pair, plus the dev-deps (`tower`, `http-body-util`, `tokio` macros) to drive the axum router. This bootstraps future WMS integration coverage.

Each crate has two tests covering JPEG and WebP separately (WebP takes a distinct `ImageFormat` branch through the cache key). All three handlers' empty-tile fast paths are now pinned: `Content-Type: image/png`, PNG magic bytes in the body, `X-Cache: EMPTY`.

The two pre-existing concerns the review flagged are out of scope for this PR and tracked separately:
- **#170** — stale-empty 304 within the must-revalidate window (the bug #162 actually surfaced; ETag doesn't distinguish empty vs populated responses).
- **#171** — WMS/Maps allocate fresh empty PNG per request vs Tiles' pre-generated `EMPTY_TILE_PNG` global. Performance cleanup.

mrauhala:
Round-2 feedback addressed:

**Blocker — WMS error-tile regression test missing** → fixed in 548c627. Added `FailingMockMapEngine` (always returns `DataServerError::Engine`) and `build_failing_router` to `crates/api-wms/tests/integration.rs`. New test `error_tile_forces_png_content_type_even_when_jpeg_requested` asserts:
- HTTP 200 (WMS swallows engine errors into an error-tile rather than 5xx)
- `Content-Type: image/png`
- `X-Cache: ERROR`
- body starts with PNG magic bytes

api-wms now has 3 integration tests covering all three exits of the GetMap match arm.

**Non-blocker — WMS 1.3.0 §7.3.3.11 spec compliance** → filed [#172](https://github.com/mrauhala/meteocore/issues/172). The empty-tile silent format substitution is technically a violation (server returns image/png when image/jpeg was requested without ServiceException). Three options sketched; the current behaviour (option 3, accept + document) matches the reviewer's preference and is the most useful for clients. Filed as a tracked item rather than just a README paragraph.

**Minor — `EmptyMockMapEngine` duplicated across three test files** → noted; a shared test-helpers crate would dedupe. Skipping for this PR since the helpers are small and a clean extraction belongs in its own change.

Issue #29, opened 2026-04-06
ODIM HDF5 engine


mrauhala:
Native ODIM_H5 radar data support (FMI, SMHI, DMI, OPERA). Pure Rust via hdf5-reader crate. Reads gain/offset/nodata/undetect from /what, projection from /where (PROJ4 string), pixel data from /dataset1/data1/data. Small files (~35KB) fully loaded into memory.

mrauhala:
Delivered. The \`engine-odim\` crate ships the native ODIM_H5 engine well beyond the original 4-phase scope:

- Core ODIM parser + PROJ-string CRS (stereographic/TM/LAEA/longlat), config + directory polling, EdrEngine + MapEngine, and the `"odim"` / `"odim-volume"` server wiring — all four design phases.
- COMP composites for DMI/DWD/OPERA/SMHI-format files; PVOL polar volumes (per-site collections, Features network inventory, cross-sections).
- Extras since: HTTP(S)/template discovery for non-listable sources, lazy PVOL pixel loading (bounded RAM, non-blocking scan), human-readable labels + site-prefixed WMS titles.

Shipped across #285/#286/#287/#290/#310–#316 (and more). Closing as part of epic #49 cleanup; the issue had been left open while the work landed via separate PRs.

PR #176, opened 2026-05-14
feat(engine-odim): ODIM_H5 weather radar engine (Phase 1, MapEngine)

mrauhala:
## Summary

Phase 1 of `engine-odim` — a native ODIM_H5 weather-radar engine implementing `MapEngine` (WMS / Maps / Tiles). Closes the MapEngine half of #29; `EdrEngine` lands in a follow-up PR before the engine ships to production.

The reader handles ODIM v2.0 → v2.4 with no version-gated branches — falling-back chains for `/where/xsize`, `/dataset1/{data1/,}what/`, and a u8→u16→f64 dtype probe absorb every layout variation we've encountered:

| Producer | ODIM version | Pixel type | Renders | Notes |
|---|---|---|---|---|
| DMI | v2.0 | u8 | ✓ | Non-canonical layout (gain/offset at root `/what`, no `/where/xsize`) |
| SMHI | v2.2 | u8 | ✗ | Canonical layout; blocked by `hdf5-reader 0.4` DEFLATE bug (file parses with `h5dump` fine) |
| DWD | v2.3 | u16 | ✓ | Canonical layout; 250m polar stere over Germany |
| OPERA | v2.4 | f64 | ✓ | Canonical layout; LAEA grid over Europe; pre-decoded physical dBZ |

`MapEngine::get_raster_tile` projects the WGS84 output bbox into the composite's native CRS pixel-by-pixel via `ds_core::geo::Crs::forward`, then nearest-neighbour samples. Pixel-accurate verified against known-precipitation cells (15.5 / 3.0 / 6.5 dBZ at exact lat/lons).

## What's in the crate
- **`proj.rs`** — Minimal PROJ-string parser for `stere` / `tmerc` / `laea` / `longlat`. Handles `+lat_ts` → k0 conversion with the **general spherical formula** so both polar (FMI/OPERA style, `lat_0=±90`) and oblique (DMI, `lat_0=56`) cases produce correct k0.
- **`reader.rs`** — HDF5 metadata parser + raw-pixel reader. Native bbox is computed from `xscale·xsize` + UL anchor (definitional per ODIM v2.4 §6.1.2), not from envelope-of-projected-corners — far more robust to projection numeric error. Logs `/Conventions` for diagnostic visibility.
- **`catalog.rs`** — strftime template → regex + local-directory scan, anchored to skip `.tmp` partial-upload markers. Sort + `max_files` cap.
- **`engine.rs`** — `MapEngine` impl with WGS84 + WebMercator output CRS, single-entry path-keyed cache. Engine constructor pre-loads the latest file so `raster_info()` answers from boot.

## ds-core change

`ds_core::geo::stere_forward` had LCC math (an `n = sqrt(1 + e²·cos⁴(lat0)/(1-e²))` cone constant) conflated into the stereographic formula, over-scaling projected coordinates by ~50% at mid-latitudes. The previous roundtrip tests passed because forward + inverse used the same wrong formula — but rendering DMI composites surfaced it immediately: ~99.999% of output pixels missed the source grid.

Replaced with the canonical EPSG 9809 oblique stereographic (no LCC `n`-factor on longitude, just `χ(φ)` for conformal latitude). Polar stereographic (FMI/OPERA-style `lat_0=±90`) falls out of the same formula without a special case. All existing geo roundtrip tests still pass.

## Server wiring

`engine_type = "odim"` now instantiates `OdimEngine` in `server::admin`. `apis = ["wms", "maps", "tiles"]` is the full supported set; `"edr"` is intentionally rejected at load time until `EdrEngine` lands so a config typo fails loudly instead of silently 404-ing.

## Collection configs added
- `radar-dmi-composite.toml` — DMI Denmark composite (testdata/radar-dmi/dk.com.*.h5)
- `radar-dwd-composite.toml` — DWD Germany HX composite (testdata/radar-dwd/composite_hx_*.h5)
- `radar-opera-composite.toml` — OPERA pan-European composite (testdata/radar-opera/OPERA@*@DBZH.h5)

## Test plan

- [ ] `cargo test -p engine-odim --tests` (31 tests cover PROJ parser, catalog, reader, RawPixels sample for u8/u16/f64)
- [ ] `cargo test -p ds-core --tests` (85 tests including all `geo::` roundtrips — verified stere change doesn't regress them)
- [ ] Boot the server with `--collections=radar-dmi-composite` and pan around Denmark in the `/preview` SPA — radar echoes appear over Denmark
- [ ] Same with `--collections=radar-opera-composite` over Europe
- [ ] Same with `--collections=radar-dwd-composite` over Germany
- [ ] Verify `gh issue list -l "priority: high" --search "odim"` (none expected since this closes the MapEngine half)

## Out of scope (follow-up PRs)
- `EdrEngine` impl (position + area queries) — must land before shipping to production
- Async catalog poller + `ArcSwap` for hot-reload
- S3-backed sources (currently local-directory only)
- SMHI workaround for the upstream `hdf5-reader` DEFLATE bug (likely manual `flate2` chunk decompression)
- PVOL polar volumes (Phase 3) — single-radar 3D scans, ODIM trajectory cross-sections via EDR


mrauhala:
Round-1 fixes pushed in `0f86984` (async poller) and `bdb09c3` (review polish):

**On Issue #1 (`stere_inverse` correctness)** — not a regression: `stere_inverse` does Newton iteration calling `stere_forward`, so it inherits the EPSG 9809 fix automatically. Forward and inverse stay locked by construction.

To prove it, added `stereographic_roundtrip_dmi_far_corners` (in `ds-core/src/geo.rs`) that round-trips the four explicit DMI corner coordinates — points hundreds of km from the projection origin in opposite directions. Errors are at the float-precision floor (≤1e-6 deg).

The other points addressed:
- **Issue #2** — `OdimEngine::new` now logs a WARN if `endpoint`/`bucket`/`prefix_pattern` are set, naming Phase 2/3 as the path forward.
- **Issue #3** — `select_entry` logs a WARN when the closest catalog entry is more than 2×poll_interval (min 60s) away from the requested time.
- **Issue #4** — Documented the exact `==` nodata/undetect comparison + why an epsilon test would be wrong.
- **Minor: `raster_info()` double-lock** — collapsed into a single `cached.lock()` acquisition.
- **Minor: `what_paths` order** — documented the earlier-producers-first rationale and the `wradlib` precedent.

Also landed (in `0f86984`) the **async catalog poller** that was carved out of the original PR scope — `OdimEngine` now holds `ArcSwap<Vec<CatalogEntry>>`, `poll_loop()` re-scans `data_dir` every `poll_interval_secs`, server `main.rs`/admin reload paths spawn + shutdown the poll loops alongside the other engines.

CI is green; 31 engine-odim + 85 ds-core tests pass.

mrauhala:
Round-2 fixes pushed in `20faeed`:

**Issue #1 (`expand_template` opaque error)** — `expand_template` now tracks `timestamp_closed` and emits a dedicated `CatalogError::SplitTimestamp` when it would have opened a second `(?P<timestamp>…)` group, naming the explicit `filename_pattern` + `timestamp_format` config form as the alternative. Pinned by `split_timestamp_template_is_an_error`.

**Issue #2 (mutex poison)** — `load_composite` and `raster_info` both now use `lock().unwrap_or_else(|e| e.into_inner())` with an ERROR-level log on recovery. The cached entry is just `(PathBuf, Arc<OdimComposite>)` with no invariants to protect, so recovery after a panic in a future caller is safe.

**Issue #3 (blocking `read_dir`)** — `poll_once` is now `async` and runs the scan via `tokio::task::spawn_blocking`. A separate ERROR branch handles JoinError if the spawned task panics. The poll loop's `tokio::select!` now awaits `poll_once()`.

CI green, 32 engine-odim tests pass.

mrauhala:
Round-3 fixes pushed in `e382bbc`:

**Issue 1 (dead config fields)** — dropped `exclude_patterns`, `time_window`, and `scan_days` from `OdimConfig`. Nothing in the engine ever read them, so operators setting `time_window = "-PT2H"` would have got silent no-op behaviour. Phase 2 (S3/STAC) can re-introduce them with real implementations. The `^…$` template anchor already blocks `.tmp` files, so `exclude_patterns` was doubly redundant.

**Issue 2 (`load_composite` blocking-context doc)** — added a doc comment stating that `load_composite` is blocking (`std::fs::read` + HDF5 parse) and must only be called from a blocking context. The WMS/Maps/Tiles handler chain already wraps `get_raster_tile` in `spawn_blocking`; the doc-comment is to keep a future `EdrEngine` author from forgetting.

32 engine-odim + 85 ds-core tests pass, workspace clippy clean.

mrauhala:
Round-4 fixes pushed in `f715be6`:

**Dead deps removed** — dropped `bytes` and `ds-storage` from `engine-odim/Cargo.toml`. Both appeared only in doc comments (the Phase 2 stubs that mentioned them landed dropped in round 3). Also trimmed the unused tokio `fs` feature — `poll_once` uses `spawn_blocking` over `std::fs`, not `tokio::fs`. Added `macros` explicitly.

**Stale-match WARN floor bumped from 60s to 600s** — the previous floor would have fired on healthy 5-min cadence radar feeds. 600s is a conservative envelope around 5-min cadence + clock skew + late arrivals; truly stalled feeds (no new file for >10 min) still trip the warning.

32 engine-odim + 85 ds-core tests pass.

mrauhala:
Round-5 fixes pushed in `a51d4c6`:

**Code changes**
- `reader.rs` — `what_paths` probe order swapped to canonical-first (`/dataset1/data1/what` → `/dataset1/what` → `/what`). Same for the `quantity` fallback chain. Matches ODIM v2.4 §7.4 precedence on hybrid files.
- `reader.rs` — `UL_lon`/`UL_lat` fallback now refuses to synthesise from LL/UR on projected CRSs. Returns `MissingAttribute` with an explicit message; the LL/UR axis-aligned synthesis only runs on `Crs::Wgs84`.
- `catalog.rs` — added `opera_at_separator_template` test pinning the `@` delimiter path against future `expand_template` refactors.

**Already addressed (round-5 reviewer's read was stale)**
- `stere_inverse` is Newton iteration calling `stere_forward`, so the EPSG 9809 fix in `forward` automatically applies to `inverse` — the contract is locked by construction. The existing `stereographic_roundtrip_dmi_far_corners` test exercises oblique roundtrip at four DMI corners (lat_0=56, ~600 km from origin). Comment B on the same line correctly identifies this.
- `OdimConfig` no longer carries `time_window` or `scan_days` (dropped in round-3 `e382bbc`). Only `endpoint`/`bucket`/`prefix_pattern` remain as Phase 2 stubs with the WARN that round-2/4 added.
- `Cargo.lock`'s `engine-odim` entry already excludes `bytes` and `ds-storage` — they were removed in `f715be6`'s lockfile regeneration alongside the `Cargo.toml` cleanup.

**Skipped (minor / non-actionable)**
- `engine.rs:467` `raster_info` double-lock — pure optimization; composite can't change between reads.
- `engine.rs:411` Mercator branch inside row loop — efficiency nit, no correctness issue.
- `engine.rs:333` `load_composite` blocking-context doc — already present in the function preamble.
- `reader.rs:146` f64 sentinel exact-equality comment — already present.

All 33 engine-odim unit tests pass; 86 ds-core tests pass; workspace builds clean; clippy clean.

mrauhala:
Round-6 review on `a51d4c6`: of the 13 inline comments, **12 are verbatim duplicates of round-5** flagging items that are already fixed at this SHA. Pushed `c600cc2` to address the one new substantive item.

**New finding addressed**
- `engine.rs:309` — `load_composite` cache-concurrency comment was only accurate for cache hits. Reworded to make the cold-miss behavior precise: two concurrent `spawn_blocking` callers may both read + write the slot once, with the second write overwriting the first with an identical immutable composite. Cost is at most one redundant read at the moment the slot fills; never visible to clients.

**Stale duplicates of round-5 (already fixed at `a51d4c6`)**
- `stere_inverse` (3 comments at `geo.rs:840`) — `stere_inverse` is Newton iteration calling `stere_forward`, so the EPSG 9809 fix in `forward` automatically applies to `inverse`. `stereographic_roundtrip_dmi_far_corners` covers oblique roundtrip at four DMI corners (lat_0=56, ~600 km from origin). The reviewer's own comment B on the same line confirms this. No action needed.
- `engine.rs:101` and `config.rs:368` — `time_window` and `scan_days` are not on `OdimConfig` (dropped in `e382bbc` round-3). Only `endpoint`/`bucket`/`prefix_pattern` remain, and the WARN already covers all three (`engine.rs:107-115`).
- `Cargo.toml:7` — `bytes` and `ds-storage` are not in `engine-odim`'s dependencies (dropped in `f715be6` round-4). `Cargo.lock` line 884 confirms.
- `catalog.rs:196` — `SplitTimestamp` variant + re-entry detection added in `e382bbc` round-3 (`catalog.rs:35`, `catalog.rs:201-205`).
- `catalog.rs:230` — `opera_at_separator_template` test added in `a51d4c6` round-5 (`catalog.rs:286-302`).
- `engine.rs:333` — blocking-context doc comment present on `load_composite` since `f715be6` round-4.
- `reader.rs:146` — f64 sentinel exact-equality rationale comment present since `e382bbc` round-3.

**Minor nits acknowledged but skipped**
- `engine.rs:467` raster_info double-lock — pure optimization; composite cannot change between reads.
- `engine.rs:411` Mercator branch in row loop — efficiency nit, no correctness issue.

CI green at `a51d4c6` (and will be at `c600cc2` once the run completes). Since round-6 is showing no new substantive progression beyond the cache-comment refinement, this PR is in a settled state from a review standpoint.

mrauhala:
Round-7 fixes pushed in `8650c07`. All 5 findings addressed.

**1. `geo.rs` — stere projection was actually still wrong, and the new test caught it**

The round-5 `stere_forward` rewrite removed the LCC cone factor (correct) but kept `(rn0·rm0).sqrt()` for the radius — that's the **double-stereographic / EPSG 9809** form (PROJ's `+proj=sterea`), not the regular **Snyder ellipsoidal oblique stereographic** (PROJ's `+proj=stere`) that DMI / OPERA actually ship. Forward+inverse roundtripped fine because the Newton-iteration inverse locks to whatever forward computes, but absolute coordinates diverged from PROJ by ~300 m at 100 km from the origin.

Switched `stere_forward` to Snyder §21 (eq. 21-26..30) ellipsoidal oblique aspect: `factor = 2·a·k₀·m_c / (cos(χ_c) · b)`. The new `stereographic_inverse_absolute_dmi` test pins `inverse(x_proj, y_proj)` against three independent PROJ reference points (computed via `cs2cs +proj=longlat +ellps=WGS84 +to +proj=stere +lat_0=56 +lon_0=10.5666 +k_0=1 +ellps=WGS84 +units=m`). All three roundtrip within 1e-5 deg (~1 m at 56°N). The old buggy formula would have failed this test — which is the whole point of the reviewer's request for an absolute-coord check.

**2. `catalog.rs::from_pattern` — `.tmp` risk doc warning**

Added prominent doc-warning explaining that unanchored patterns will match `radar.h5.tmp` / `radar.h5.part` partial-upload markers and could surface a corrupt tile as the "latest" timestep. Callers are instructed to include `^` and `$` unless they have a deliberate reason not to.

**3. `catalog.rs::scan_local_directory` — log dropped DirEntry errors**

Replaced `read_dir(...)?.flatten()` with an explicit per-entry `match` that emits `tracing::warn!("[catalog] failed to read entry in {dir}: {e}")` and `continue`s. A broken symlink or permissions error now has a log signal.

**4. `reader.rs` — diagnose SMHI DEFLATE failure**

When `hdf5-reader 0.4` silently fails DEFLATE decompression on an SMHI file, all three dtype probes return `Err` and the operator sees `UnsupportedPixelType` — misleading (implies dtype mismatch, not decompression bug). Now the u8 probe error is logged at WARN before the fallback chain, so the upstream `hdf5-reader` issue is diagnosable.

**5. `crates/engine-odim/tests/integration.rs` — end-to-end pipeline test**

Added a real integration test using a checked-in 35 KB DMI v2.0 fixture (`testdata/odim-dmi-fixture.h5`). Four tests cover:
- `raster_info_reports_dmi_extents`: native CRS = stereographic, WGS84 corners cover Denmark, 1 timestep.
- `get_raster_tile_over_denmark_renders`: 64×64 WGS84 tile rendered without error.
- `get_raster_tile_in_web_mercator_renders`: same bbox in WebMercator output (exercises the `merc_y_to_lat` branch).
- `get_raster_tile_full_extent_does_not_panic`: 32×32 tile spanning the full composite extent.

Tile content (pixel-by-pixel values) is intentionally not asserted because the fixture is a clear-air January snapshot — most pixels are `undetect` and get masked to `None` by design. The pipeline composes correctly across `OdimEngine::new` → catalog scan → `Hdf5File` read → projection parse → `stere_forward` → nearest-neighbor sample. The "scope creep" caveat: a full HTTP-layer test wiring through `api-wms::router` + `tower::ServiceExt` would require adding `axum`/`tower`/`http-body-util`/`api-wms` as dev-deps to engine-odim; the trait-level test catches the same regressions modulo the WMS query-param parser which is exhaustively covered by `api-wms`'s own tests.

**Stats**: 87 ds-core + 33 engine-odim unit + 4 engine-odim integration = 124 tests pass; clippy clean; workspace builds.

mrauhala:
Round-8 fixes pushed in `d2ef639`. All 4 new substantive findings addressed.

**1. `geo.rs` polar stereographic (Medium)**

The round-7 Snyder switch fixed the oblique case but `+lat_0=±90` (FMI/OPERA polar composites) relied on `cos(χ_c)` being numerically nonzero by IEEE 754 rounding alone — no algebraic guarantee. Added explicit polar branches in both directions:

- **forward**: Snyder eq. 21-33 `ρ = 2·a·k₀·t/D` where `D = √((1+e)^(1+e)·(1−e)^(1−e))`.
- **inverse**: Snyder eq. 21-35..37 — iterative φ from t (initial guess `π/2 − 2·atan(t)`, then `φ_{i+1} = π/2 − 2·atan(t·((1−e·sinφ_i)/(1+e·sinφ_i))^(e/2))`), longitude from `atan2(dx, −dy)` (north) / `atan2(dx, dy)` (south). The Newton-on-forward inverse is genuinely singular at the pole because longitude is degenerate there, so this needed a dedicated branch.

New `stereographic_inverse_absolute_polar` test pins against three PROJ reference points for `+proj=stere +lat_0=90 +lon_0=25 +lat_ts=60 +ellps=WGS84`; all three roundtrip within 1e-4 deg (~11 m). Switching happens at `|lat0| − π/2| < 1e-10` (exact equality modulo float precision).

**2. `proj.rs` ellipsoidal `k0` from `+lat_ts` (Medium follow-on)**

While verifying the polar fix, found that our spherical `k0 = (1 + sin|lat_ts|)/2` is ~0.012% off for the ellipsoid — ~200 m on a 3000 km radius. PROJ's `+proj=stere +lat_ts=…` internally uses case-2 `ρ = a·m_c·t/t_c` (Snyder eq. 21-39) which is independent of any explicit `k0`. To make our case-1 formula match exactly, the ellipsoidal-corrected `k0` is:

```
k0 = m_c · D / (2 · t_c)
   where m_c = cos(lat_ts) / √(1 − e²·sin²(lat_ts))
         t_c = tan(π/4 − lat_ts/2) · ((1+e·sinlat_ts)/(1−e·sinlat_ts))^(e/2)
         D   = √((1+e)^(1+e) · (1−e)^(1−e))
```

For `lat_ts=60°` this gives `k0 ≈ 0.9330690717363566` (vs spherical `0.9330127018922193`). Only applied for polar projections; oblique stays on the general spherical form because (a) no matching case-2 formula applies and (b) DMI / SMHI / DWD all ship `+k=1` explicitly anyway. Updated `fmi_polar_stereographic_with_lat_ts_converts_to_k0` test for the new expected value.

**3. `reader.rs` dtype probe WARN spam (Medium)**

The round-7 WARN-on-u8-probe-fail fired for every DWD (u16) and OPERA (f64) load, polluting the log with bogus "possible DEFLATE bug" lines. Moved per-probe failures to `debug!`, only escalating to `warn!` when all three probes fail. At that point a likely upstream cause (`hdf5-reader 0.4` mishandling DEFLATE on SMHI files) is worth surfacing.

**4. `reader.rs` `v as u32` range check (Medium)**

Unchecked f64→u32 cast for `/where/xsize`/`/where/ysize`: negative wraps to huge u32, oversize float truncates, NaN becomes 0. All three would corrupt the bbox silently. Added explicit `match` rejecting non-finite, non-positive, or > `u32::MAX` values with a clear `ReadError::DatasetRead` message. Missing attribute still falls back to data-array shape (the documented path); only malformed values are rejected.

**5. `catalog.rs::from_pattern` runtime warning (Low)**

Doc comment alone isn't enough for operators who read config docs but not source. Added a `tracing::warn!` at runtime when `from_pattern` receives a pattern without `^`/`$` anchors, mirroring the doc-comment text and naming the offending pattern.

**Stale duplicates of rounds 5-7 (already fixed)**

The reviewer cited 11-12 of the usual stale items again at `geo.rs:850` (stere_inverse — note this round still says "EPSG 9809" although we switched to Snyder in round-7), `reader.rs:146` f64 sentinel comment, `engine.rs:101` S3 fields with non-existent time_window/scan_days, `engine.rs:472` double lock, `config.rs:368` S3 fields, `catalog.rs:215` split timestamp, `engine.rs:338` blocking I/O, `engine.rs:416` Mercator branch nit, `Cargo.toml:7` dead deps, `catalog.rs:249` OPERA test. All addressed in earlier commits.

**Tests**: 88 ds-core lib + 33 engine-odim lib + 4 integration = 125 pass; clippy clean.

mrauhala:
Round-9 fixes pushed in `2707021`. 2 new substantive findings addressed; remaining 17 inline comments are the now-familiar stale duplicates from rounds 5-8 (still citing the EPSG-9809 stere_inverse comment even though we've been on Snyder since `8650c07` round-7).

**1. `proj.rs` `+R=` silent ignore (Medium, latent)**

Phase 1's three tested producers (DMI, DWD, OPERA) all ship `+ellps=WGS84`, so the hardcoded `WGS84_A` / `WGS84_E2` constants in `stere_forward` already match. But `+R=` was silently ignored — a future FMI / KNMI sphere-based producer (`+R=6371228`) would combine the producer's sphere-metre `xscale` / `yscale` with our ellipsoid-metre forward output, drifting the bbox by ~0.1% (~5 px at OPERA scale).

Added `check_radius()` called from `parse_stere`, `parse_tmerc`, `parse_laea`, returning a new `ParseError::SphereProjection { radius, percent }` when `+R=` differs from `WGS84_A` by more than 1 m. A redundant `+R=6378137` (= `WGS84_A` exactly) is still accepted as a no-op. Threading `R` properly through `Crs::Stereographic` and adjusting `stere_forward` is tracked as a follow-up.

Three new tests cover the rejection: `sphere_radius_distinct_from_wgs84_is_rejected`, `sphere_radius_matching_wgs84_a_is_accepted`, `sphere_radius_rejected_for_tmerc_and_laea`. Also updated `fmi_polar_stereographic_with_lat_ts_converts_to_k0` to use `+ellps=WGS84` instead of the now-rejected `+R=6371228`.

**2. `engine.rs` cache different-path concurrent miss (informational)**

Documented in the `load_composite` cache docstring that two adjacent-timestep requests during a time-slider scrub can both miss, both read the file, and have the second writer evict the first's entry. Not a correctness issue (both callers hold valid `Arc`s, immutable data), but worth surfacing for the LRU upgrade — a multi-entry cache eliminates the redundant read.

**Tests**: 88 ds-core lib + 36 engine-odim lib + 4 integration = 128 pass; clippy clean.

mrauhala:
Round-10 fixes pushed in `3a127b1`. 4 new substantive findings addressed; remaining 17 of 24 inline comments are the now-familiar stale duplicates from rounds 5-9 (still citing the EPSG-9809 stere_inverse comment even though we've been on Snyder since round-7, and the `+time_window`/`scan_days` OdimConfig fields that were dropped in round-3).

**1. `engine.rs` shutdown race (Medium, real bug)**

The reviewer correctly identified that `let (shutdown_tx, _) = watch::channel(())` drops the initial receiver, and that `watch::Sender::send` returns `Err` when there are no live receivers AND skips bumping the value version. So a `shutdown()` call before `poll_loop()` invokes `subscribe()` was silently lost.

Both reviewer-suggested fixes turn out to be insufficient on their own:
- Keeping a no-op receiver alive lets `send` succeed and bump the version, but `watch` is level-triggered — a fresh `subscribe()` after the bump sees the latest value as its starting state, so `changed().await` blocks until the *next* send.
- `tokio_util::sync::CancellationToken` does work, but adds a new top-level dep.

Switched instead to `AtomicBool + Notify`:
- `shutdown: AtomicBool` is the authoritative, edge-triggered flag. Checked at the top of every poll iteration, so it catches the pre-loop, between-tick, and post-tick races uniformly.
- `shutdown_notify: Notify` wakes the loop when it's currently parked on `select!` so `shutdown()` doesn't have to wait for the next `interval.tick()`.

Added `shutdown_before_poll_loop_takes_effect` regression test that pre-fires the flag *and* calls `notify_waiters()` with no live waiters (worst-case timing), then spawns the loop and asserts it exits within 1 second under a 1-hour interval — proving the AtomicBool check is doing the work, not the notify.

**2. `proj.rs` thiserror format spec (reviewer claim unfounded)**

The reviewer claimed `#[error("... {percent:.3}% ...")]` would render the format spec literally because `thiserror` only supports bare `{name}` references. That's not how thiserror works — `#[error(...)]` desugars to a standard `write!`/`format_args!` invocation, so any format spec `format!` accepts is supported.

Verified empirically by running `format!("{e}")` on `E::Sphere { radius: 6371228.0, percent: 0.108125 }` in a scratch crate: output was `"bbox off by ~0.108% with radius=6371228"` — exactly the formatted value, not the literal `{percent:.3}`.

Pinned the contract in our codebase with `sphere_projection_error_renders_percent_format_spec`, which asserts the rendered message contains `"0.108"` and explicitly does NOT contain `"{percent"`. The test would fail loudly if any future change (e.g. thiserror major version bump) regressed the behavior to literal rendering.

**3. `engine.rs::merc_y_to_lat` (Low, doc improvement)**

Added a comment explaining why `π/2 - 2·atan(exp(-y/R))` is preferred over the textbook `2·atan(exp(y/R)) - π/2` — the negated exponent decays toward zero as |y| grows rather than growing without bound, keeping the math numerically stable across the full ±π/2 range under f64 arithmetic. Linked to https://epsg.io/3857.

**4. `geo.rs::conformal_latitude` (Low, doc improvement)**

Added a note that `φ = ±π/2` is never passed to this helper because both `stere_forward` and `stere_inverse` short-circuit through their explicit polar branches first. The helper would otherwise see `tan(π/2)` which is algebraically ∞ (IEEE 754 evaluates to ~1.63e16, and the surrounding math still rounds to π/2 correctly, but the comment makes the contract explicit so a future maintainer doesn't worry).

**Tests**: 88 ds-core lib + 38 engine-odim lib + 4 integration = 130 pass; clippy clean; full workspace builds.

mrauhala:
Round-11 fixes pushed in `e824211`. 2 new substantive findings addressed — both legitimate follow-ons to the round-8 reader hardening. The remaining 22 inline comments are stale duplicates from rounds 5-10 (the reviewer is still citing the EPSG-9809 `stere_inverse` comment, the `time_window`/`scan_days` fields removed in round-3, the `bytes`/`ds-storage` deps removed in round-4, the `proj.rs` thiserror format-spec claim disproven in round-10, etc.).

**1. `reader.rs` xsize/ysize cap (Medium)**

The round-8 range check used `v <= u32::MAX`, which still admits ~4.3 billion pixels per dimension. A malformed file declaring `xsize = 4e9` would pass the cast guard and reach `hdf5-reader`'s allocation path, attempting a multi-gigabyte allocation before any controlled error fires.

Tightened the upper bound to `MAX_GRID_DIM = 100_000` — comfortably larger than any real ODIM composite (OPERA's pan-European grid is ~4400 columns wide) so a legitimate file is never rejected, but obviously-corrupt values now fail cleanly at the metadata stage. Also refactored the duplicated xsize/ysize match arms into a shared `read_grid_dim` closure.

**2. `reader.rs` xscale/yscale validation (Medium)**

`xscale`/`yscale` were read with a bare `read_f64_attr(...)?` and never validated. A zero, negative, or NaN scale propagates silently into `src_dx`/`src_dy` in `get_raster_tile`, where every sample maps outside the grid and the tile renders fully transparent — with no error. Added a `read_scale` helper rejecting non-finite or non-positive values at read time with a clear `ReadError::DatasetRead`. This mirrors the validation the round-8 fix already applied to xsize/ysize.

**Tests**: 88 ds-core lib + 38 engine-odim lib + 4 integration = 130 pass; clippy clean.

mrauhala:
Round-12 review on `e824211`: **no new substantive findings.** All 23 inline comments are verbatim re-cites of items already addressed across rounds 5-11. I read each one line-by-line against the current code:

| Comment | Status |
|---|---|
| `geo.rs` stere_inverse "not updated to EPSG 9809" (×3) | Stale — `stere_forward`/`stere_inverse` switched to Snyder `+proj=stere` in round-7 (`8650c07`); inverse is Newton-on-forward + an explicit polar branch; absolute-coord tests pin both oblique and polar against PROJ |
| `engine.rs`/`config.rs` S3 fields incl. `time_window`/`scan_days` | Stale — those two fields were dropped in round-3 (`e382bbc`); the remaining `endpoint`/`bucket`/`prefix_pattern` already WARN at `OdimEngine::new` |
| `Cargo.toml` `bytes`/`ds-storage` dead deps | Stale — removed in round-4 (`f715be6`) |
| `catalog.rs` split-timestamp / OPERA test / from_pattern anchors | Stale — `SplitTimestamp` error (round-3), `opera_at_separator_template` test (round-5), doc warning + runtime WARN (round-7) all present |
| `reader.rs` f64 sentinel comment / WARN spam | Stale — sentinel rationale comment (round-3); WARN moved to `debug!`-per-probe + `warn!`-only-if-all-fail (round-8) |
| `reader.rs:380` xscale/yscale not validated | Stale — `read_scale` helper added in round-11 (`e824211`, lines 368-379) rejects non-finite/non-positive |
| `geo.rs` polar IEEE-754 / roundtrip-not-sufficient | Stale — `stereographic_inverse_absolute_polar` + `stereographic_inverse_absolute_dmi` added rounds 7-8 |
| `proj.rs:37` thiserror format spec | Stale — disproven empirically in round-10; `sphere_projection_error_renders_percent_format_spec` test pins it |
| `engine.rs` blocking-I/O doc / EPSG:3857 comment / conformal_latitude note | Stale — all three comments added rounds 4/10 |
| `engine.rs` double-lock / Mercator-branch nits | Acknowledged non-blocking nits — pure optimizations, no correctness impact |

**Convergence**: round-11 surfaced 2 genuinely new items (both follow-ons to the round-8 reader hardening), round-12 surfaces 0. The remaining 23 comments are the fixed set `claude-review` re-emits every round because it re-evaluates the full diff statelessly. CI green across `Check, Lint & Test`, `claude-review`, and `Docker Build`.

**Final state**: 88 ds-core lib + 38 engine-odim lib + 4 engine-odim integration = 130 tests pass; clippy clean; workspace builds. From a review standpoint this PR is settled — 12 rounds, every substantive finding addressed.

PR #177, opened 2026-05-15
feat(engine-odim): EdrEngine — position + area queries (Phase 1.5)

mrauhala:
## Summary

ODIM Phase 1 (#176) shipped `MapEngine` only — WMS/Maps/Tiles raster output. Phase 1.5 adds the `EdrEngine` impl so the same ODIM_H5 radar composites answer OGC API - EDR **position** and **area** queries, reaching capability parity with the GeoTIFF raster engine.

- **position** — `POINT(lon lat)` / bare `lon,lat` → bilinear interpolation at that point for every catalog timestep in range → `PointSeries` coverage. Bilinear reprojects into the composite's native CRS, samples the four surrounding pixel centres, and degrades to nearest-neighbour when a neighbour is nodata/undetect.
- **area** — WKT polygon or `w,s,e,n` bbox → a regular WGS84 lon/lat output grid sized to the source's native resolution (capped 256/dim to bound response size), each cell bilinearly sampled, cells outside the polygon masked to `None` → `Single(Grid)` coverage.
- metadata: parameters (with configured unit), temporal extent + discrete `available_times` from the catalog, spatial extent from the seed composite, `supported_query_types = ["position","area"]`.
- no station list on a gridded radar field → `get_locations` empty, `query_location` unsupported (matches GeoTIFF).

## Implementation notes

- New `crates/engine-odim/src/edr.rs`. `OdimEngine`'s catalog/parameter/cache fields and `load_composite` became `pub(crate)` so the sibling module reuses the Phase 1 catalog + composite-cache machinery — no duplication.
- `server/src/admin.rs`: `odim` supported-APIs set gains `"edr"`; the odim load arm registers into `edr_engines` when the collection lists `edr`.
- The three ODIM collection configs (`radar-dmi`, `radar-dwd`, `radar-opera`) add `"edr"` to `apis`.
- No `api-edr` changes — the position/area endpoints + CoverageJSON serializers are engine-agnostic and already exist.

## Test plan

- [x] `cargo test -p engine-odim` — 43 lib + 10 integration tests pass (6 new EDR integration tests against the checked-in DMI fixture: position PointSeries shape, off-grid → `None`, unknown-parameter rejection, area Grid shape + NdArray-length invariant, polygon masking, metadata consistency)
- [x] `cargo test -p server` — 45 + 10 pass (admin wiring)
- [x] `cargo test -p api-edr` — CoverageJSON schema-validation suite passes
- [x] `cargo clippy -p engine-odim -p server --all-targets -D warnings` — clean
- [x] End-to-end against a live server: `/edr/collections/radar-dmi-composite/position` returns a 5-timestep `PointSeries`; `/area` returns a `[256,182]` `Grid` with 151 non-null interpolated pixels from a real DMI file; malformed `coords` → HTTP 400

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -130,7 +130,7 @@ Currently implemented: `PointSeries`, `Grid`.
 | GeoJSON | `FeatureEngine` | Features |
 | GeoTIFF | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
 | GRIB | `EdrEngine` + `MapEngine` | EDR, WMS, Maps, Tiles |
-| ODIM | `MapEngine` (EDR follow-up) | WMS, Maps, Tiles |
+| ODIM | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
 | PostGIS | `EdrEngine` + `FeatureEngine` | EDR (position, locations, area), Features |
 
--- a/collections.d/radar-dmi-composite.toml
+++ b/collections.d/radar-dmi-composite.toml
@@ -18,7 +18,7 @@ Files are 500m maximum-reflectivity composite (`dk.com.*.500_max.h5`) \
 in stereographic projection centered on Denmark.\
 """
 engine_type = "odim"
-apis = ["wms", "maps", "tiles"]
+apis = ["edr", "wms", "maps", "tiles"]
 data_path = "testdata/radar-dmi"
 
 [odim]
--- a/collections.d/radar-dwd-composite.toml
+++ b/collections.d/radar-dwd-composite.toml
@@ -20,7 +20,7 @@ ODIM_H5 v2.3. 250m polar stereographic grid covering Germany and \
 neighbouring countries.\
 """
 engine_type = "odim"
-apis = ["wms", "maps", "tiles"]
+apis = ["edr", "wms", "maps", "tiles"]
 data_path = "testdata/radar-dwd"
 
 [odim]
--- a/collections.d/radar-opera-composite.toml
+++ b/collections.d/radar-opera-composite.toml
@@ -18,7 +18,7 @@ Maximum reflectivity (DBZH) on a 2km Lambert Azimuthal Equal Area grid \
 covering Europe.\
 """
 engine_type = "odim"
-apis = ["wms", "maps", "tiles"]
+apis = ["edr", "wms", "maps", "tiles"]
 data_path = "testdata/radar-opera"
 
 [odim]
--- a/crates/engine-odim/src/edr.rs
+++ b/crates/engine-odim/src/edr.rs
@@ -22,6 +22,12 @@
 //! directories hold at most `max_files` entries (typically ≤288 at
 //! 5-min cadence), so this is acceptable for v1; a multi-entry
 //! composite cache is a follow-up.
+//!
+//! Those reads are blocking (HDF5 parse). The api-edr handlers call
+//! `EdrEngine` methods directly from `async fn`s without
+//! `spawn_blocking`, so the work currently lands on a Tokio worker —
+//! a pre-existing api-edr gap that affects every EDR engine, tracked
+//! in issue #178 and to be fixed in the api-edr handlers.
 
 use std::collections::HashMap;
 
--- a/crates/engine-odim/src/edr.rs
+++ b/crates/engine-odim/src/edr.rs
@@ -1,0 +1,536 @@
+//! `EdrEngine` impl for engine-odim — position and area queries
+//! against the same ODIM_H5 composite grid the `MapEngine` impl
+//! serves.
+//!
+//! Both query types resample the native (projected) radar grid to
+//! WGS84 query geometry:
+//!
+//! - **position** — bilinear interpolation at a single `(lon, lat)`,
+//!   one value per catalog timestep, returned as a `PointSeries`
+//!   coverage.
+//! - **area** — a regular WGS84 lon/lat grid covering the requested
+//!   polygon's bounding box, each cell bilinearly sampled and then
+//!   masked to the polygon, returned as a `Grid` coverage.
+//!
+//! ODIM composites carry no station list, so `get_locations` is
+//! empty and `query_location` is unsupported — clients use the
+//! position endpoint instead.
+//!
+//! Phase 1.5 scope: the position query loads one composite per
+//! timestep through `OdimEngine`'s single-entry cache, so a query
+//! spanning N timesteps performs N sequential file reads. ODIM
+//! directories hold at most `max_files` entries (typically ≤288 at
+//! 5-min cadence), so this is acceptable for v1; a multi-entry
+//! composite cache is a follow-up.
+
+use std::collections::HashMap;
+
+use chrono::{DateTime, Utc};
+use ds_core::edr_engine::EdrEngine;
+use ds_core::error::DataServerError;
+use ds_core::feature::{parse_area_coords, QueryPolygon};
+use ds_core::model::{
+    AreaQueryResult, DomainDescription, Location, NdArray, ParameterDescription, QueryResult,
+};
+
+use crate::catalog::CatalogEntry;
+use crate::engine::OdimEngine;
+use crate::reader::OdimComposite;
+
+/// Per-dimension cap on the area-query output grid. An ODIM
+/// composite can be ~4400 px across; without a cap an area query
+/// over the whole grid would emit a multi-megabyte coverage per
+/// timestep. 256 keeps a single-timestep area response well under
+/// 1 MB while still being finer than most display use cases need.
+const MAX_AREA_DIM: usize = 256;
+
+/// Parse an EDR `coords` value for a position query into
+/// `(lat, lon)`. Accepts WKT `POINT(lon lat)` and the bare
+/// `lon,lat` shorthand — the same two forms the other engines
+/// accept (each engine carries its own copy; there is no shared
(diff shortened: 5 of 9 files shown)


mrauhala:
Round-1 review addressed in `594cd25`.

**1. `query_location` error variant — fixed.** Changed `InvalidParameter` (HTTP 400) → `LocationNotFound` (HTTP 404). A location id against a collection with no named locations is genuinely "not found", not a malformed request. Matches `CsvEngine`/`PostgisEngine`. (Note: `GeoTiffEngine` still uses `InvalidParameter` here — that's a pre-existing minor inconsistency in the sister engine; ODIM now does the more correct thing.)

**2. `get_spatial_extent` reading the render cache — fixed (refactored).** The literal bug as described — permanent `None` for an `apis = ["edr"]`-only collection — could not actually occur: `OdimEngine::new` unconditionally seeds the cache with the most-recent composite before returning (the Phase 1 code that lets `raster_info()` answer immediately). So `cached` is `Some` from construction onward.

That said, coupling collection metadata to the on-demand render cache is a real smell, so I did the refactor you suggested: `OdimEngine` now captures the seed composite's CRS label and WGS84 corner envelope into two plain fields (`seed_native_crs`, `seed_spatial_extent`) at construction. Every timestep of an ODIM collection shares the same grid, so they're stable for the engine's lifetime. Both `raster_info()` (MapEngine) and `get_spatial_extent()` (EDR) now read those fields — no `Mutex` acquisition, no dependency on cache warm-up.

**3. y-axis order (edr.rs:323) — respectfully not changed; the premise doesn't hold.** The comment states "The GeoTIFF area query emits y-values in ascending order. ODIM should match." GeoTIFF actually emits y-values in **descending** order, same as ODIM:

- `GeoTransform::pixel_to_world` (`core/src/geo.rs:343`) computes `y = origin_y - (row + 0.5)·pixel_height`, with `pixel_height` stored positive (`geo.rs:270`).
- `GeoTiffEngine::query_bbox` builds `y_values` by mapping `row_start..row_end` (ascending row index) through `pixel_to_world` — ascending row → **descending latitude**, so `y_values[0]` is the northernmost.

ODIM's `query_polygon` emits `y_values` north→south too — it is already consistent with the sister raster engine. GeoTIFF's area output passes the CoverageJSON schema-validation suite in `api-edr/tests/covjson_validation.rs`, and CoverageJSON axis `values` are coordinate values indexed positionally by the `NdArray` (the spec's `start`/`stop`/`num` axis form explicitly permits `stop < start`), so a descending axis is not a schema violation.

Flipping only ODIM to ascending would make it inconsistent with GeoTIFF. If there's a genuine spec concern about descending y, it applies equally to GeoTIFF and should be fixed in both engines together as a separate consistency change — not unilaterally here.

**4. `POINT (lon lat)` space-form test (edr.rs:52) — already covered.** No action needed: `parse_point_coords_accepts_wkt` already asserts both `POINT(10.5 56.0)` and `POINT (10.5 56.0)` (with the space) parse identically.

Tests: 43 engine-odim lib + 10 integration pass; clippy clean; server builds.

mrauhala:
Round-2 addressed in `bcf96f8`.

**`load_composite` stale docstring (engine.rs:363) — fixed.** The docstring described the `EdrEngine` impl as hypothetical; it now exists. Updated to describe both real call paths: `MapEngine::get_raster_tile` runs inside `spawn_blocking` (WMS/Maps/Tiles handlers), while `EdrEngine::query_position`/`query_area` are called directly from api-edr's async handlers without it.

**Blocking I/O on the Tokio worker (edr.rs:230) — filed as issue #178.** Confirmed: `api-edr/src/handlers.rs` `position_query`/`area_query`/`location_query` are `async fn`s that call the engine directly with no `spawn_blocking`. As the review notes, this is an api-edr-level gap affecting **every** EDR engine — GeoTIFF, QueryData, and GRIB all do blocking I/O inside `query_position` too — so the fix belongs in the api-edr handlers, not per-engine. Tracked in **#178** and referenced from both `load_composite`'s docstring and the `edr.rs` module docstring. No ODIM code change is appropriate here.

**y-axis order (edr.rs:323) — re-cited verbatim; still not changed.** I checked the file the comment points at: `api-edr/src/response.rs::build_domain` (line 193) inserts the engine's `y` vec into the JSON **verbatim** — there is no reversal at the serialization layer. So whatever order an engine puts in `DomainDescription::Grid.y` is what ships. `GeoTiffEngine::query_bbox` builds `y_values` by mapping an ascending row range through `GeoTransform::pixel_to_world` (`y = origin_y − (row+0.5)·pixel_height`, `pixel_height` positive), which yields **descending** latitude. ODIM emits descending too — the two raster engines are already consistent, and GeoTIFF's output passes the CoverageJSON schema-validation suite.

If descending y is genuinely a spec problem it affects GeoTIFF identically and should be fixed in both engines together (a separate change) — flipping only ODIM here would make a client that queries both engines' EDR area endpoints see opposite y-orderings, which is strictly worse than the current consistent behavior. Happy to do that cross-engine change as a follow-up if you'd like, but it shouldn't land unilaterally in this PR.

**`POINT (lon lat)` space-form test (edr.rs:52) — already covered**, no action (the `parse_point_coords_accepts_wkt` test asserts the space form).

Tests: 43 engine-odim lib + 10 integration pass; clippy clean.

mrauhala:
Round-3 addressed in `a801d74`. Two new substantive findings fixed; the other five comments are verbatim re-cites of items already resolved in rounds 1-2.

**Area query — no timestep cap (resource exhaustion) — fixed.** An area coverage is an `ny × nx` grid (each ≤ 256) *per timestep*, so an unfiltered `datetime = None` query over a full 5-min-cadence catalog (~288 entries) would pre-allocate hundreds of MB. Added `MAX_AREA_TIMESTEPS = 64`: `query_polygon` now rejects an over-cap area query with `InvalidParameter` (HTTP 400) and a message telling the client to narrow `datetime`, rather than silently truncating. Worst case is now 64 × 256 × 256 ≈ 67 MB. The position query keeps no cap — it yields `N` scalars, not `N · ny · nx`.

**Area query — hard-fail on a bad first file — fixed.** It did `load_composite(&entries[0].path)?` purely to read `xsize`/`ysize` for grid sizing, so one unreadable first file aborted the whole query — inconsistent with the position query and the area query's own per-timestep loop (which degrade gracefully). Took the reviewer's option 2: captured the seed composite's grid dimensions into new `OdimEngine` fields `seed_xsize`/`seed_ysize` at construction. `query_polygon` now sizes the output grid from those fields with **no probe load at all** — which also eliminates the redundant double-load of `entries[0]` flagged in the separate minor comment.

New integration test `edr_query_area_rejects_too_many_timesteps` builds a 65-file catalog (fixture copied to 5-min-spaced timestamps) and asserts the unbounded query is rejected while a 7-timestep window succeeds.

**Re-cited (already addressed):**
- y-axis order (edr.rs:329) — addressed in the round-1/2 discussion: `response.rs::build_domain` passes the engine's `y` vec verbatim, GeoTIFF's engine emits descending y via `GeoTransform::pixel_to_world`, so ODIM already matches the sister engine. Flipping only ODIM would diverge them.
- blocking I/O (edr.rs:236) — filed as #178 (api-edr-level, affects all EDR engines), referenced from both docstrings.
- `load_composite` docstring (engine.rs:370) — already updated in `bcf96f8`; the current text (engine.rs:353-369) describes both real call paths and cites #178. The re-cite quotes the pre-`bcf96f8` text.
- `POINT (lon lat)` space test (edr.rs:58) — already covered by `parse_point_coords_accepts_wkt`.

Tests: 43 engine-odim lib + 11 integration pass; clippy clean; server builds.

mrauhala:
Round-4 review on `a801d74`: **no new findings.** All 7 inline comments are verbatim re-cites of items already resolved — including three (#5/#6/#7 below) fixed in `a801d74` itself, the very commit under review. Verified each against current code line-by-line:

| Comment | Status |
|---|---|
| y-axis order (edr.rs:356) | Addressed rounds 1-3. `response.rs::build_domain` emits the engine's `y` vec verbatim; `GeoTiffEngine::query_bbox` produces descending y via `GeoTransform::pixel_to_world` — ODIM already matches the sister engine. |
| `POINT (lon lat)` space test (edr.rs:68) | Already covered by `parse_point_coords_accepts_wkt` (asserts both `POINT(...)` and `POINT (...)`). |
| `load_composite` docstring (engine.rs:376) | Updated in `bcf96f8`; current text (engine.rs:353-369) describes both call paths and cites #178. The re-cite quotes pre-`bcf96f8` text. |
| blocking I/O (edr.rs:246) | Filed as #178 — api-edr-level, affects every EDR engine; fix belongs in the api-edr handlers. |
| hard-fail on bad first file (edr.rs:391) | **Fixed in `a801d74`.** The `load_composite(&entries[0].path)?` probe is gone — `query_polygon` now sizes the grid from `seed_xsize`/`seed_ysize` (edr.rs:328-336). No `?`, no hard-fail. |
| no timestep cap (edr.rs:419) | **Fixed in `a801d74`.** `MAX_AREA_TIMESTEPS = 64` check at edr.rs:320-326 rejects over-cap area queries with HTTP 400. |
| `entries[0]` double-load (edr.rs:421) | **Fixed in `a801d74`.** The `first` probe load was removed entirely, so there is no double-load to reuse. |

`claude-review` re-evaluates the full diff statelessly each round, so it re-emits prior findings even after they're fixed (same pattern as PR #176). Round-3 surfaced 2 genuinely new substantive items; round-4 surfaces 0. CI green across `Check, Lint & Test`, `claude-review`, and `Docker Build`.

**Final state**: 43 engine-odim lib + 11 integration tests pass; clippy clean; server builds; verified end-to-end against a live server. From a review standpoint this PR is settled — 4 rounds, every substantive finding addressed, plus follow-up #178 filed for the cross-engine api-edr `spawn_blocking` gap.

PR #179, opened 2026-05-15
fix(engine-odim): report the true WGS84 envelope of projected grids

mrauhala:
## Summary

The ODIM reader derived a composite's reported WGS84 extent as `[LL_lon, LL_lat, UR_lon, UR_lat]` — the lower-left and upper-right corner attributes treated as `[west, south, east, north]`. That holds only for an axis-aligned lon/lat grid. A **projected** grid (LAEA, stereographic, …) is a quadrilateral that bows in lon/lat, so the LL→UR diagonal under-reports the real extent.

Surfaced as an EDR `/collections` `extent.spatial.bbox` mismatch against the **GeoTIFF** version of the same OPERA data — GeoTIFF edge-samples (`ds_core::geo::GeoTransform::bbox`), ODIM used the corner shortcut.

### OPERA (LAEA) — the worst case

| Edge | Corner shortcut | True envelope | Miss |
|---|---|---|---|
| west | −10.43° (`LL_lon`) | **−39.54°** (`UL` corner) | ~29° |
| north | 67.62° (`UR_lat`) | **73.92°** (top-edge bow) | ~6° |

The north miss is the interesting one: the LAEA grid's *top edge* bows ~6° past **both** top corners (`UL` 67.0°, `UR` 67.6°), so even a four-corner envelope would still be wrong. Per-edge sampling is what catches it.

## Fix

Compute the extent the way GeoTIFF already does. New `wgs84_envelope(crs, native_bbox)` samples 20 points along every edge of the native projected `bbox`, inverse-projects each through the CRS, and takes the min/max envelope. The `OdimComposite` field is renamed `wgs84_corners` → `wgs84_bbox` to stop implying it holds raw corner values.

Corner attributes `LL_lat`/`UR_lon` are no longer read (only used by the dropped shortcut); `LL_lon`/`UR_lat` are still read lazily as the `UL`-synthesis fallback for axis-aligned `Wgs84` grids that ship only LL+UR.

This affects every **reported** extent (EDR `/collections`, WMS/Maps/Tiles `GetCapabilities`). Tile/coverage **rendering** was already correct — it forward-projects per pixel and never consulted `wgs84_corners`.

## Test plan

- [x] New `wgs84_envelope_captures_laea_trapezoid` — reproduces OPERA's LAEA grid, asserts west ≈ −39.5° and north > 72° (the top-edge bow), not the corner values
- [x] New `wgs84_envelope_is_identity_for_wgs84_grid` — axis-aligned grids are unaffected
- [x] DMI integration extent assertions updated to the (correct, edge-sampled) values via a shared `assert_dmi_envelope` helper
- [x] `cargo test -p engine-odim` — 45 lib + 11 integration pass; clippy clean; server builds
- [x] Verified against a live server: OPERA EDR `extent.spatial.bbox` `[-10.43, 31.75, 57.81, 67.62]` → `[-39.54, 31.75, 57.81, 73.92]`

### DIFF
--- a/crates/engine-odim/src/engine.rs
+++ b/crates/engine-odim/src/engine.rs
@@ -174,7 +174,7 @@ impl OdimEngine {
             );
 
         let seed_native_crs = crs_label(&composite.crs);
-        let seed_spatial_extent = composite.wgs84_corners;
+        let seed_spatial_extent = composite.wgs84_bbox;
         let seed_xsize = composite.xsize;
         let seed_ysize = composite.ysize;
 
--- a/crates/engine-odim/src/reader.rs
+++ b/crates/engine-odim/src/reader.rs
@@ -170,11 +170,18 @@ pub struct OdimComposite {
     /// outer pixel boundaries (not pixel centres). For projected
     /// CRSes this is in metres; for `Wgs84` it's in degrees.
     pub bbox: [f64; 4],
-    /// WGS84-corner bbox `[ll_lon, ll_lat, ur_lon, ur_lat]` straight
-    /// from the file's `/where` group. Kept alongside `bbox` so
-    /// callers can pick whichever frame is cheaper for their query
-    /// without re-running the projection forward.
-    pub wgs84_corners: [f64; 4],
+    /// WGS84 bounding box `[west, south, east, north]` — the
+    /// envelope of the grid in lon/lat, computed by sampling points
+    /// along every edge of `bbox` and inverse-projecting them.
+    ///
+    /// **Not** the file's raw `LL`/`UR` corner attributes: for a
+    /// projected grid (LAEA, stereographic, …) the grid is a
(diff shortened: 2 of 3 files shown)


mrauhala:
Round-1 addressed in `3c8728c`.

**`wgs84_envelope` failure fallback — fixed.** Good catch. The all-reprojections-failed branch returned the native `bbox`, which for a projected CRS is in metres — a unit-confusion bug handing the caller metre-scale values where WGS84 degrees are expected. Changed it to return `[-180, -90, 180, 90]` (conservative, dimensionally valid). The branch remains unreachable for any real grid (`Crs::Wgs84::inverse` is the identity; projected inverses succeed near the grid), but a wrong-units fallback is still worse than a conservative-but-valid one.

45 engine-odim lib + 11 integration tests pass; clippy clean.

Issue #181, opened 2026-05-15
Migrate engine-geotiff and engine-grib onto ds-storage::discovery


mrauhala:
## Background

`engine-odim`'s S3 source support (PR for the ODIM S3 phase) introduced a shared `discovery` module in `ds-storage`:

- `TimeWindow` — ISO 8601 duration parsing, `to_range`, `scan_dates`, `max_scan_days`
- `expand_prefix_for_dates` / `expand_prefix_pattern` — strftime prefix-template expansion

These were *copied* into `ds-storage` rather than lifted from an existing engine, deliberately, to keep the blast radius small — `engine-odim` is the only consumer today.

## The duplication

`engine-geotiff` and `engine-grib` each carry their **own** copy of this logic:

- `crates/engine-geotiff/src/time_window.rs`
- `crates/engine-grib/src/time_window.rs` — its header comment literally reads `//! Copied from engine-geotiff — shared utility for time-based filtering.`
- `expand_prefix_for_dates` / `expand_prefix_pattern` live inline in `engine-geotiff/src/lib.rs`

So the same `TimeWindow` implementation now exists in **three** places. That's exactly the drift risk this issue exists to close.

## Task

Migrate `engine-geotiff` and `engine-grib` to `ds_storage::discovery`:

1. Delete `engine-geotiff/src/time_window.rs` and `engine-grib/src/time_window.rs`; replace `use`s with `ds_storage::discovery::TimeWindow`.
2. Delete the inline `expand_prefix_for_dates` / `expand_prefix_pattern` in `engine-geotiff/src/lib.rs`; use the `ds_storage::discovery` versions.
3. Move any unit tests that exercised the deleted code onto the shared module (or confirm `discovery.rs` already covers them).
4. Confirm `engine-grib` and `engine-geotiff` still build, clippy clean, and their test suites pass — pay attention to any behavioural differences between the copies (they should be verbatim, but verify).

## Scope notes

- This is pure DRY cleanup — no behaviour change intended.
- `scan_remote_with_limit` (geotiff) and the grib index-sidecar scan are **engine-specific** (COG byte-range header reads vs. index files vs. whole-file HDF5 download) and are *not* in scope — only the time-window + prefix-expansion primitives are shared.

mrauhala:
[...]

PR #182, opened 2026-05-15
feat(engine-odim): S3 object-store source (Phase 2)

mrauhala:
## Summary

Adds an S3 object-store source to `engine-odim`. The S3 config fields (`endpoint` / `bucket` / `prefix_pattern`) existed but were ignored with a WARN — this wires them up so ODIM composites can be streamed straight from an S3-compatible bucket with no local mirror.

- **Shared `ds_storage::discovery` module** (new): `TimeWindow` (ISO 8601 duration parsing, scan-date computation) + `expand_prefix_for_dates`/`expand_prefix_pattern` (strftime prefix expansion). New canonical home — *not* copied into `engine-odim`. `engine-geotiff`/`engine-grib` keep their own copies for now; migration tracked in #181.
- **`engine-odim`**: new `Source` enum — `endpoint`+`bucket` selects S3, else local `data_path`. `catalog::scan_remote` lists date-expanded prefixes, matches basenames, applies the time-window filter + `max_files` cap. Seed load, poll loop, and `load_composite` all branch on the source. `OdimConfig` gains `time_window`. `OdimEngine::new` now takes `data_path: Option<&str>`, mirroring `GeoTiffEngine::new`.
- **Config**: `collections.d/radar-opera-composite-s3.toml` — streams OPERA DBZH composites from CloudFerro's `openradar-24h` bucket with a 12 h `time_window`.

## Test plan

- [x] `cargo test -p ds-storage -p engine-odim` — 47 unit + 11 integration tests pass; new `scan_remote` tests cover match/filter/cap and multi-prefix merge via a local-filesystem object store.
- [x] `cargo clippy --workspace` — clean.
- [x] Live server load against CloudFerro: EDR `bbox` `[-39.5, 31.7, 57.8, 73.9]`, 12 h temporal window, EDR position query → 200 CoverageJSON, WMS GetMap renders a real OPERA composite over Europe.

## Notes

- CloudFerro public GET works unauthenticated — the prior "403 on GET" assumption was stale; `ds-storage` already special-cases `.cloudferro.com` with `skip_signature`.
- An EDR position query with *no* `datetime` returns 400 (datetime-range default) — pre-existing EDR-layer behaviour, identical on the local OPERA collection, not introduced here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -790,6 +790,7 @@ name = "ds-storage"
 version = "0.1.0"
 dependencies = [
  "bytes",
+ "chrono",
  "ds-core",
  "futures",
  "object_store",
@@ -885,6 +886,7 @@ dependencies = [
  "arc-swap",
  "chrono",
  "ds-core",
+ "ds-storage",
  "hdf5-reader",
  "ndarray",
  "regex",
--- a/collections.d/radar-opera-composite-s3.toml
+++ b/collections.d/radar-opera-composite-s3.toml
@@ -1,0 +1,37 @@
+# EUMETNET OPERA pan-European radar composite (ODIM_H5 v2.4) — S3 source.
+#
+# Streams `OPERA@YYYYMMDDTHHMM@0@DBZH.h5` composites directly from
+# CloudFerro's public `openradar-24h` bucket — no local mirror, no
+# gdalwarp preprocessing. New timesteps land every 5 minutes; the
+# bucket keeps a rolling 24h window.
+#
+#   https://s3.waw3-1.cloudferro.com/openradar-24h/YYYY/MM/DD/OPERA/COMP/
+#
+# 4400x3800 grid at 2km/px over Europe, Lambert Azimuthal Equal Area
+# (+lat_0=55 +lon_0=10 +ellps=WGS84). DBZH quantity in physical dBZ.
+#
+# The local-fixture twin of this collection is `radar-opera-composite`.
+
+id = "radar-opera-composite-s3"
+title = "OPERA Pan-European Radar Composite (DBZH, S3)"
+description = """\
+EUMETNET OPERA pan-European weather-radar composite via ODIM_H5 v2.4, \
+streamed from CloudFerro S3. Maximum reflectivity (DBZH) on a 2km \
+Lambert Azimuthal Equal Area grid covering Europe.\
+"""
+engine_type = "odim"
+apis = ["edr", "wms", "maps", "tiles"]
+
+[odim]
+endpoint = "https://s3.waw3-1.cloudferro.com"
+bucket = "openradar-24h"
+prefix_pattern = "%Y/%m/%d/OPERA/COMP/"
+filename_template = "OPERA@%Y%m%dT%H%M@0@DBZH.h5"
+parameter = "reflectivity"
+unit = "dBZ"
+# Keep only the last 12h of the 24h bucket in the catalog.
+time_window = "-PT12H"
+poll_interval_secs = 300
+
+[wms]
+colormap = "radar_dbz"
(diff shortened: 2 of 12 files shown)


mrauhala:
Addressed all four review findings in c305647:

- **`PathBuf` for S3 keys** (`catalog.rs`) — `CatalogEntry.path: PathBuf` replaced with a `Location` enum: `Local(PathBuf)` vs `Remote { store, key: String }`. Object keys are now `String`, never round-tripped through `PathBuf`. `Remote` also carries its own `DataStore` handle so an entry is self-contained.
- **Silent `prefix_pattern` default** (`engine.rs`) — `build_source` now returns `EngineError::MissingPrefixPattern` when an S3 source has no `prefix_pattern`, instead of defaulting to `""`. An explicit empty string is still accepted as a deliberate flat-bucket opt-in.
- **`NoFiles` doesn't name the store** (`engine.rs`) — `Source::Remote` retains `endpoint`/`bucket`; `source_label` now emits `s3 <endpoint>/<bucket>/<prefix>`.
- **No duplicate-timestamp test** (`catalog.rs`) — added `scan_remote_dedups_duplicate_timestamps_keeping_last_key`, pinning the lexicographically-last-key tie-break.

Also fixed the `cargo fmt --check` CI failure (test code added after the prior fmt pass). Workspace fmt + clippy + full test suite all green locally.

mrauhala:
Round 2 — addressed in 229caf4:

- **No size guard in `fetch_bytes`** — remote fetches now `head()` the object and reject it against `MAX_REMOTE_FILE_SIZE` before `get()`, closing the grew-after-listing TOCTOU and bounding the seed / `load_composite` paths.
- **Dedup comment clarity** — expanded to spell out that the descending-key sort puts the lexicographically-greatest key first, which is the element `dedup_by` (keep-first) retains.

The other three comments on this commit (`prefix_pattern` default, `NoFiles` store naming, `PathBuf` for keys) are re-cites of round-1 findings already fixed in c305647 — verified line-by-line against the current tree:
- `build_source` returns `MissingPrefixPattern` (engine.rs:549-552)
- `source_label` emits `s3 <endpoint>/<bucket>/<prefix>` (engine.rs:583-592)
- `CatalogEntry` uses the `Location` enum; no `PathBuf::from(key)` (catalog.rs:289-295)

mrauhala:
Round 3 — addressed in ae740ad:

- **`time_window` silently dropped for local sources** — `OdimEngine::new` now emits a `tracing::warn!` when `time_window` is set on a local `data_path` source (it only affects S3 prefix expansion / timestamp filtering).
- **`expand_*`/`max_scan_days` dead API** — removed `TimeWindow::max_scan_days` from `ds_storage::discovery`; it was `pub` but unused (engine-odim's no-`time_window` scan uses a fixed `DEFAULT_SCAN_DAYS`). The #181 migration can reintroduce it when engine-geotiff/grib — which do use it — move onto the shared module.

The other five comments on `229caf4` are re-cites of round-1/round-2 findings already fixed — verified line-by-line against the current tree:
- `build_source` → `MissingPrefixPattern` for a missing S3 `prefix_pattern`
- `source_label` emits `s3 <endpoint>/<bucket>/<prefix>`
- `CatalogEntry` uses the `Location` enum (no `PathBuf` for keys)
- `fetch_bytes` HEAD-checks remote object size before GET (engine.rs:130-137)
- the `scan_remote` dedup comment already spells out the sort/dedup-direction consistency

mrauhala:
Round 4 — addressed in 05d13ce:

- **`TimeWindow::parse` silently dropped trailing characters** — `PT2H5` parsed as 2h (bare `5` discarded), `PT2H30M5SFOO` left `FOO` unchecked. The parser now advances past the `S` component and errors on any non-empty remainder. Added `PT2H5` / `PT2H30M5SFOO` / `PT2HFOO` to the rejection test.

The other six comments on `ae740ad` are re-cites of round-1/2/3 findings already fixed (`prefix_pattern` guard, `source_label`, `Location` enum, `fetch_bytes` size guard, dedup comment, local `time_window` warn) — verified against the current tree.

Issue #186, opened 2026-05-15
engine-odim PVOL: radial spoke artifacts in polar→Cartesian rendering


mrauhala:
## Observation

The PVOL Cartesian renderer (`PolarVolumeEngine` / `volume_engine.rs::polar_sample`, milestone M2 on branch `feat/odim-pvol`) produces visible **radial spoke artifacts** — thin gaps fanning out from the radar site, most pronounced at long range.

This is **not a bug** — it is the expected consequence of two deliberate M2 interim simplifications:

1. **Nearest-neighbour sampling.** Each output Cartesian pixel maps to the single closest `(ray, bin)` polar cell. Azimuth rays diverge with range (at 250 km a 1° ray is ~4.4 km wide), so far from the radar adjacent output pixels fall between rays and leave un-sampled gaps → spokes.
2. **Ground-range = slant-range approximation.** `polar_sample` treats the sweep's range axis as ground range (`bin = (d - rstart) / rscale`). The proper slant-range → ground-range conversion (and beam height) under the standard 4⁄3-Earth refraction model is deferred.

## Refinements to do later

- **Bilinear (or area-weighted) interpolation** over the four surrounding `(ray, bin)` cells instead of nearest-neighbour — closes the inter-ray gaps. Mind the azimuth wrap at 0°/360° and the `nodata`/`undetect` mask (don't interpolate masked cells into valid output).
- **4⁄3-Earth ground-range geometry** — convert each output pixel's ground distance to the correct slant range for the sweep's elevation angle, `r = f(ground_range, elevation)`, and optionally expose beam height. This also sets up the geometry needed for M4 (vertical cross-sections).

## Scope / sequencing

Cosmetic at low elevation / short range, so low priority. Best done *after* the M2 interim lands and *before* (or together with) M4, since M4's cross-section sampling needs the same 4⁄3-Earth beam geometry. Not a blocker for M2.

Reference: ODIM_H5 v2.4 spec, §5.1 (polar data geometry) — https://www.eumetnet.eu/wp-content/uploads/2021/07/ODIM_H5_v2.4.pdf

mrauhala:
[...]

Issue #185, opened 2026-05-15
Add a vertical (elevation/level) dimension to the MapEngine and EdrEngine traits


mrauhala:
## Problem

Neither `MapEngine` nor `EdrEngine` has a vertical dimension.

- `MapEngine::get_raster_tile(bbox, width, height, time, output_crs, parameter)` — selectors are only `time` and `parameter`. A WMS/Maps/Tiles request cannot ask for a specific elevation angle, pressure level, or altitude.
- `EdrEngine` (`query_position` / `query_area` / `query_location`) — takes only `datetime` and `parameters`. There is no EDR `z` selector, no vertical extent in collection metadata, and no `z` axis in the CoverageJSON domains (only `PointSeries` and `Grid`, both vertical-less, are implemented).

This blocks proper support for every 3-D source:

- **ODIM polar volumes** (`engine-odim`, `odim-volume`) — each radar is a stack of elevation sweeps. M2 renders only the lowest sweep as an interim; EDR ("sites as locations, elevation as `z`, moments as parameters") cannot be built at all.
- **GRIB** — pressure-level, model/hybrid-level, height-level, isentropic products.
- **QueryData** — pressure / model / height levels in `.sqd` files.
- Multi-layer / 3-D GeoTIFF (COG bands; the ODIM cartesian `zsize`/`ztop` model).

## Scope

A single cross-cutting change covering **both** the Map and EDR surfaces, so the two share one vertical-dimension representation in `ds-core` rather than diverging.

### `ds-core`
- A shared vertical-dimension descriptor (levels + label + unit + coordinate kind) used by **both** `RasterInfo` (Map) and the EDR collection metadata.
- `MapEngine::get_raster_tile` gains a vertical selector (`z: Option<f64>`, or a small `Dimensions` struct so future non-vertical dims don't force another signature change).
- `EdrEngine::query_position` / `query_area` / `query_location` gain the same vertical selector; EDR collection metadata gains a vertical extent.

### `api-wms` / `api-maps` / `api-tiles`
- `api-wms`: parse the WMS 1.3.0 `ELEVATION` dimension; emit `<Dimension name="elevation">` in `GetCapabilities`.
- `api-maps` / `api-tiles`: accept an `elevation` / `z` query parameter; thread it through.

### `api-edr`
- Parse the EDR `z` query parameter on position/area/locations queries.
- Advertise the vertical extent in collection metadata.
[...]

PR #187, opened 2026-05-15
feat(engine-odim): PVOL polar-volume reader + Cartesian display

mrauhala:
## Summary

First two milestones of ODIM polar-volume (PVOL) support — *read* polar volumes and *display* them in Cartesian coordinates. Polar volumes are multi-elevation, multi-moment radar data in spherical coordinates, distinct from the 2-D `COMP` composites the engine handled before.

- **`pvol.rs`** — `read_polar_volume()` parses multi-dataset / multi-moment ODIM polar volumes into `PolarVolume { site, sweeps }`. Does **not** gate on `/what/object` (FMI mislabels polar volumes as `"SCAN"`); range-bin counts vary per elevation sweep.
- **`volume_engine.rs`** — `PolarVolumeEngine` implements `MapEngine`. One collection per radar network; each radar **site** is a layer group and each ODIM **quantity** a sub-layer (`parameter = "<site>:<quantity>"`, modelled on multi-parameter GRIB). `get_raster_tile` renders the **lowest elevation sweep**, polar→Cartesian via haversine distance + bearing.
- **`OdimConfig.parameter`/`unit`** are now `Option` (a volume collection is multi-parameter); the COMP engine requires them via a `MissingCompField` error.
- **`RawPixels::sample()`** is now NaN-aware (NaN `nodata` sentinels).
- **`engine_type = "odim-volume"`** wired into the server (WMS/Maps/Tiles), with poll loop + graceful shutdown.

### Interim scope (deliberate)

- Lowest elevation sweep only; nearest-neighbour sampling; ground-range ≈ slant-range.
- A vertical (elevation) `MapEngine` dimension is deferred to #185.
- Spoke-artifact / 4⁄3-Earth ground-range refinements tracked in #186.
- **Local-directory source only** — FMI S3 / DMI STAC sources are the next milestone (in progress).

## Test plan

- [x] `cargo test` — 6 PVOL reader tests + 17 volume-engine tests (geodesic, polar→Cartesian, parameter parsing, fixture render); full `engine-odim` + `ds-core` suites green.
- [x] `cargo fmt --all --check` + `cargo clippy` warning-clean across `engine-odim` / `ds-core` / `server`.
- [x] Reader verified against a real 15 MB FMI Anjalankoski polar volume (13 sweeps, 16 moments).
- [x] Live end-to-end: server loads a `radar-fmi-volume` collection, WMS `GetMap` for `radar-fmi-volume/fianj:DBZH` returns a valid PNG with the radar echoes rendered in Cartesian.

### DIFF
--- a/collections.d/radar-fmi-volume.toml
+++ b/collections.d/radar-fmi-volume.toml
@@ -1,0 +1,27 @@
+# FMI weather-radar polar volumes (ODIM_H5 PVOL).
+#
+# Each radar site is a stack of elevation sweeps, each with multiple
+# dual-pol moments (TH, DBZH, VRADH, ZDR, RHOHV...). The odim-volume
+# engine renders the lowest sweep of a chosen site as a Cartesian
+# raster; the WMS layer tree is one group per radar site, with the
+# moments as quantity sub-layers (parameter = `<site>:<quantity>`).
+#
+# data_path here points at the local test fixture directory; an S3
+# source for the FMI volume bucket lands in a later milestone.
+
+id = "radar-fmi-volume"
+title = "FMI Radar Polar Volumes"
+description = """\
+Finnish weather-radar polar volumes (ODIM_H5 PVOL). One layer group \
+per radar site; reflectivity and dual-polarisation moments as \
+sub-layers, rendered from the lowest elevation sweep.\
+"""
+engine_type = "odim-volume"
+apis = ["wms", "maps", "tiles"]
+data_path = "testdata/radar-fmi-pvol"
+
+[odim]
+poll_interval_secs = 300
+
+[wms]
+colormap = "radar_dbz"
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -334,12 +334,15 @@ pub struct OdimConfig {
     pub timestamp_format: Option<String>,
 
     /// Parameter name advertised to clients (e.g. `"reflectivity"`).
-    /// One parameter per collection in Phase 1 — multi-parameter PVOL
-    /// volumes ship in Phase 3.
-    pub parameter: String,
+    /// Required for single-parameter `engine_type = "odim"` (COMP)
+    /// collections. Unused — and may be omitted — by the multi-parameter
+    /// `engine_type = "odim-volume"` (PVOL) engine, which derives one
+    /// parameter per `<site>:<quantity>` pair from the volume files.
+    pub parameter: Option<String>,
     /// Unit of measurement (e.g. `"dBZ"`). Pure metadata: the engine
-    /// does not convert between units.
-    pub unit: String,
+    /// does not convert between units. Required for `engine_type = "odim"`;
+    /// optional for `engine_type = "odim-volume"`.
+    pub unit: Option<String>,
 
     /// Override nodata sentinel. Takes precedence over the file's
     /// `/dataset1/data1/what/nodata` attribute. Useful when a
--- a/crates/core/src/map_engine.rs
+++ b/crates/core/src/map_engine.rs
@@ -33,7 +33,7 @@ pub struct RasterInfo {
     pub native_crs: String,
     /// Native spatial extent [west, south, east, north] in WGS84.
     pub spatial_extent: Option<[f64; 4]>,
-    /// Available timestamps, most recent first.
+    /// Available timestamps, oldest first (ascending).
     pub times: Vec<DateTime<Utc>>,
     /// Default parameter name (e.g., "reflectivity").
     pub parameter: String,
--- a/crates/engine-odim/src/engine.rs
+++ b/crates/engine-odim/src/engine.rs
@@ -232,6 +232,12 @@ pub enum EngineError {
         #[source]
         source: crate::reader::ReadError,
     },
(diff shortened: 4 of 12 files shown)


mrauhala:
Addressed all four review findings in 0370c09:

- **O(n²) cache eviction** — dropped the `seen` Vec entirely. The parse cache is now evicted against the catalog itself, and against the **`max_files`-capped** catalog, so the cache is bounded too (not just the metadata).
- **Unbounded parse cache / ignored `max_files`** — each site's volume list is now capped to the most-recent `OdimConfig.max_files`. `VolumeEntry` carries its source path so the cache trims to exactly the retained set.
- **Missing parameter dedup** — `parameters.dedup_by(|a,b| a.0 == b.0)` after the sort.
- **Double-negated assertion** — the due-north bearing test now asserts `az < 0.01` directly.

clippy/fmt clean; `engine-odim` suite green.

mrauhala:
Reviewed all three comments on `0370c09` against the authoritative ODIM_H5 v2.4 spec — **no code changes needed**:

**1. `volume_engine.rs:291` "missing dedup"** — stale re-cite. `parameters.dedup_by(|a, b| a.0 == b.0)` was added in this very commit (`0370c09`, the round-1 fix); it is present immediately after the `sort_by`.

**2 & 3. `pvol.rs:414` / `volume_engine.rs:312` — "`rstart` is kilometres"** — incorrect. ODIM_H5 v2.4 **Table 4** (§4.3.1, *`where` Attributes for polar data objects*, document p.12) defines:

| Name | Type | Units | Description |
|---|---|---|---|
| `rstart` | double | **meters** | The range of the start of the first range bin |
| `rscale` | double | **meters** | The distance between two successive range bins |

Both are metres. The reader storing `rstart` as-read is correct — there is no 1000× error. (Independently moot anyway: FMI volumes ship `rstart = 0.0`.) If a producer is ever found shipping km, that's a producer-quirk to handle empirically, not a spec conformance fix.

**4. `volume_engine.rs:562` — "`a1gate` assumption is FMI-specific"** — not FMI-specific; it's spec-mandated. ODIM_H5 v2.4 §5.1: *"All full-sweep data must be sorted according to the above description: clockwise and starting from north, **even if they haven't been acquired that way**."* `a1gate` exists only to allow *temporal* reconstruction of acquisition order — it must not offset the stored-array index for a full sweep. The code is correct for full sweeps (FMI volumes are 360-ray full sweeps). Sector `SCAN`s are explicitly exempted by §5.1 and aren't in scope here.

CI is green on `0370c09`; the PR is ready to merge.

mrauhala:
### Review receipt — `rstart` units & `a1gate` ordering

Both findings re-cited in the latest round were verified against the authoritative **ODIM_H5 v2.4** spec (EUMETNET OPERA, OD8). Outcome:

**`rstart` units — no bug.** Table 4 (`where` attributes for polar/SCAN data) defines:
> `rstart  double  meters  The range of the start of the first range bin`
> `rscale  double  meters  The distance between two successive range bins`

Both are **metres** — no km→m scaling. The `* 1000.0` suggested fix would be incorrect for v2.4. (Older OPERA/BUFR conventions did use km, hence the confusion.) Commit `cd7089f` adds a spec-citing comment at the read site to settle this.

**`a1gate` ordering — no bug.** §5 mandates:
> "All full-sweep data must be sorted according to the above description: clockwise and starting from north, even if they haven't been acquired that way."

`a1gate` is "Index of the first azimuth gate **radiated**" — acquisition-order metadata for temporal reconstruction only. A conformant file is always stored north-first, so `a1gate` must not offset the array index. The existing comment in `volume_engine.rs` already states this accurately.

No code change needed for either beyond the documentation commit.

mrauhala:
### Review receipt — round on `cd7089f`

**`a1gate` (volume_engine.rs:565) — no change.** ODIM_H5 v2.4 §5 (SCAN/full-sweep storage):
> "All full-sweep data must be sorted according to the above description: clockwise and starting from north, even if they haven't been acquired that way."
> "The first azimuth gate in the scan does not have to be the first ray of data collected for that scan, but the `a1gate` Attribute contains the information on this, thereby allowing a temporal reconstruction of the data collection."

Full-sweep data is stored **geographic-north-first**; `a1gate` recovers *acquisition* order, not geographic order. Applying `(ray_geo + nrays - a1gate) % nrays` would rotate every conformant file with `a1gate ≠ 0` (e.g. DMI) into wrongness. The current lookup is spec-correct.

**`nbins`/`nrays` cast guard (pvol.rs:423) — no change.** `nbins`/`nrays` size no allocation: `read_moment_array` reads the array at the HDF5 dataset's *intrinsic* shape and `check()` validates `(nrays, nbins)` against it, emitting `"array shape … doesn't match sweep metadata …"`. A corrupt attribute (negative→0 or huge) fails that check with an actionable error — there is no `ndarray` OOM path from these values.

**S3 warn-and-continue (volume_engine.rs:374) — out of scope.** #187 is the local-directory M2 PR; S3 support lands on `feat/odim-pvol-s3`, which is where the warn-vs-error decision belongs (as the comment itself notes).

Merging — CI green, radar verified rendering in `/preview`.

PR #188, opened 2026-05-16
feat(engine-odim): S3 source for the PVOL volume engine

mrauhala:
## Summary

- Adds an **S3/HTTP object-store source** to the PVOL volume engine (`odim-volume`), mirroring the COMP engine's two-source model — a collection reads `.h5` polar volumes from either a local `data_path` directory or an S3 bucket (`endpoint` + `bucket` + `prefix_pattern`).
- `time_window` bounds **both** the date-partitioned prefixes scanned **and** the timestamps kept. Objects are filtered by a key-timestamp pre-fetch check *before* any `get`, so an FMI open-data bucket spanning a full day (~288 volumes) is never downloaded wholesale.
- The poll-loop scan now runs on `tokio::task::spawn_blocking` — an S3 `list` plus multi-MB downloads no longer stall a Tokio worker thread. Mirrors `OdimEngine::poll_once`.

Includes `collections.d/radar-fmi-vihti-volume.toml` — a ready-to-use config for FMI's Vihti radar from the `fmi-opendata-radar-volume-hdf5` bucket.

## Test plan

- [ ] `cargo test -p engine-odim` — the integration suite exercises the full remote scan path (`list` → `get` → parse → group-by-site) against a `LocalFileSystem`-backed `DataStore`
- [ ] a local-`data_path` PVOL collection still loads and renders (no regression)
- [ ] point a collection at the live FMI S3 bucket and confirm `time_window` bounds the download set

### DIFF
--- a/collections.d/radar-fmi-vihti-volume.toml
+++ b/collections.d/radar-fmi-vihti-volume.toml
@@ -1,0 +1,37 @@
+# FMI Vihti weather-radar polar volumes, streamed from S3 (ODIM_H5 PVOL).
+#
+# This collection streams the Vihti radar's polar volumes directly from
+# FMI's open-data S3 bucket — no local copy. FMI publishes one ODIM_H5
+# PVOL file per acquisition under a date-partitioned, per-site key
+# prefix; Vihti's ODIM site code is `fivih`, so the objects land at
+# `s3://fmi-opendata-radar-volume-hdf5/YYYY/MM/DD/fivih/` with filenames
+# `YYYYMMDDHHMM_fivih_PVOL.h5`.
+#
+# The odim-volume engine renders the lowest elevation sweep of the site
+# as a Cartesian raster; the WMS layer tree is one group per radar site
+# with the dual-pol moments as quantity sub-layers (TH, DBZH, VRADH,
+# ZDR, RHOHV...). Parameter names are `<site>:<quantity>`, e.g.
+# `fivih:DBZH`.
+#
+# `time_window = "-PT3H"` bounds both the date-partitioned prefixes
+# listed and the volume timestamps kept to the most recent 3 hours.
+
+id = "radar-fmi-vihti-volume"
+title = "FMI Vihti Radar Polar Volumes"
+description = """\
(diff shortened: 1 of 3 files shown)


mrauhala:
### Review receipt — `2f26ac7`

**`parse_key_timestamp` placement (volume_engine.rs:391) — fixed.** Correct catch: the fn was wedged between `enumerate_remote`'s doc block and its `#[allow(clippy::type_complexity)]`, so both attached to `parse_key_timestamp`. Moved `parse_key_timestamp` (with its own doc) ahead of the `enumerate_remote` doc block — each attribute/doc now lands on its intended fn.

**In-closure `head()` (volume_engine.rs:488) — kept, comment clarified.** The comparison to `catalog::scan_remote` is to the wrong function: `scan_remote` is *enumeration*. The COMP engine's *fetch* path is `engine::fetch_bytes` (engine.rs:120-138), which does `head` → size-check → `get` — exactly what this closure does. That head-before-get was added by #182's review specifically to close the grow-between-list-and-get gap; removing it here would diverge from COMP and reopen that gap. The two engines are consistent. Comment updated to name `fetch_bytes` so the parallel is explicit.

mrauhala:
### Review receipt — re-run on `2f26ac7`

**`head()` re-check (volume_engine.rs:490)** — acknowledged as mirroring COMP's `fetch_bytes`; no change. Comment updated to name `fetch_bytes` explicitly.

**Startup hard-fail when all S3 prefixes fail (volume_engine.rs:507)** — valid asymmetry, but deferred to a follow-up issue rather than changed in this PR:

- The behaviour is consistent with `OdimEngine::new()`, which propagates the same scan error (`engine.rs:279`) and is in fact *stricter* — it hard-fails even on an empty no-error catalog.
- The suggested fix of returning `Ok((vec![], time_filter))` from `enumerate_remote` is **not safe**: `poll_once` shares that path, so a transient *runtime* S3 blip would swap an empty catalog over a previously-good one. A correct fix must be scoped to `new()` alone.
- Whether `new()` should fail-fast (surfacing misconfiguration) or start empty (resilient restart) is a genuine design tradeoff worth deciding deliberately, not folding into a feature PR.

Filed as #190.

PR #191, opened 2026-05-17
fix(server): bind listen port before loading collections

### LINKED ISSUE #189: Server binds listen port after engine loading — port conflict panics late
## Problem

In `crates/server/src/main.rs` the listen socket is bound only at the
end of startup (`main.rs:320`) — *after* `admin::load_collections()`
(`main.rs:163`) has constructed every engine.

S3-backed engines (COMP `odim`, PVOL `odim-volume`, GeoTIFF, GRIB) run
a synchronous boot scan in their constructors that downloads and parses
data files; with a wide `time_window` this can take many seconds to
minutes.

If the configured port is already in use, `TcpListener::bind()` fails —
but the operator only learns this *after* sitting through the entire
data load, when the process panics via `.expect("Failed to bind")`.

## Expected

A port conflict should fail fast, before any data is fetched.

## Fix

Bind the `TcpListener` early — right after the config is parsed and the
bind address is known, before `load_collections()` — and hold it until
`axum::serve()`. Also consider replacing `.expect("Failed to bind")`
with a clean error message and a non-panic exit.

mrauhala:
## Summary

- Binds the listen socket **before** `load_collections()` instead of after. Engine construction runs synchronous S3/disk boot scans that can take minutes; with the bind at the end of startup, a port conflict surfaced only as a late panic — the operator waited through the entire data load to learn the port was in use.
- A bind failure now logs a clean `ERROR` and exits non-zero, rather than panicking via `.expect("Failed to bind")`.

Closes #189.

## Test plan

- [x] `cargo build -p server` + `cargo clippy -p server` clean
- [x] With the port free, the server binds, loads collections, and serves normally
- [x] With the port already held by another instance, the second instance logs `Failed to bind 127.0.0.1:8000: Address already in use` **immediately** (before any `Loaded … collections` log) and exits — verified no panic

### DIFF
--- a/crates/server/src/main.rs
+++ b/crates/server/src/main.rs
@@ -160,6 +160,20 @@ async fn main() {
     let base_url = config.server.base_url();
     info!("Base URL: {base_url}");
 
+    // Bind the listen socket before loading collections: engine
+    // construction runs synchronous S3/disk scans that can take
+    // minutes, so a port conflict must fail fast rather than after
+    // that whole load. The listener is held until `axum::serve`.
+    let addr = format!("{}:{}", config.server.host, config.server.port);
+    let listener = match tokio::net::TcpListener::bind(&addr).await {
+        Ok(listener) => listener,
+        Err(e) => {
+            tracing::error!("Failed to bind {addr}: {e}");
+            std::process::exit(1);
+        }
+    };
+    info!("Listening on {addr}");
+
     let result = admin::load_collections(&config.collections, &config.style_bundles, &base_url);
 
     let loaded = result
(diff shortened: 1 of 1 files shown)


mrauhala:
### Review receipt — `cc32f3a`

**Pre-serve log message (main.rs:175)** — applied. The log now reads `Socket bound to {addr} — loading collections, not yet serving`, and `Server ready, accepting requests` remains the single authoritative HTTP-ready marker.

On the broader point: yes, binding early means `listen(2)` is active during the load, so a **TCP-only** readiness probe (k8s `tcpSocket`, bare `nc`) will see the port open before HTTP works — that's inherent to fixing #189 (the port conflict must be detectable before the minutes-long load). The correct probe is an HTTP check against `/health`, which only answers once `axum::serve` is up. Added an inline comment to that effect so the tradeoff is documented at the call site.

Issue #197, opened 2026-05-18
odim-volume EDR (M3b): elevation as z, VerticalProfile


mrauhala:
Part of #195. Depends on **#185** (vertical dimension) and **M3a**.

Once #185 lands the shared vertical dimension, complete the `odim-volume` EDR surface:

## Scope

- `z` = elevation angle selector on `query_position` / `query_location` / `query_area`.
- Advertise the elevation sweeps as the collection's vertical extent in EDR metadata.
- Response shape per the #185 EDR query → CoverageJSON table:
  - Point query, no `z` → `CoverageCollection` of `VerticalProfile` (reflectivity-vs-elevation-angle at the site, one per timestep).
  - Point query, `z` → `PointSeries` per selected sweep.
- Drop the M2 lowest-sweep-only interim in the `MapEngine` path — expose all sweeps via the `z` selector (this is the radar consumer of #185's `MapEngine` vertical work).

mrauhala:
Completed and merged. The PVOL `EdrEngine` exposes elevation angle as the vertical dimension:
- `get_vertical_extent()` returns the deduped sweep angles (`VerticalKind::ElevationAngle`)
- no-`z` point queries → `CoverageCollection` of `VerticalProfile`s (one per timestep)
- `z` selects sweeps → `PointSeries` per level

Implemented via `volume_profile` / `level_series` / `site_coverages` / `resolve_levels` in `crates/engine-odim/src/volume_engine.rs`. The split-cut z-axis dedup edge cases were fixed in #269/#270. Closing as done.

Issue #196, opened 2026-05-18
odim-volume EDR (M3a): sites as locations + position queries


mrauhala:
Part of #195. The non-vertical slice of `EdrEngine` for `PolarVolumeEngine` — does **not** depend on #185.

`PolarVolumeEngine` is `MapEngine`-only today. Implement `EdrEngine` for the parts that need no vertical dimension:

## Scope

- `get_locations()` — map `catalog.by_site` + `RadarSite` (lon/lat/height/nod/plc/wmo) → EDR `Location`s: Point geometry, id = NOD code (e.g. `fivih`), label = PLC.
- `get_parameters()` — the ODIM quantities (DBZH, TH, VRADH, ZDR, RHOHV…), **without** the `<site>:` prefix the MapEngine layer naming uses — in EDR the site is the location, so the parameter is just the quantity.
- `query_position` and `query_location` (Point geometry) with **no `z`** — reuse the polar-sampling math already in `volume_engine.rs` (`ground_distance_bearing` + the resampler): nearest site → range/azimuth bin → `PointSeries` over time.
- `query_area` over the network — `CoverageCollection` of per-site `PointSeries`.
- `supported_query_types()` → `["locations", "position", "area"]`.
- Wire `odim-volume` into the EDR registry (`server/src/admin.rs` / `main.rs`) and accept `edr` in the collection `apis` config.
- Widen the FMI S3 collection (`radar-fi-volume-s3-h5`) prefix to all 12 FMI sites so `locations` lists the whole network, not just Vihti.

## Out of scope → M3b

The `z` / elevation-angle selector, the `VerticalProfile` domain, and the no-`z`→profile response shape — all gated on #185.

mrauhala:
Completed and merged. `PolarVolumeEngine` implements the full M3a `EdrEngine` surface in `crates/engine-odim/src/volume_engine.rs`:
- `get_locations()` — radar sites → EDR `Location`s (id = NOD code, antenna position)
- `get_parameters()` — bare ODIM quantities (no `<site>:` prefix)
- `query_position()` (nearest site), `query_location()`, `query_area()` (per-site `CoverageCollection`)
- `supported_query_types()` → `["locations", "position", "area", "trajectory"]`

The FMI S3 collection `collections.d/radar-fi-volume-s3-h5.toml` uses a network-wide `%Y/%m/%d/` prefix, so `locations` lists the whole FMI network. Closing as done.

PR #199, opened 2026-05-18
feat(engine-odim): EDR support for the odim-volume engine (M3a)

mrauhala:
Implements **#196** — the non-vertical slice (M3a) of EDR support for the PVOL `odim-volume` engine, which until now was `MapEngine`-only.

## Summary

`impl EdrEngine for PolarVolumeEngine`:

| Query | Behaviour | CoverageJSON |
|---|---|---|
| `locations` (list) | one location per radar site, id = ODIM NOD code, point geometry at the antenna | Locations `FeatureCollection` |
| `position` | samples the lowest sweep of the radar site **nearest** the requested point | `PointSeries` |
| `locations/{nod}` | same, for a site addressed by its NOD code | `PointSeries` |
| `area` | one coverage per radar site whose antenna falls inside the polygon | `CoverageCollection` of `PointSeries` |

`get_parameters` exposes the ODIM quantities (DBZH, TH, VRADH, ZDR…) **without** the `<site>:` prefix the Map layer naming uses — in EDR the site is the *location* and the quantity is the *parameter*.

Only the lowest elevation sweep is sampled. The vertical (`z`) dimension — turning a no-`z` query into a profile across all sweeps — is **M3b (#197)**, gated on **#185**.

## Supporting changes

- `ds_core::feature::parse_point_coords` — a shared WKT `POINT(lon lat)` / bare `lon,lat` parser, sibling to `parse_area_coords`.
- `sample_sweep_moment` factored out of `polar_sample` so the EDR point query and the Map raster path share one sampling routine (no duplicated polar math).
- `odim-volume` wired into the EDR registry in `server/admin.rs`; both FMI volume collections gain `edr` in `apis`.
- The FMI S3 collection (`radar-fi-volume-s3-h5`) prefix widened from `.../fivih/` to the whole date partition so every FMI radar site is discovered and exposed as a location.

## Test plan

- [x] `cargo test --workspace` — green; new unit tests (`build_site_point_series`, `nearest_site`, `parse_point_coords`) + integration tests exercising `get_locations` / `query_position` / `query_location` / `query_area` against the real FMI Anjalankoski volume (skip gracefully when the fixture is absent)
- [x] CoverageJSON schema validation — new `covjson_validation.rs` case for the blank-unit multi-quantity `PointSeries` the engine emits
- [x] End-to-end: server `/edr/collections/radar-fi-volume-local-h5/{locations,position,area}` return schema-shaped Coverage / CoverageCollection with real radar moments

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -130,7 +130,8 @@ Currently implemented: `PointSeries`, `Grid`.
 | GeoJSON | `FeatureEngine` | Features |
 | GeoTIFF | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
 | GRIB | `EdrEngine` + `MapEngine` | EDR, WMS, Maps, Tiles |
-| ODIM | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
+| ODIM COMP | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
+| ODIM PVOL | `EdrEngine` + `MapEngine` | EDR (position, locations, area), WMS, Maps, Tiles |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
 | PostGIS | `EdrEngine` + `FeatureEngine` | EDR (position, locations, area), Features |
 
--- a/collections.d/radar-fi-volume-local-h5.toml
+++ b/collections.d/radar-fi-volume-local-h5.toml
@@ -13,7 +13,7 @@ id = "radar-fi-volume-local-h5"
 title = "FMI — radar polar volumes (local, ODIM_H5)"
 description = "Finnish radar polar volumes (ODIM_H5 PVOL) from local sample files — every elevation sweep and dual-polarisation moment."
 engine_type = "odim-volume"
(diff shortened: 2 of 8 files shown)


mrauhala:
### Review receipt — `b00144e`

All three findings addressed:

1. **`.expect()` on the `nearest_site` key (volume_engine.rs)** — replaced with `.ok_or_else(|| DataServerError::Engine(…))?`. The invariant still holds (`nearest_site` returns a key it read from `by_site`), but a graceful error beats a panicking request thread if a refactor ever breaks it.

2. **`query_area` only skipped `LocationNotFound`** — valid catch. Now also skips `InvalidParameter`: a heterogeneous network mixes single- and dual-pol radars, so a quantity filter like `ZDR` legitimately matches some sites and misses others — that must not 400 the whole area query. A genuinely bogus parameter still surfaces (every site skips → empty collection → 404).

3. **`parse_point_coords` `(lat, lon)` vs grib's `(lon, lat)`** — the function deliberately matches the `(lat, lon)` order already used by `engine-odim`'s `edr.rs` and `engine-geotiff` (2 of the 3 existing parsers). Rather than re-order it, the doc comment now states the return order in bold and explicitly warns that `engine-grib`'s local `parse_coords` uses the opposite order, so a future migration accounts for the swap. Migrating grib itself is out of scope for this PR.

PR #200, opened 2026-05-18
feat: vertical (elevation/level) dimension for MapEngine + EdrEngine

mrauhala:
## Summary

Implements #185 — a single cross-cutting vertical dimension shared by both the Map and EDR surfaces.

- **ds-core**: new `VerticalKind` enum + `VerticalDimension` descriptor (`crates/core/src/vertical.rs`). `MapEngine::get_raster_tile` gains `z: Option<f64>`; `RasterInfo` gains `vertical`. `EdrEngine` query methods gain `z: Option<&[f64]>`, gain `get_vertical_extent()`, and now return a unified `CoverageResponse` (renamed from `AreaQueryResult`). New `VerticalProfile` CoverageJSON domain plus optional `z` axes on `PointSeries`/`Grid`.
- **api-wms**: parses the WMS 1.3.0 `ELEVATION` dimension and emits `<Dimension name="elevation">` in GetCapabilities.
- **api-maps / api-tiles**: accept an `elevation` query parameter; advertise the vertical extent in collection metadata; OpenAPI updated.
- **api-edr**: parses the `z` query parameter, advertises the vertical extent, emits a `z` axis / `VerticalCRS` referencing, and follows the issue's EDR query→CoverageJSON response table.
- **engine-odim `odim-volume`**: drops the lowest-sweep interim — `z` selects the elevation sweep for rendering; a no-`z` EDR query returns a `CoverageCollection` of `VerticalProfile`s (reflectivity vs. elevation angle), a `z` query a `PointSeries` per sweep.
- Engines with no vertical dimension ignore the selector; the API layer returns HTTP 400 when `z`/`ELEVATION` is passed against them.

## Test plan

- [x] `cargo test --workspace` — 47/47 test binaries pass
- [x] `cargo clippy --workspace --all-targets` — clean
- [x] New CoverageJSON schema-validation tests for `VerticalProfile`, `Grid`-with-`z`, `PointSeries`-with-`z`, and a `VerticalProfile` `CoverageCollection`
- [x] `engine-odim` integration tests updated to assert the new PVOL EDR shapes

## Notes for operators

`CoverageCollection` responses no longer carry a top-level `referencing` array — each coverage's domain now carries its own (a `VerticalProfile` needs a `VerticalCRS` reference that a `PointSeries` does not, so a collection may mix domain shapes). The CoverageJSON schema permits this. Any client reading `collection.referencing` directly should read it per-coverage at `coverages[].domain.referencing` instead.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -107,7 +107,7 @@ All CoverageJSON output **must** validate against the OGC CoverageJSON 1.0 schem
 - **Coverage**: requires `type` ("Coverage"), `domain`, `parameters`, `ranges`
 - **Domain**: requires `type` ("Domain"), `axes`, `referencing`. `domainType` triggers axis constraints.
 - **PointSeries**: x/y are single-value axes, t is string-values axis, z is an optional single-value axis.
-- **Grid**: x/y are numeric-values axes, t and z optional. NdArray shape: `[t, y, x]` or `[y, x]`.
+- **Grid**: x/y are numeric-values axes, t and z optional. NdArray shape: `[t, y, x]`, `[y, x]`, `[z, y, x]`, or `[t, z, y, x]`.
 - **VerticalProfile**: x/y single-value, z numeric-values axis, t optional single-value. NdArray shape: `[z]`.
 - **Parameter**: requires `observedProperty` with `label` as i18n object `{"en": "..."}`.
 - **NdArray**: requires `shape` and `axisNames` when values has >1 item. `values.length` must equal product of `shape`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -106,8 +106,9 @@ All CoverageJSON output **must** validate against the OGC CoverageJSON 1.0 schem
 
 - **Coverage**: requires `type` ("Coverage"), `domain`, `parameters`, `ranges`
 - **Domain**: requires `type` ("Domain"), `axes`, `referencing`. `domainType` triggers axis constraints.
-- **PointSeries**: x/y are single-value axes, t is string-values axis. No additional axes allowed.
-- **Grid**: x/y are numeric-values axes, t is optional. NdArray shape: `[t, y, x]` or `[y, x]`.
+- **PointSeries**: x/y are single-value axes, t is string-values axis, z is an optional single-value axis.
+- **Grid**: x/y are numeric-values axes, t and z optional. NdArray shape: `[t, y, x]` or `[y, x]`.
+- **VerticalProfile**: x/y single-value, z numeric-values axis, t optional single-value. NdArray shape: `[z]`.
 - **Parameter**: requires `observedProperty` with `label` as i18n object `{"en": "..."}`.
 - **NdArray**: requires `shape` and `axisNames` when values has >1 item. `values.length` must equal product of `shape`.
 - **i18n objects**: Keys must be BCP 47 language tags (e.g. `"en"`).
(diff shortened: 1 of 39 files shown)


mrauhala:
Addressed all six review comments in 08bbdf8:

- **`inf`/`nan` parsing (api-edr `parse_z`, api-maps/api-tiles/api-wms `elevation`)** — all four parsers now reject non-finite values via `.filter(|v| v.is_finite())`, so a `z`/`ELEVATION` of `inf`/`nan` returns 400 instead of aliasing to `i64::MAX` in the cache key or breaking `nearest_sweep`'s NaN comparison.
- **`coverage_response_to_json` domainType** — added a `debug_assert!` that all coverages in a collection share one domain type.
- **`finalize_single_site`** — added a `debug_assert_eq!` making the one-coverage-per-single-level invariant explicit.

`cargo test` + `cargo clippy --all-targets` clean across all changed crates.

mrauhala:
Second review round addressed in 23694b2:

- **`vrs`** — dropped from the vertical extent objects in EDR / Maps / Tiles metadata. A human-readable label is not a vertical-CRS identifier and no standard OGC URI exists for radar elevation angle; `vrs` is optional, so it is now omitted rather than emitting an unresolvable placeholder.
- **`finalize_single_site`** — collapsed to a `map_or`, removing the separate dead-code fallback branch (the `debug_assert_eq!` of the one-coverage invariant is kept).
- **Collection-level `referencing` removal** — documented in the test (`coverage_collection_validates`) and added a "Notes for operators" section to the PR description.

`cargo fmt --all --check`, `cargo clippy --all-targets`, and `cargo test` are clean across the changed crates. The earlier CI failure was a formatting miss on the first review-fix commit, fixed here.

mrauhala:
Third review round addressed in 85341fe:

- **`finalize_single_site`** — now `expect`s the one-coverage invariant instead of a `debug_assert` + empty-collection fallback, so a future contract violation fails loudly in release rather than silently returning an empty `CoverageCollection`.
- **`parse_z`** — an empty list element (stray comma, e.g. `z=850,`) returns a clear "empty element" error; doc comment notes the `min/max` interval form is unsupported.
- **WMS `ELEVATION`** — a comma-separated list returns an explicit "single value only" error rather than the misleading "not a finite number".
- **`quantize_z`** — gains `debug_assert!(z.is_finite())` plus a contract note in its doc.

`cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` clean on the changed crates.

mrauhala:
Fourth review round addressed in 755a8ce:

- **`coverage_response_to_json`** — a collection-level `domainType` is now emitted only when every coverage agrees; a heterogeneous collection omits it entirely (each coverage's domain still carries its own `domainType`). This removes the debug-only assertion and makes it impossible to emit a `domainType` that mismatches some coverages in release.
- **Maps/Tiles `elevation` docs** — corrected to "Rejected with HTTP 400 for collections with no vertical dimension" to match the handler behaviour.
- **EDR `z` OpenAPI description** — now documents that requested values snap to the nearest advertised level and that the response domain reports the snapped level (discoverability gap from the round-3 note).

`cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` clean on the changed crates.

mrauhala:
Fifth review round addressed in c6524b9:

- **`finalize_single_site`** — now returns `Result`; a missing coverage surfaces as a logged `DataServerError::Engine` (clean 500) instead of an opaque `spawn_blocking` panic.
- **`coverage_response_to_json`** — hoists the union of every coverage's parameters to collection level, so a mixed-parameter collection would advertise them all rather than only the first coverage's. This removes the first-coverage-only assumption without a debug-only assertion.
- **MULTIPOINT + vertical collections** — the EDR `coords` (point) OpenAPI description now documents that per-point coverages are flattened into one `CoverageCollection`.

`cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` clean on the changed crates.

All substantive findings from rounds 1–5 are resolved and CI (Lint & Test, claude-review, Docker) is green. The remaining review surface has converged to defensive-assertion styling and exotic-combination documentation; this PR is ready for review/merge from my side.

mrauhala:
Sixth review round addressed in 93a4777: the `CLAUDE.md` Grid NdArray shape note now lists the `z`-present forms (`[z, y, x]`, `[t, z, y, x]`).

mrauhala:
Seventh review round addressed in 15dd9d7:

- **`resolve_quantities`** — now unions moments across *every* sweep of the most-recent volume (a `VerticalProfile` samples all sweeps; a quantity may exist only on higher elevations, e.g. dual-PRF `VRADH`/`WRADH`) rather than only the lowest sweep's.
- **`VerticalDimension`** — dropped the stored `label`/`unit` `String` fields; they are now derived from `VerticalKind` via accessors, so the CoverageJSON `VerticalCRS` axis description and the advertised collection metadata share one source of truth and cannot diverge.
- **z-rejection coverage** — added per-API-surface tests asserting HTTP 400 for `z`/`elevation`/`ELEVATION` against a collection with no vertical dimension (EDR, Maps, Tiles, WMS).

`cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` clean on the changed crates.

PR #218, opened 2026-05-20
ci: track every workspace crate in release-please for proper changelog attribution

mrauhala:
Closes the changelog-attribution gap that left the TM35FIN coarse-grid projection change (#214) out of the v0.5.0 release notes ([#215](https://github.com/mrauhala/meteocore/pull/215)).

## Why

release-please attributes commits to packages **strictly by file path inside the package directory**. Until now only `crates/server` was configured as a release-please package, so a commit confined to other crates was attributed to nothing release-tracked and silently dropped from the release notes. Verified by reading the release-please-action log on #214's merge: it filtered the 2 release-triggering commits, called `Updating existing candidate pull request for server, path: crates/server`, and concluded `PR #215 remained the same` because #214's diff lives in `crates/engine-geotiff` + `crates/core`.

The earlier engine-odim entries in v0.4.0's CHANGELOG slipped through only because each one **also** touched `crates/server` (engine registration in `server/main.rs`). Pure within-an-engine changes — exactly what perf optimisations look like — have no such accident.

## What

- **Every non-server crate's `Cargo.toml` is bumped from 0.1.0 to 0.4.0** so the workspace sits at one synchronised baseline.
- **All 17 workspace crates are added to `release-please-config.json::packages`**, each with its own `component` + `package-name` and a `bootstrap-sha` pinned to the v0.4.0 release commit (`fb12a38`) so release-please does not scan pre-0.4.0 history for the newly-tracked crates.
- **`linked-versions` plugin added** so every crate moves to the same version on each release. Combined with the existing `cargo-workspace` plugin's `merge: true`, the per-crate release PRs collapse into one PR with one combined CHANGELOG-summary, one bumped version, and one `vX.Y.Z` tag (no per-component tags, since `include-component-in-tag` stays false).
- Each crate will grow its own `crates/<name>/CHANGELOG.md` on the next release for its own commits; `crates/server/CHANGELOG.md` continues to summarise server-touching changes.

`cargo build --workspace` passes; `cargo.lock` updated.

## Caveats

- **The next release PR (#215, when re-generated by release-please after this merges) is the live test of the configuration.** The `release-please` guard step added in #216 fails loudly if a release PR merges without a release being cut, so any breakage surfaces immediately rather than silently.
- 17 components mapped to a single component-less `v0.5.0` tag is an unusual setup — `linked-versions` + `cargo-workspace { merge: true }` should produce one combined release, but if release-please refuses, this commit is straightforward to revert and we fall back to amending v0.5.0's notes by hand.
- Merging this PR does **not** retroactively add #214 to #215. After this merges I recommend re-running the latest release workflow (or letting the next push trigger it) so release-please regenerates #215 with the perf section now attributed.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,3 +1,19 @@
 {
+  "crates/core": "0.4.0",
+  "crates/storage": "0.4.0",
+  "crates/render": "0.4.0",
+  "crates/ds-mvt": "0.4.0",
+  "crates/engine-csv": "0.4.0",
+  "crates/engine-geojson": "0.4.0",
+  "crates/engine-geotiff": "0.4.0",
+  "crates/engine-grib": "0.4.0",
+  "crates/engine-odim": "0.4.0",
+  "crates/engine-postgis": "0.4.0",
+  "crates/engine-querydata": "0.4.0",
+  "crates/api-edr": "0.4.0",
+  "crates/api-features": "0.4.0",
+  "crates/api-wms": "0.4.0",
+  "crates/api-maps": "0.4.0",
+  "crates/api-tiles": "0.4.0",
   "crates/server": "0.4.0"
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,7 +66,7 @@ checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "api-edr"
-version = "0.1.0"
+version = "0.4.0"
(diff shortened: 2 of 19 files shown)


mrauhala:
Closing in favour of the usual single-binary-monorepo setup (setup #1): keep release-please tracking only `crates/server`, and amend release notes by hand for the rare change that lands entirely outside the server crate. The 17-component → single-tag configuration here is more machinery than this project needs. The v0.5.0 case (missing #214) will be handled by amending the release notes + CHANGELOG after release.

Issue #222, opened 2026-05-24
perf(storage): DataStore::block_on creates a new Tokio Runtime per call from rayon tile pool (remote GeoTIFF)


mrauhala:
## Summary

`ds-storage`'s sync→async bridge falls back to **creating a brand-new Tokio `Runtime` on every call** when it is invoked from a thread that has no current Tokio runtime handle — which is exactly what happens when the GeoTIFF engine fetches remote tiles from its **rayon** thread pool. Each `Runtime::new()` spawns a fresh set of OS worker threads, uses them for one byte-range read, then tears them down.

## Mechanism (code-certain)

`DataStore::block_on` (`crates/storage/src/lib.rs:94-117`):
```rust
match tokio::runtime::Handle::try_current() {
    Ok(handle) => { tokio::task::block_in_place(|| handle.block_on(timed)) }   // line 109
    Err(_)     => { let rt = tokio::runtime::Runtime::new()?; rt.block_on(timed) } // line 114
}
```

Remote GeoTIFF tile reads run under a rayon pool, not the Tokio runtime:
- `crates/engine-geotiff/src/reader.rs:1276` defines `static TILE_FETCH_POOL: LazyLock<rayon::ThreadPool>`.
- `reader.rs:1712` `TILE_FETCH_POOL.install(|| tiles.par_iter().map(|t| read_remote_chunk_with_retry(... read_remote_chunk_f64(...))))`.
- `read_remote_chunk_f64` (`reader.rs:988`) calls `store.get_range()` → `DataStore::block_on`.

Because rayon worker threads have **no** current Tokio handle, `try_current()` returns `Err` and the `Runtime::new()` branch fires — **once per tile, per render** for remote GeoTIFF collections (e.g. `opera-reflectivity`, `met-norway-radar`, `fmi-radar`). Standing up and tearing down a multi-threaded runtime per tile is large, pointless overhead, and it competes for CPU/threads with the main runtime under concurrent WMS load.

The local `fmi-radar-composite-dbz` collection is **not** affected: `read_bbox_inner` takes the sequential (non-rayon) branch for local files, so this path never fires there.

## Fix

Capture the main runtime `Handle` before entering the rayon pool and use it inside the closures so `block_on` runs on the existing runtime instead of `try_current()` failing:
```rust
let handle = tokio::runtime::Handle::current();
TILE_FETCH_POOL.install(|| tiles.par_iter().map(|t| {
[...]

Issue #221, opened 2026-05-24
perf(server): poll/scan loops block the shared Tokio runtime → periodic multi-second WMS p99 spikes at low load


mrauhala:
## Summary

Production shows **periodic multi-second WMS p99 latency spikes even at low request load**. After the #203 coarse-grid projection fix the median is fast (~0.09 s), but Prometheus shows WMS p99 jumping from a ~0.3 s baseline to **~3.7–4.85 s** every few minutes, sometimes spanning 1–2 minutes (back-to-back per-minute buckets). The histogram confirms it: **4,059 requests >5 s and 25,292 >1 s out of 1.48 M**. The render semaphore is fully free at scrape — these are not queueing under load.

Root cause: **engine poll/scan loops run blocking network I/O directly on Tokio worker threads** (no `spawn_blocking`). `ds-storage`'s sync bridge uses `block_in_place(|| handle.block_on(...))` (`crates/storage/src/lib.rs:107-110`), which parks the *current worker thread* for the full network round-trip. With only **12 Tokio worker threads** (nproc=12, default) shared by every collection's polls *and* all WMS renders, a poll that blocks one or more workers for seconds starves the pool and stalls unrelated requests — including the fast local radar collection.

## The offenders (code-certain)

**GRIB — worst (CRITICAL).** `GribEngine::poll_loop` calls `self.scan_once()` directly in the `tokio::select!` arm with no `spawn_blocking` (`crates/engine-grib/src/lib.rs:203`). `scan_once()` then does, sequentially on that one worker thread:
- 2–4 × `store.list()` (one per prefix),
- 1 × `store.get(index_file)` per new index,
- `probe_new_parameters()` → up to `MAX_PROBES_PER_SCAN = 32` sequential `fetch_grid_by_entry()` calls (`lib.rs:529-533`), each doing `store.head()` + `store.get_range()` (`crates/engine-grib/src/reader.rs:30,49`).

That's **~35 sequential `block_in_place` network round-trips on one worker**: at 100 ms/RTT ≈ 3.5 s, at 200 ms ≈ 7 s, at 500 ms ≈ 17 s — matching the observed 3.7–4.85 s spikes. Fires when a new run/step is discovered (ECMWF IFS / GFS, ~6 h runs with steps trickling in).

[...]

Issue #211, opened 2026-05-19
perf(engine-geotiff): raster_info() allocates fresh Vec<DateTime> on every Maps/Tiles request


mrauhala:
Part of #201 (GeoTIFF radar WMS performance epic).

## Problem

The Maps and Tiles handlers call `engine.raster_info()` **unconditionally**, before the render semaphore is acquired. For GeoTIFF, `raster_info()` clones every catalog timestamp into a fresh `Vec<DateTime<Utc>>` on each call.

## Evidence

- `raster_info()` allocation: `crates/engine-geotiff/src/lib.rs:1458` — `catalog.entries.keys().cloned().collect()`
- Unconditional call sites: `crates/api-maps/src/handlers.rs:724`, `crates/api-tiles/src/handlers.rs:1079`
- WMS guards both its calls behind `layer_parameter.is_some()` / `elevation.is_some()` checks, so the WMS hot path is unaffected.

## Impact

LOW at current traffic — the local-GeoTIFF allocation is tiny (~24 timestamps) and Maps/Tiles traffic is negligible (203 tile requests vs 61,608 WMS). For STAC-backed collections, `raster_info()` can additionally trigger a catalog HTTP fetch on the async thread. Worth fixing if Maps/Tiles adoption grows.

## Suggested fix

Cache `RasterInfo` on the engine in an `ArcSwap<Option<Arc<RasterInfo>>>`, regenerated after each catalog swap (`poll_once`); return a cheap `Arc` clone from `raster_info()`. For STAC, ensure the path never blocks an async worker.

## Labels
`priority: low`, `performance`, `effort: small` — milestone v0.3

_Audit: hunter-concurrency, survived devil's-advocate review (low — not on the WMS hot path)._

Issue #206, opened 2026-05-19
perf(engine-geotiff): Vec<Option<f64>> decode intermediate inflates memory 16x (~33MB/render)


mrauhala:
Part of #201 (GeoTIFF radar WMS performance epic).

## Problem

All tile decode paths return `Vec<Option<f64>>`. `Option<f64>` is **16 bytes** (8-byte f64 + 1-byte discriminant + 7-byte padding) — a 16× inflation over the 1-byte/pixel U8 radar data on disk.

## Evidence

- `decode_chunk_f64`: `crates/engine-geotiff/src/reader.rs:611-691` (U8 case `reader.rs:651-662` — 1 input byte → 16-byte `Option<f64>`)
- `decode_raw_tile_f64`: `crates/engine-geotiff/src/reader.rs:877-982`
- Output buffer: `crates/engine-geotiff/src/lib.rs:1267` — `Vec::with_capacity(total_pixels)` of `Option<f64>`
- For 1920×1080: 2,073,600 × 16 = **33.2 MB** output buffer alone, plus 4–16 × ~1 MB tile-decode buffers → 37–49 MB allocated/freed per render. At 3 concurrent renders: 111–147 MB cycling through the allocator simultaneously.

## Impact

Large per-render allocations under concurrent load create allocator pressure (glibc per-arena serialization) — a plausible contributor to fat-tail latency. Companion to the `IntegerLutColorMap` issue (#201).

## Suggested fix

Carry native integer types (`u8`/`u16`/`i16`) through the decode + resample pipeline instead of `Option<f64>`; represent nodata with the source sentinel rather than `Option`. For U8 radar this shrinks the output buffer from 33 MB to ~2 MB and removes the 16× amplification. Pairs naturally with wiring in `IntegerLutColorMap` (apply the colormap in-place on the integer buffer).

## Labels
`priority: medium`, `performance`, `effort: medium` — milestone v0.3

_Audit: hunter-render, survived devil's-advocate review._

mrauhala:
[...]

PR #223, opened 2026-05-24
docs: add Engine Performance & Concurrency Rules to CLAUDE.md

mrauhala:
Distills the generic, reusable lessons from the radar WMS performance epic (#201) and the p99-spike investigation (#221/#222) into a new CLAUDE.md section, so new engines and rendering features don't repeat them.

Covers: poll loops must `spawn_blocking` (shared runtime; #221/#208, ODIM as reference), `ds-storage` sync-bridge misuse incl. the rayon `Runtime::new`-per-call trap (#222), coarse-grid projection vs per-pixel (#203), cache-key-must-match-access-pattern (#202), compact decode types (#206/#207), O(1) hot-path metadata accessors (#211), and a note to check spike magnitude before blaming a mechanism.

Docs-only.


mrauhala:
Addressed both:
- **Counter-examples implied fixed** — reworded the rule: the real #221 mechanism is the dedicated background runtime (`poll_runtime()`), not per-engine `spawn_blocking` (which would panic via `ds-storage`'s `block_in_place` on a blocking-pool thread). Now states explicitly that grib/geotiff/querydata still block directly and are made safe only by running on that runtime — so no reader assumes they were individually converted.
- **Hardcoded line number** — replaced `engine-grib/src/lib.rs:203` with the function name (`scan_once` inside `poll_loop`).

PR #225, opened 2026-05-24
perf(render): emit lossless WebP (#224)

### LINKED ISSUE #224: render: switch WebP output from lossy to lossless
## Summary

`encode_webp` (`crates/render/src/encode.rs:82`) currently emits **lossy** WebP at quality 80 (`webp::Encoder::from_rgba(...).encode(80.0)`). Switch it to **lossless**.

## Why

The rendered tiles are colormapped raster (radar dBZ classes, NWP fields) with hard color boundaries and a small palette. Lossy WebP introduces ringing/blocking artifacts on those sharp class edges and can shift colors (misrepresenting data values), while palette-like imagery compresses very well losslessly. Lossless gives pixel-exact output that matches the PNG path, at comparable or better size for this kind of content.

## Change

In `crates/render/src/encode.rs`:
```rust
let encoder = webp::Encoder::from_rgba(rgba, width, height);
let memory = encoder.encode_lossless();   // was: encoder.encode(80.0)
```
Update the doc comment ("Uses lossy encoding with quality 80…" → lossless). The `webp` crate 0.3 exposes `Encoder::encode_lossless()`.

## Verify
- `cargo test -p ds-render` (the existing `test_encode_webp_valid` RIFF/WEBP header check still holds).
- Add an assertion that re-decoding the lossless WebP yields the exact input RGBA (pixel-exact), to lock the lossless guarantee.

## Labels
`enhancement`, `performance`, `priority: medium`, `effort: tiny` — milestone v0.3.

mrauhala:
## What

Switch WebP image output in `crates/render/src/encode.rs` from lossy (quality 80) to lossless via `Encoder::encode_lossless()`.

## Why

Our raster output is colormapped radar/NWP tiles: hard class boundaries drawn from a small palette. Lossy WebP introduces ringing around those edges and can shift pixels off the palette — which effectively corrupts the encoded data values (a viewer can no longer reliably map a pixel color back to its data class). Lossless preserves every pixel exactly while still compressing the limited palette well.

## Changes

- `encode_webp` now calls `encoder.encode_lossless()` instead of `encoder.encode(80.0)`.
- Updated the doc comment to describe lossless encoding and the rationale.
- Kept the existing `test_encode_webp_valid` (RIFF/`WEBP` header check).
- Added `test_encode_webp_lossless_roundtrip`: encodes a small sharp-edged RGBA buffer (distinct per-channel values + varied alpha including a fully transparent pixel) and decodes it back with `webp::Decoder`, asserting the decoded RGBA equals the input byte-for-byte. No new dependency was needed — the `webp` crate's `Decoder` is available with `default-features = false`.

## Verification

- `cargo test -p ds-render` — 40 passed (incl. the new round-trip test).
- `cargo clippy -p ds-render --all-targets` — clean.
- `cargo fmt` applied.

Closes #224

### DIFF
--- a/crates/render/src/encode.rs
+++ b/crates/render/src/encode.rs
@@ -77,7 +77,11 @@ pub fn encode_jpeg(rgba: &[u8], width: u32, height: u32) -> Result<Vec<u8>, Data
 
 /// Encode an RGBA buffer to WebP bytes.
 ///
-/// Uses lossy encoding with quality 80 for a good size/quality tradeoff.
+/// Uses lossless encoding. Our raster output is colormapped radar/NWP tiles:
+/// hard class boundaries drawn from a small palette. Lossy WebP introduces
+/// ringing around those edges and can shift pixels off the palette, which
+/// effectively corrupts the encoded data values. Lossless preserves every
+/// pixel exactly while still compressing the limited palette well.
 /// WebP supports alpha channel natively, so no transparency compositing needed.
 pub fn encode_webp(rgba: &[u8], width: u32, height: u32) -> Result<Vec<u8>, DataServerError> {
     let expected_len = (width * height * 4) as usize;
@@ -92,7 +96,7 @@ pub fn encode_webp(rgba: &[u8], width: u32, height: u32) -> Result<Vec<u8>, Data
     }
 
     let encoder = webp::Encoder::from_rgba(rgba, width, height);
-    let memory = encoder.encode(80.0);
+    let memory = encoder.encode_lossless();
     Ok(memory.to_vec())
 }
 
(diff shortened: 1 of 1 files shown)


mrauhala:
Addressed in c9d9b9a.

Gave the round-trip test's fully-transparent pixel a non-zero RGB — `(200, 100, 50, 0)` — instead of `(0,0,0,0)`. That immediately surfaced a real defect: libwebp's lossless mode rewrites the RGB of fully transparent regions to improve compression, so `encode_webp` was **not** byte-exact (the pixel decoded back as `(0,0,0,0)`).

Fix: `encode_webp` now mirrors `encode_lossless()` via `encode_advanced` with libwebp's `exact` flag set (`config.exact = 1`), preserving every channel of every pixel including transparent ones. The round-trip test now asserts byte-exact equality across the whole buffer, transparent pixel's RGB included.

- `cargo test -p ds-render` — 40 passed (round-trip green).
- `cargo clippy -p ds-render --all-targets` — clean.
- `cargo fmt` applied.

mrauhala:
Addressed in 3ac239b.

Good catch — `WebPConfig::new()` uses the default preset, which leaves `method = 4` (a lossy-oriented effort level), whereas `encode_lossless()` applies the lossless preset that sets `method = 0` (fastest lossless encoder). The hand-built config was inheriting the slower effort level. Set `config.method = 0` to match `encode_lossless()`, avoiding the extra encode latency under the render semaphore. The round-trip lossless test still passes byte-exact at `method = 0`.

The earlier comment about the transparent pixel's RGB was resolved in c9d9b9a (transparent pixel is now `(200,100,50,0)` and the assertion is byte-exact).

- `cargo test -p ds-render` — 40 passed (round-trip green).
- `cargo clippy -p ds-render --all-targets` — clean.
- `cargo fmt` applied.

Issue #208, opened 2026-05-19
fix(engine-geotiff,engine-querydata): poll_once() blocks Tokio async worker on filesystem I/O


mrauhala:
Part of #201 (GeoTIFF radar WMS performance epic).

## Problem

`GeoTiffEngine::poll_loop` calls `self.poll_once()` **synchronously** inside a `tokio::select!` arm, blocking the async worker thread on filesystem I/O (`std::fs::read_dir` + per-new-file TIFF header reads).

## Evidence

- Sync call in `select!`: `crates/engine-geotiff/src/lib.rs:813-814`
- Blocking readdir: `crates/engine-geotiff/src/catalog.rs:267` (`std::fs::read_dir`)
- Correct pattern for contrast: `crates/engine-odim/src/engine.rs:384-389` wraps the scan in `tokio::task::spawn_blocking`
- Same bug in QueryData: `crates/engine-querydata/src/engine.rs:80-93` — `interval.tick() => self.poll_once()` calls `load_file()` → `QueryData::open()` (mmap of a large `.sqd` file) synchronously.

## Impact

For local directories (~60 files) this is 1–5 ms once per minute — negligible at current load, but the **wrong pattern**. For remote/S3-backed GeoTIFF collections (`StoreMode::Remote`, `lib.rs:433-442`) the inline S3 LIST blocks an async worker thread for the full network round-trip (100–500 ms) — a real reliability bug there.

## Suggested fix

Wrap the `poll_once()` body in `tokio::task::spawn_blocking(...)` in both the GeoTIFF and QueryData engines, matching the ODIM engine pattern.

## Labels
`priority: medium`, `bug`, `reliability`, `effort: small` — milestone v0.3

_Audit: hunter-concurrency, survived devil's-advocate review._

mrauhala:
Confirmed in production as **the** cause of periodic multi-second WMS p99 spikes at low load — see #221, which extends this same bug (poll/scan loop blocking the shared Tokio runtime) to the GRIB engine (the worst offender: ~35 sequential `block_in_place` network round-trips per new-run probe) and adds Prometheus/nexus evidence. Suggest fixing all engines together (geotiff + querydata here, grib in #221), using the ODIM engine's `spawn_blocking` pattern (engine-odim/src/engine.rs:389) as the reference.

mrauhala:
[...]

Issue #201, opened 2026-05-19
Epic: GeoTIFF radar WMS performance — refresh baseline and remaining bottlenecks


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep as an epic; prioritize remaining children.

**Current assessment and challenge:** Many GeoTIFF/render fixes landed. Old aggregate p95 is not a current baseline; rank surviving child problems independently.

The original roadmap below is historical context; shipped phases should not be counted as new work or as independent priority candidates. Remaining child issues retain their individual implementation and verification scope.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

epic: GeoTIFF radar WMS performance — p95 1–5s above 10 req/s

Epic tracking the GeoTIFF radar WMS performance work (p95 1–5 s above ~10 req/s) plus the follow-on spike investigation.

> **Triage 2026-06-04:** #204, #207, #246 verified FIXED in current code and already closed — checklist updated below. Remaining round-1 work: **#205, #206, #209, #210**. Added round-3 meta-tile-assembly findings #240 / #241 (newer; not previously tracked here).

### GeoTIFF render audit (round 1)
- [x] #202 — Rendered-image cache wastes 518 MB at ~3% hit rate (→ internal meta-tiling)
- [x] #203 — Per-pixel TM35FIN projection in the resampling loop (→ coarse-grid resampler, v0.5.0)
- [x] #204 — Per-request `File::open` + 8 KB BufReader + IFD re-parse (local COG) → mmap cache (`Arc<OnceLock<Mmap>>`)
- [ ] #205 — Cold-file disk reads on radar file rotation (the original 172 reqs >5 s)
- [ ] #206 — `Vec<Option<f64>>` decode intermediate inflates memory 16×
- [x] #207 — `IntegerLutColorMap` built but never wired into WMS/Maps/Tiles → wired via `maybe_wrap_integer_lut`
- [x] #208 — `poll_once()` blocks a Tokio async worker on filesystem I/O
- [ ] #209 — Make `render_concurrency` configurable (hardcoded 2×CPU)
- [ ] #210 — `copy_tile_to_result` per-pixel loop blocks auto-vectorization
[...]

PR #226, opened 2026-05-24
perf(server,engine-grib): isolate poll loops from request runtime + skip settled GRIB runs (#221)

mrauhala:
Fixes the periodic multi-second WMS p99 spikes at low load (#221, epic #201).

## Root cause
Collection poll loops do blocking I/O on the **shared request-serving Tokio runtime**: `ds-storage`'s sync bridge parks the worker thread via `block_in_place` for the whole network round-trip (storage/src/lib.rs:109), and the GRIB new-run probe does up to ~35 sequential byte-range reads. With only 12 worker threads, overlapping polls starve the pool and stall unrelated WMS renders — Prometheus showed ~0.3 s baseline p99 jumping to ~4–5 s every few minutes.

## Changes
1. **Dedicated background runtime for poll loops** (`poll_runtime()` in `server/src/main.rs`; used by startup *and* the `admin.rs` reload path). Poll `block_in_place` now parks background threads, never request workers.
   - Note: the obvious "wrap `scan_once` in `spawn_blocking`" would **panic** — `block_in_place` is illegal on a blocking-pool thread — so a separate multi-thread runtime is the correct fix.
2. **GRIB poll default 300 s → 600 s** (NWP models publish ~6-hourly; some less often). Per-collection `poll_interval_secs` still overrides for the rare faster model.
3. **GRIB skips settled run prefixes.** Listing a run prefix returns all its (hundreds of) step files. Once a newer run exists, an older run is static (runs publish sequentially), so each scan now lists only unknown prefixes (new / not-yet-published) plus the single newest run still gaining steps. `settle_completed_runs` settles all-but-newest of the runs listed with hits and prunes to the scan window — unit-tested for the multi-scan evolution (new-run appears, prune, newest-stays-active).

## Tests
`cargo test -p ds-core -p engine-grib` green (incl. 3 new `settle_*` tests); `cargo build -p server` green; clippy clean; fmt applied.

This is the cross-engine fix; ODIM already used `spawn_blocking` correctly and is moved onto the background runtime too (neutral). Geotiff/querydata poll_once (previously #208) are fixed by the runtime isolation as well.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -382,7 +382,11 @@ pub struct OdimConfig {
 }
 
 fn default_grib_poll_interval() -> u64 {
-    300
+    // 10 minutes. NWP models typically publish a new run every 6 h (some less
+    // often), with steps trickling in over ~1-2 h, so frequent polling mostly
+    // re-lists unchanged runs. Override per collection via `poll_interval_secs`
+    // for the rare model that updates more often.
+    600
 }
 
 fn default_grid_cache_mb() -> u64 {
@@ -401,7 +405,7 @@ pub struct GribConfig {
     pub index_suffix: Option<String>,
     /// Suffix for GRIB data files. Default: ".grib2"
     pub data_suffix: Option<String>,
-    /// Poll interval in seconds. Default: 300 (5 min)
+    /// Poll interval in seconds. Default: 600 (10 min)
     #[serde(default = "default_grib_poll_interval")]
     pub poll_interval_secs: u64,
     /// Maximum number of forecast runs to retain. Default: 4
--- a/crates/engine-grib/src/lib.rs
+++ b/crates/engine-grib/src/lib.rs
@@ -305,8 +305,6 @@ impl GribEngine {
             .is_none_or(|t| t.elapsed() >= SETTLED_REVALIDATE_INTERVAL);
         let settled_snapshot = self.settled_prefixes.lock().unwrap().clone();
         let mut listed_with_hits: Vec<String> = Vec::new();
-        let mut any_list_ok = false;
-        let mut any_list_err = false;
         for (_ref_time, prefix) in &prefixes {
             if let Some(budget) = scan_budget {
                 if runs_with_hits >= budget {
@@ -322,7 +320,6 @@ impl GribEngine {
             let mut hits_in_this_prefix = 0usize;
             match self.store.list(&obj_prefix) {
                 Ok(objects) => {
-                    any_list_ok = true;
                     for obj in objects {
                         let loc = obj.location.as_ref();
                         if !loc.ends_with(index_suffix) {
@@ -338,7 +335,6 @@ impl GribEngine {
                     }
                 }
                 Err(e) => {
-                    any_list_err = true;
                     tracing::debug!(
                         "Collection '{}': failed to list prefix '{}': {}",
                         self.collection_id,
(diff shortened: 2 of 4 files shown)


mrauhala:
Addressed the review (commit a4b80f5):

- **[confirmed bug] settled skip hides late/added step files** — added a periodic full re-validation: every `SETTLED_REVALIDATE_INTERVAL` (1 h) a scan ignores `settled_prefixes` and re-lists all run prefixes, so late/corrected steps appended to an already-scanned run are picked up within that bound, while the per-poll re-list is still skipped the rest of the time. Note: a same-key *content* rewrite of an existing index file is a **pre-existing** limitation of the path-keyed `known_indexes` dedup (it skips any path it has already seen) — not introduced by this PR; documented on the new const, can be a follow-up (ETag/size check) if needed.
- **[shutdown] leaked OnceLock runtime** — documented the rationale: on exit poll tasks are aborted, which is safe here because a poll only reads the read-only store and atomically swaps an in-memory `ArcSwap` catalog (no partial/corrupt state; rebuilt on next start), and the loops also honor the per-engine shutdown watch channel. No risky graceful-drain refactor of a `'static` runtime.
- **[reload contention] 2 workers** — bumped `poll_runtime` to 4 worker threads for headroom; also noted `block_in_place` spins up temporary replacement workers while a poll blocks, so concurrent polls aren't hard-capped at the base count.

`cargo test -p ds-core -p engine-grib` green; clippy clean; fmt applied.

mrauhala:
Round 2:

- **[moderate] timer reset before listing** — fixed (commit just pushed). `last_full_scan` is now only *read* up front; it is reset after the listing loop, gated on `force_full && any_list_ok`, so a total S3 outage retries the forced full scan next poll instead of waiting another interval.
- **[low] reload runs old + new poll loops concurrently for up to `poll_interval_secs`** — the "up to 600 s" bound is not accurate: the poll loops `select!` on `rx.changed()` (a watch channel) alongside the sleep, so on a reload signal an old loop exits at its next `select!` iteration — immediately if idle, or right after the in-flight `scan_once()` completes (seconds), **not** after the full sleep. So the concurrent window is at most one in-flight scan, and the background runtime (4 workers + `block_in_place` replacement workers) absorbs it. Leaving as-is rather than adding JoinHandle tracking for a transient, bounded overlap.

The earlier re-cited items (settled-skip / leaked-runtime shutdown) were addressed in the previous round (commit a4b80f5) — see the prior receipt.

mrauhala:
Round 3:

- **settle_completed_runs newest-first contract** — made explicit: renamed the parameter to `listed_with_hits_newest_first` and documented that the ordering is load-bearing (first = active newest, rest settled), so a future reorder can't silently invert settling. (A `debug_assert!` on ordering isn't reliable here — the function only receives prefix strings, and for `{run}`-hour patterns lexical order ≠ time order; the rename/doc is the robust guard.)
- **hardcoded `worker_threads(4)`** — leaving as-is for this PR. `block_in_place` lets Tokio spin up replacement workers while a poll blocks, so 4 is the steady-state base, not a hard cap on concurrent polls; with the new 10-min GRIB interval + settled-skip the per-poll work is also much smaller. A `[server] poll_worker_threads` knob is a reasonable future enhancement — happy to file a follow-up issue if you'd like it, but it's tunability rather than a defect.

mrauhala:
Round 4:

- **[new code] partial S3 outage still reset the clock** — fixed. Now tracking `any_list_err`; the re-validation clock resets only when `force_full && any_list_ok && !any_list_err`, i.e. a forced full scan in which **every** list call succeeded. Any list error (total or partial) leaves the clock unset so the full scan retries next poll. No settled prefix can be skipped past its interval because of an errored sibling in the same scan.
- **[pre-existing] failed reload kills old poll loops** — confirmed pre-existing (the reload signals old engines to shut down before `load_collections`, so a zero-collection reload leaves no poll loops). Out of scope for this perf PR; filed as **#227** with a fix sketch (build/spawn new before signalling old).

mrauhala:
Round 5:

- **[bug, regression I introduced in round 4] a single persistently-failing prefix pins `force_full` on forever** — fixed by simplifying. The escalating conditions on the clock reset (round 2: not on total outage; round 4: not on partial outage) were over-fitting and created a worse failure mode (chronic bad prefix ⇒ full scan every poll, optimization defeated). The clock reset is now **unconditional** on a forced scan. Rationale (documented in code): the clock only *paces* a best-effort ~hourly re-list of settled runs; a transient outage during that one scan merely defers re-validation to the next interval — harmless, because the **newest (unsettled) run is listed every poll regardless**, so steady-state freshness never depends on the hourly pass. The `any_list_ok`-on-empty-Ok point is moot with the unconditional reset.

I'm treating the re-validation timer as settled at this point: it's a best-effort hourly safety net layered on top of the user-requested "skip settled runs" behavior, and further micro-edge-cases on its pacing (transient-outage timing) are below the threshold of a meaningful defect for this PR. The pre-existing reload gap is tracked in #227.

mrauhala:
Round 6: both findings are **pre-existing reload-lifecycle bugs, not introduced by this PR** (it changes *where* poll loops are spawned — onto the background runtime — not the `subscribe()`/shutdown-signal pattern, which is unchanged):

- **Bug 2 (shutdown before load success)** — already tracked in **#227**.
- **Bug 1 (watch-channel pre-subscription race)** — `subscribe()` runs inside the spawned task, so a shutdown sent before the task starts is marked "seen" and never observed. Same reload-lifecycle family; folded into **#227** (the fix for both is to make the reload swap atomic: subscribe `rx` synchronously before spawning, build/spawn new before signalling old).

Neither is a defect in this PR's actual change (runtime isolation + GRIB settled-skip), which the review confirms is "sound and correct." Keeping #226 scoped to the perf fix; the reload-lifecycle hardening lands in #227.

I consider #226 converged: the remaining review findings are pre-existing issues tracked separately, not regressions in this change.

PR #232, opened 2026-05-24
perf(storage,engine-geotiff): reuse runtime for remote tile fetches instead of Runtime-per-tile (#222)

### LINKED ISSUE #222: perf(storage): DataStore::block_on creates a new Tokio Runtime per call from rayon tile pool (remote GeoTIFF)

mrauhala:
Closes #222.

## Problem
Remote GeoTIFF tile reads run on the rayon `TILE_FETCH_POOL`. Rayon workers have **no current Tokio handle**, so `DataStore::block_on` (object_store path) and `block_on_async` (HTTP path) fall through to `Runtime::new()` — constructing and tearing down a whole multi-thread Tokio runtime **per tile fetch**, for every remote render (opera-*, met-norway-radar, fmi-radar S3). Thread/alloc churn that grows with render traffic.

## Fix
Capture the current runtime `Handle` before entering the rayon pool and thread it down so each fetch runs via `handle.block_on` on the existing runtime — no per-call runtime, and no `block_in_place` (which would *panic* off a worker thread; that subtlety is why the issue's original 'enter the handle' suggestion wouldn't work).

- **ds-storage:** new `DataStore::get_range_on(path, range, &Handle)`; `block_on` refactored into `block_on_with(Option<&Handle>, fut)` (the `None` path keeps the existing block_in_place / temp-runtime behaviour for worker-thread and test callers).
- **engine-geotiff reader:** `read_remote_chunk_f64` / `read_http_chunk_f64` / `read_http_range` take `Option<&Handle>`; `read_bbox_parallel` and `read_bbox_parallel_http` capture `Handle::try_current()` and pass it into the rayon closures; single-pixel callers pass `None` (already on a worker).

## Tests
New `get_range_on_from_foreign_thread` proves a fetch from a non-Tokio (foreign/rayon-like) thread returns correct bytes without panicking or creating a runtime. `cargo test -p ds-storage -p engine-geotiff` green; clippy clean.

Note: this is the latent remote-render fire flagged during the spike investigation; it was **not** the local-collection spike (that was #221, shipped). Part of epic #201.

### DIFF
--- a/crates/engine-geotiff/src/reader.rs
+++ b/crates/engine-geotiff/src/reader.rs
@@ -995,6 +995,9 @@ fn read_remote_chunk_f64(
     file_path: &Path,
     band_index: usize,
     ifd_index: u16,
+    // Runtime handle to drive the fetch on when called from a non-Tokio
+    // (rayon) thread; `None` for callers already on a runtime worker. See #222.
+    handle: Option<&tokio::runtime::Handle>,
 ) -> Result<Vec<Option<f64>>, DataServerError> {
     let idx = chunk_index as usize;
     if idx >= tile_info.tile_offsets.len() {
@@ -1037,13 +1040,19 @@ fn read_remote_chunk_f64(
         ))
     })?;
 
+    // Fetch a byte range, reusing the caller's runtime handle when on a rayon
+    // thread (avoids a per-call Runtime::new — see #222).
+    let fetch_range = |range: std::ops::Range<usize>| match handle {
+        Some(h) => store.get_range_on(obj_path, range, h),
+        None => store.get_range(obj_path, range),
+    };
+
     // Check cache for compressed bytes (keyed by file + chunk + IFD level)
     let compressed = if let Some(c) = cache {
         if let Some(cached) = c.get(file_path, chunk_index, ifd_index) {
             cached
         } else {
-            let fetched = store
-                .get_range(obj_path, offset..end)
+            let fetched = fetch_range(offset..end)
                 .map_err(|e| DataServerError::Engine(format!("Failed to read tile range: {e}")))?;
             // Validate response length matches request
             if fetched.len() != byte_count {
@@ -1058,8 +1067,7 @@ fn read_remote_chunk_f64(
             fetched
         }
     } else {
-        let fetched = store
-            .get_range(obj_path, offset..end)
+        let fetched = fetch_range(offset..end)
             .map_err(|e| DataServerError::Engine(format!("Failed to read tile range: {e}")))?;
         if fetched.len() != byte_count {
             return Err(DataServerError::Engine(format!(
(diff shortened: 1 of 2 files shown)


mrauhala:
Addressed both minor doc nits (commit 49b4566): reworded the `rt_handle` capture comment (it's not specific to spawn_blocking — `try_current()` also succeeds in an async task; we just pass the handle down) and narrowed `get_range_on`'s prohibition to *async tasks* (a running future), noting that a spawn_blocking thread or rayon worker is safe for `handle.block_on`. No code change; the fix logic was confirmed correct in the review.

PR #245, opened 2026-05-25
perf(engine-geotiff): bounded-upscale overview selection to kill the render cliff

### LINKED ISSUE #246: perf(engine-geotiff): overview-selection cliff — output just above largest overview decodes full resolution

## Problem

`TiffMetadata::select_overview` used a strict never-upscale rule (`>= output`), so an output only slightly wider than the largest overview fell through to the **full-resolution IFD**. On the production FMI composite (local, EPSG:3067, base 4963×7316 = 36 MP, largest overview 2481 px), desktop retina GetMaps (2650–3783 px) decoded the entire 36 MP base every frame; iPhone (~1 MP) used a small overview and was fast.

## Evidence

Live cache-busted loopback renders (local COG → pure server render time):

| width | render | level |
|---|---|---|
| 2481 | 348 ms | overview (9 MP) |
| 2550 | 653 ms | full-res 36 MP |
| 3783 | 1076 ms | full-res |

A 69-px increase across the 2481 boundary ~doubled render time.

## Fix

PR #245: two-pass `select_overview` — strict never-upscale for anything an overview can satisfy (no mid-zoom change), then a bounded (≤2×) clamp to the largest overview when the output exceeds it, else full resolution only at/near native resolution.

Scope: benefits the **direct** render path (non-web-merc CRS, or meta-tiling disabled). The meta-tile path already avoids the cliff via per-256px-tile overview selection.

## Labels
`performance`, `priority: high`, `effort: small` — milestone v0.2

mrauhala:
## Problem

The long-standing "desktop radar player is slow (long initial wait), iPhone is fast" complaint was **not** encode (PNG vs WebP), **not** rayon/meta-tiling. It's an **overview-selection cliff** in `engine-geotiff`.

`TiffMetadata::select_overview` required an overview to be `>=` the requested output resolution (strict never-upscale). The production FMI composite COG is **4963×7316 (36 MP)**, local, **EPSG:3067 / TM35FIN**, with overviews at **2481**, 1240, … px. Desktop retina GetMaps are **2650–3783 px wide** — just *above* the 2481 px largest overview — so `select_overview` returned `None`, and because 36 MP `< MAX_MAP_PIXELS (64 M)` the engine **decoded the entire 36 MP base every frame**. iPhone (~1 MP) picks a small overview and is fast. With `metatile_cache_mb = 0` (current prod), all WMS GetMaps go through this direct path — confirmed by the `slow WMS direct render` logs.

## Evidence (live, cache-busted loopback renders on prod — local COG, so this is pure server-side render time)

| WIDTH | render | level read |
|---|---|---|
| 2481 | 348 ms | overview (9 MP) |
| **2550** | **653 ms** | **full-res 36 MP** ← +69 px ≈ 2× |
| 2650 | 773 ms | full-res |
| 3783 | 1076 ms | full-res |
| 4963 | 1438 ms | full-res |

## Fix (two-pass, surgical)

- **Pass 1** — the original strict never-upscale selection, unchanged. Every output an overview *can* satisfy is selected exactly as before (no mid-zoom retune).
- **Pass 2** — only when the output exceeds the biggest overview: clamp to that overview if the upscale stays within `MIN_OVERVIEW_FRACTION` (0.5 → ≤ 2×); otherwise `None` → full resolution. This **bounds** the upscale and confines full-res decodes to ~native-resolution requests, rather than relocating the cliff.

The resample is **nearest-neighbour** (`lib.rs:1397`), so an upscale is blocky pixel replication; nodata is `None` before resample and colorize runs after, so there's **no cross-class averaging or nodata blending**. For the user's widths the upscale is 1.07–1.52×. Helps every geotiff collection (opera/dmi/smhi/fmi).

## Tests

Cliff fixed (2650 → ov0), largest retina (3783 → ov0), **mid-zoom unchanged** (1300 → ov0, guards against a global retune), **bounded-upscale cap** (sparse pyramid: >2× → full-res), native (4963 → full-res), and a **TM35FIN (EPSG:3067)** fixture exercising the production curved-CRS path. Full `engine-geotiff` suite + clippy green.

Closes #246

### DIFF
--- a/crates/engine-geotiff/src/lib.rs
+++ b/crates/engine-geotiff/src/lib.rs
@@ -1289,9 +1289,11 @@ impl ds_core::map_engine::MapEngine for GeoTiffEngine {
                 {
                     let full_pixels = ((c1 - c0) as usize) * ((r1 - r0) as usize);
                     if full_pixels > reader::max_map_pixels() {
-                        // Find the finest overview that fits under the limit
-                        // (overviews are sorted coarsest-last, iterate in reverse)
-                        for ov in metadata.overviews.iter().rev() {
+                        // Pick the finest overview that fits under the limit.
+                        // Overviews are sorted finest-first, so iterate in order
+                        // and take the first that fits — the highest quality
+                        // within the pixel budget.
+                        for ov in &metadata.overviews {
                             let ov_gt = metadata.overview_geo_transform(ov);
                             if let Some((oc0, or0, oc1, or1)) =
                                 ov_gt.bbox_to_pixels(west, south, east, north)
@@ -1302,7 +1304,8 @@ impl ds_core::map_engine::MapEngine for GeoTiffEngine {
                                 }
                             }
                         }
-                        // All overviews still exceed limit — use coarsest
+                        // Even the coarsest overview exceeds the limit — use it
+                        // anyway as the smallest available source.
                         metadata.overviews.last()
                     } else {
                         None
--- a/crates/engine-geotiff/src/lib.rs
+++ b/crates/engine-geotiff/src/lib.rs
@@ -1273,10 +1273,16 @@ impl ds_core::map_engine::MapEngine for GeoTiffEngine {
         let overview = metadata
             .select_overview(west, south, east, north, width, height)
             .or_else(|| {
-                // select_overview returned None — either no overviews exist,
-                // or all overviews have enough pixels. Check if full resolution
-                // exceeds the source pixel limit, and if so pick the finest
-                // overview that fits under the limit.
+                // select_overview returned None, which now means one of: no
+                // overviews exist; full resolution already fits the output; or
+                // the output exceeds the biggest overview by more than the
+                // bounded-upscale factor (i.e. it is at/near native resolution).
+                // In the last two cases we'd read full resolution — but if that
+                // would blow the MAX_MAP_PIXELS decode budget we must fall back
+                // to an overview regardless of the upscale bound: the hard pixel
+                // cap overrides the soft 2× quality preference (an over-cap
+                // upscale is still logged by the `Using overview …` debug below).
+                // Pick the finest overview that fits under the limit.
                 if let Some((c0, r0, c1, r1)) = metadata
                     .geo_transform
                     .bbox_to_pixels(west, south, east, north)
--- a/crates/engine-geotiff/src/reader.rs
+++ b/crates/engine-geotiff/src/reader.rs
@@ -542,9 +542,24 @@ impl TiffMetadata {
             return None;
         }
 
-        // Find the smallest overview where the bbox still covers enough
-        // source pixels to fill the output without upscaling.
-        // Overviews are sorted by decreasing resolution (largest first).
+        // Pick the coarsest overview that still provides at least
+        // `MIN_OVERVIEW_FRACTION` of the requested resolution — i.e. tolerate a
+        // bounded bilinear upscale (at most `1 / MIN_OVERVIEW_FRACTION`).
+        //
+        // A strict never-upscale rule (`>= output_*`) creates a latency cliff:
+        // any output only slightly wider than the largest overview falls all the
+        // way through to full resolution. On the production FMI composite (base
+        // 4963×7316, largest overview 2481 px) a 2650-px-wide retina GetMap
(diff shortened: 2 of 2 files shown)


mrauhala:
### Review responses (commit b991454)

Ran an internal devil's-advocate teardown alongside claude-review. Receipts:

**Fixed**
- **Nearest-neighbour, not bilinear** — verified the resample is `col_f.floor()` lookup (`lib.rs:1397-1407`). Comments corrected; an upscale is blocky pixel replication, not smooth interpolation. (Upside, also verified: nodata is `None` pre-resample and colorize runs after, so no cross-class averaging / 255-blending.)
- **No production-CRS test** — production is local **EPSG:3067 / TM35FIN** (curved). Added a TM35FIN fixture so the test exercises the 20-samples/edge `bbox_to_pixels` curvature envelope, not just axis-aligned WGS84.
- **Global mid-zoom retune** — reworked to two-pass: pass 1 is the original strict never-upscale (unchanged for every output an overview can satisfy), pass 2 only clamps at the cliff. Added a `mid_zoom_unchanged` test pinning 1300 px → ov0.
- **"Cliff just relocated to 2×"** — pass 2 now *bounds* the upscale (≤ 2×) and falls to full resolution only at/near **native** resolution, where full-res is the correct source. Added a `bounded_upscale_caps_at_2x` test (sparse pyramid, >2× → full-res).
- **Stale/duplicate doc blocks** above `select_overview` (claude-review) — replaced with one accurate summary.

**Considered but not changed**
- *Per-axis `AND` + `break` skips a valid level* — disagreed: both axes shrink monotonically across coarser overviews, so the qualifying set is a prefix and `break` is correct. The cited 4963×200 case correctly returns full-res (4963 cols genuinely needs it).
- *S3 byte-range cost dominates* — that was a different collection (`*-s3-cog`). The measured collection (`fmi-radar-composite-dbz`) is **local**, so the loopback numbers are pure render time.
- *Animation still convoys (fix shifts the curve, not flattens it)* — true; per-frame cost is the binding cost on the direct path and this ~halves it. Render concurrency is a separate axis, tracked separately.
- `MIN_OVERVIEW_FRACTION` could become per-collection config — left as a documented const for now; easy follow-up if a collection wants a different sharpness/latency trade.

Will post before/after loopback numbers once this is deployed.

mrauhala:
### Review round 2 (commit b29c1d3)

Both claude-review catches addressed:

- **Pass 2's 2× cap vs the `lib.rs` `or_else` fallback** — correct: when full resolution exceeds `MAX_MAP_PIXELS` (64 M), the fallback reads an overview regardless of the 2× bound. That's the intended precedence (you physically cannot decode >64 MP, so a steeper upscale is the only option) — the hard pixel cap wins over the soft quality preference. The bug was the *stale comment*, now rewritten to spell out the three `None` cases and the cap override (an over-cap upscale is still surfaced by the existing `Using overview …` debug log). Behaviour unchanged, and it never fires for the 36 MP production COG (36 M < 64 M).
- **Row-axis rejection untested** — added the suggested `3000×5000` assertion (width fits, height needs >2× → full-res), so the `(r1 - r0) >= min_rows` half of Pass 2 is now covered.

All `engine-geotiff` tests + clippy green.

mrauhala:
### Review round 3 (commit 25dbbd3)

- **`or_else` fallback returned coarsest, not finest (lib.rs:1285)** — real bug, fixed. It iterated `.iter().rev()` and returned the first overview under the 64 MP cap (the coarsest), contradicting both the comment and intent. Now iterates finest-first → returns the highest-quality overview within the pixel budget. Dormant for all current collections (the FMI COG is 36 MP < 64 MP, so the fallback never fires), but it was wrong.
- **`_ => break` in Pass 1 (reader.rs:563)** — not a bug, but clarified. Overviews are visited finest-first; both fall-through reasons are monotone in that direction (a too-small level is followed only by smaller ones; a sub-pixel `None` only by more-sub-pixel ones), so `break` cannot skip a qualifying level. This matches the pre-refactor behaviour; added a comment spelling out the monotonicity for each case.

A unit test for the `or_else` cap path would need a synthetic >64 MP COG fixture (it's inline in `get_raster_tile`); given it's dormant for every current collection I've left that as a possible follow-up rather than expanding this PR. Tests + clippy green.

Issue #240, opened 2026-05-25
perf(render): meta-tile assembly is 62% of slow-render time — per-pixel HashMap lookups; flat-index + parallelize


mrauhala:
Part of #201 (GeoTIFF radar WMS performance epic). Found via the #238 render-phase instrumentation.

## Finding (production, 1,790 slow renders)

The cold-render tail (~1 s p99) is **pure CPU in the meta-tile assembly**, not I/O and not queue wait:

| phase | avg | share of render |
|---|---|---|
| `sem_wait_ms` | 0 ms | — |
| `tile_loop_ms` | 344 ms | 35% |
| **`assemble_ms`** | **601 ms** | **62%** |
| `encode_ms` | 30 ms | 3% |
| total | 977 ms | (max 3,989 ms) |

Driver: the production client requests **retina-resolution fullscreen — avg 2762×1619 = 4.5 MP** (up to 5.6 MP), so each render covers **~110 tiles** (max 208) and assembles **4.5 M output pixels**. Per-tile engine cost is **4.95 ms/miss**, matching the dev bench (2.4 ms) — so the engine isn't I/O-bound, there are just 5–7× more tiles than a 1024² render. Affected layers: `fmi-radar-composite-dbz`, `met-radar-composite-dbz`, `chmi-radar-composite-dbzh`, `opera-reflectivity`.

Crucially, **assembly cost is independent of cache hits** — even a fully-warm render still assembles 4.5 M pixels — so warm renders are slow too. The metatile cache (2 GB, 55% hit) doesn't help this phase.

## Root cause

In `crates/render/src/metatile.rs`, `sample_bilinear` → `global_pixel` does a **`HashMap<(i64,i64), Option<Arc<[u8]>>>::get()`** lookup for *each of the 4 bilinear neighbours per output pixel*. That's **4.5 M px × 4 ≈ 18 M hashmap lookups per render** → the ~600 ms.

## Fix

1. **Flat-index the mosaic.** The covering tiles form a contiguous `col0..=col1 × row0..=row1` grid. Store them in a `Vec<Option<Arc<[u8]>>>` indexed by `(row-row0)*ncols + (col-col0)` (built directly in the tile loop, which already iterates in that order) and index it in `global_pixel` instead of hashing. Removes all 18 M hashmap lookups.
[...]

PR #249, opened 2026-05-25
perf(render): single-threaded meta-tile assembly — kill the rayon convoy

mrauhala:
## Problem

The meta-tile assembly parallelises its per-output-row loop with rayon `par_chunks_mut` (#240/#242). In isolation that's cheap (~25 ms for a retina frame). But the assembly runs **inside a per-request render task**, and the WMS handler serves **many renders concurrently** — a radar player prefetches **~10 timesteps per view change**. Those ~10 concurrent rayon assemblies all route through the **one global rayon pool** and **convoy**: they oversubscribe it and park on its latches, inflating each assembly to ~750 ms and serialising the burst into **~6–8 s** — with the **CPU idle** (parking, not compute), which is why it never looked CPU-bound.

## Evidence (live prod, via the dev radar app, metatile on, southern-Finland zoom)

Pan to a fresh area, then **back to a visited area**. Server logs for the revisit:

```
render_ms=766  misses=0  tile_loop=0  assemble_ms=755  encode_ms=5
render_ms=720  misses=0  tile_loop=0  assemble_ms=709  encode_ms=5
render_ms=714  misses=0  tile_loop=0  assemble_ms=702  encode_ms=5
render_ms=742  misses=0  tile_loop=0  assemble_ms=731  encode_ms=5
```

Every tile cached (`misses=0`, decode free), PNG encode 5 ms — and the render is **still ~750 ms, entirely in `assemble_ms`**. The metatile cache works perfectly and delivers nothing, because the assembly dominates and convoys. Client saw **5.8–8.2 s** per prefetch burst (durations climbing in even steps = serialising). Single-render assembly uncontended is ~25 ms; the 30× inflation is pure pool contention.

## Fix

Make the assembly **sequential** (`chunks_mut`, not `par_chunks_mut`). Each render's assembly now uses one core, and the ~10 concurrent prefetch renders **parallelise across cores via request concurrency** instead of fighting over one pool — the burst collapses from ~7 s toward sub-second. This is the *no per-request rayon on a concurrently-served path* rule; it's the intent of the reverted #243, redone fresh on current main (the rayon `par_chunks_mut` had been reintroduced when #245 merged).

`rayon` is now unused in `ds-render`, so the dependency is dropped.

## Notes
- Independent of the meta cache logic (which is correct — `misses=0` proves it warms).
- `#245` (overview cliff) only affects the **direct** path; the meta path requests 256 px tiles that never hit the cliff, so this is the meta path's separate, dominant cost.
- Orthogonal follow-ups: the client over-requests a ~7.9 MP image (~4× the viewport) which inflates *both* paths; and an all-nodata/transparent fast-path in the assembly would help the oversized-but-mostly-empty render.

Tests: full `ds-render` suite (incl. the `assembled_pixels_track_geography` fidelity test) + clippy green; `cargo check -p server` clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -782,7 +782,6 @@ dependencies = [
  "jpeg-encoder",
  "png",
  "quick_cache",
- "rayon",
  "webp",
 ]
 
--- a/crates/render/Cargo.toml
+++ b/crates/render/Cargo.toml
@@ -11,6 +11,3 @@ webp = { version = "0.3", default-features = false }
 quick_cache = "0.6"
 bytes = "1"
 chrono = { version = "0.4", features = ["serde"] }
-# CPU data-parallelism for the meta-tile assembly (per-output-row, pure compute,
-# no I/O). A compute dependency, not a framework.
-rayon = "1"
--- a/crates/render/src/metatile.rs
+++ b/crates/render/src/metatile.rs
@@ -36,7 +36,6 @@ use std::time::Instant;
 
 use chrono::{DateTime, Utc};
 use quick_cache::sync::Cache;
-use rayon::prelude::*;
 
 use ds_core::error::DataServerError;
 use ds_core::map_engine::RasterTile;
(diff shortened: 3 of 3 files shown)


mrauhala:
Closing — based on a stale local `main` (pre-#243, which still had the rayon `par_chunks_mut`). On current `origin/main` the rayon assembly was already removed by #243's revert and is what's deployed, so this PR is redundant and conflicts. The ~750 ms assemble measured in prod is therefore the **sequential** HashMap assembly cost (per-pixel `sample_bilinear` over a ~7.9 MP, ~86%-transparent oversized image), not a rayon-pool convoy. Real fix is to make the sequential assembly cheaper (skip empty/transparent regions) and/or shrink the oversized client request — will open a fresh PR off current main.

Issue #207, opened 2026-05-19
perf(render): IntegerLutColorMap implemented but never wired into WMS/Maps/Tiles handlers


mrauhala:
Part of #201 (GeoTIFF radar WMS performance epic).

## Problem

`IntegerLutColorMap` provides O(1) direct-index color lookup for integer data (U8: 256 entries, U16: up to 65,536). It is implemented and exported but **never used in production** — the WMS/Maps/Tiles handlers always build a float-path `LutColorMap`/`LinearColorMap`. `IntegerLutColorMap::from_colormap` is exercised only in tests.

## Evidence

- `IntegerLutColorMap`: `crates/render/src/colormap.rs:114-167` (direct array index, 0 FP ops/pixel)
- `LutColorMap::color`: `crates/render/src/colormap.rs:88-104` — 3 FP ops/pixel (`(v - min) / (max - min)` + `round()` + `clamp()`)
- `LutColorMap::to_integer_lut`: `crates/render/src/colormap.rs:60-62` — conversion helper already exists, uncalled from production code
- Handler call site: `crates/api-wms/src/handlers.rs:142` — `style_info.colormap.clone()` is always a float-path colormap

## Impact

~3 FP ops/pixel saved → ~10 ms per 2M-pixel fullscreen render. Modest on its own; a natural companion to the `Vec<Option<f64>>` integer-pipeline issue (#201).

## Suggested fix

In `StyleInfo` construction (where colormaps are built from config), call `LutColorMap::to_integer_lut(min, max)` when the data type is integer-bounded and the range fits ≤65,536 entries; use the result as the active `Box<dyn ColorMap>`. No API change.

## Labels
`priority: medium`, `performance`, `effort: small` — milestone v0.2

_Audit: hunter-render, survived devil's-advocate review._

mrauhala:
Closed by PR #250 (merged as `ff0d459` on 2026-05-27). The wiring landed in `build_colormap_from_wms_config` via the new `maybe_wrap_integer_lut` helper; any colormap whose value range fits in 17–65 536 integer entries is now pre-baked into an `IntegerLutColorMap` at config-load time, so the WMS/Maps/Tiles colorize hot loop is one offset-shift + round + clamp + LUT index per pixel.

Round-5 review hardening that landed alongside:
- All three `ColorMap` impls unified on round-nearest + saturate-clamp + nodata-on-NaN (previously diverged).
[...]

PR #250, opened 2026-05-27
perf(server): wire IntegerLutColorMap into the WMS/Maps/Tiles colorize path

mrauhala:
## Problem

`IntegerLutColorMap` (in `ds-render`) provides **O(1) direct-index** colour lookup — replacing the ~3 FP ops/pixel of the float-path `LutColorMap`/`LinearColorMap` with a single `(v - offset).round() as i64` index. It was implemented, exported, and used in tests, but **never used in production**: the WMS/Maps/Tiles handlers always built a float-path colormap. Closes **#207**.

## Fix

Wired in `build_colormap_from_wms_config` (`server/src/admin.rs`), the one place WMS/Maps/Tiles `StyleInfo` colormaps are built from config. A small `maybe_wrap_integer_lut(cmap, min, max)` helper replaces the float-path colormap with an integer LUT when:

- bounds are finite and `min ≤ max`,
- the integer span is **≥ 16 steps** — narrower would be coarser than 1 LUT entry per unit, visibly degrading a fine-resolution gradient (e.g. viridis on 0..1),
- the span fits the **65 536-entry cap** (`IntegerLutColorMap::from_colormap` returns `None` otherwise).

| Colormap example | Range | Wrapped? |
|---|---|---|
| `radar_dbz` | -32..95 (128 entries) | ✅ |
| `temperature`, `wind_speed`, `precipitation_rate` | 0..40+ | ✅ |
| `viridis` default 0..1 | 2 entries | ❌ (preserves gradient) |
| user-configured huge range (>65 535 span) | — | ❌ |

## Behaviour notes

- The wrap is **transparent** to callers — `Arc<dyn ColorMap>` in, `Arc<dyn ColorMap>` out.
- `IntegerLutColorMap::color()` now rounds-nearest (`(v - offset).round() as i64`) and **saturate-clamps** out-of-range to the first/last LUT entry — matching `LutColorMap`'s `.round()` + `.clamp()` behaviour exactly at integer entries.
- Sub-unit-precision data (e.g. fmi radar's `scale = 0.5` → 0.5-step physical dBZ) snaps to the nearest integer dBZ in the LUT. Invisible at the class granularity radar colormaps are designed around.
- **NaN / ±∞ now uniformly colour as `nodata_color`** across all three colormap impls. Prior to this PR `LutColorMap` returned `lut[0]` (an artefact of `NaN as usize = 0`) and `LinearColorMap` returned the *last* stop — inconsistent and accidental. Deliberate fix.

## Expected impact

~10 ms saved per 2 MP fullscreen render at integer-LUT depths (per #207's estimate). At the prod 7.9 MP retina request size that's ~40 ms/render — compounds on the 10-timestep prefetch (~400 ms/view change) and helps every direct-path GeoTIFF render of integer-typed data.

## Tests

- `admin.rs`: wrap-when-wide-enough (radar_dbz integer-match + saturate-clamp at boundaries), skip-narrow (viridis), skip-over-cap, skip-non-finite.
- `colormap.rs`: out-of-range saturates (rename of the old `_transparent` test), single-value-range saturates, rounds-nearest-not-toward-zero (separate, on a distinct-per-integer LinearColorMap), all-colormaps-agree-on-NaN.

Full `cargo test -p ds-render -p server` green; clippy clean.

### DIFF
--- a/crates/render/src/colormap.rs
+++ b/crates/render/src/colormap.rs
@@ -89,6 +89,14 @@ impl ColorMap for LutColorMap {
     fn color(&self, value: Option<f64>) -> [u8; 4] {
         match value {
             None => self.nodata_color,
+            // NaN/±∞ are not real data — colour them as nodata, matching
+            // `IntegerLutColorMap` and `LinearColorMap`. Without this guard
+            // the normalised index goes through `NaN as usize = 0` (Rust 1.45+
+            // saturating cast) and returns `lut[0]` (the first stop's colour),
+            // which the integer-LUT wrap (#207) would silently switch to
+            // `nodata_color` — a real divergence for any pipeline that lets
+            // NaN reach the colorizer.
+            Some(v) if !v.is_finite() => self.nodata_color,
             Some(v) => {
                 if self.max <= self.min {
                     return self.lut[0];
@@ -202,6 +210,10 @@ impl ColorMap for LinearColorMap {
     fn color(&self, value: Option<f64>) -> [u8; 4] {
         match value {
             None => self.nodata_color,
+            // Match the NaN guard in LutColorMap / IntegerLutColorMap — otherwise
+            // a NaN falls through every comparison in `interpolate_stops` and
+            // returns the LAST stop's colour, diverging from the other paths.
+            Some(v) if !v.is_finite() => self.nodata_color,
             Some(v) => interpolate_stops(&self.stops, v),
         }
     }
@@ -793,6 +805,35 @@ mod tests {
         // Out of range above → last entry.
         assert_eq!(lut.color(Some(11.0)), linear.color(Some(10.0)));
         assert_eq!(lut.color(Some(1e9)), linear.color(Some(10.0)));
+    }
+
+    #[test]
+    fn test_all_colormaps_treat_nan_and_infinity_as_nodata() {
+        // All three colormap types must agree on non-finite inputs — otherwise
+        // `maybe_wrap_integer_lut` (the #207 wrap) silently changes behaviour
+        // when an engine lets NaN reach the colorizer.
+        let stops = vec![
+            ColorStop {
+                value: 0.0,
+                color: [255, 0, 0, 255],
+            },
+            ColorStop {
+                value: 10.0,
+                color: [0, 255, 0, 255],
+            },
+        ];
+        let lut = LutColorMap::from_stops(&stops, 0.0, 10.0);
+        let lin = LinearColorMap::new(stops);
+        let int_lut = IntegerLutColorMap::from_colormap(&lin, 0, 10).unwrap();
+        for v in [f64::NAN, f64::INFINITY, f64::NEG_INFINITY] {
+            assert_eq!(lut.color(Some(v)), [0, 0, 0, 0], "LutColorMap on {v}");
+            assert_eq!(lin.color(Some(v)), [0, 0, 0, 0], "LinearColorMap on {v}");
+            assert_eq!(
+                int_lut.color(Some(v)),
+                [0, 0, 0, 0],
+                "IntegerLutColorMap on {v}"
+            );
+        }
     }
 
     #[test]
--- a/crates/render/src/colormap.rs
+++ b/crates/render/src/colormap.rs
@@ -2,6 +2,13 @@
 pub trait ColorMap: Send + Sync {
     /// Map a value to an RGBA color. None = nodata → transparent.
     fn color(&self, value: Option<f64>) -> [u8; 4];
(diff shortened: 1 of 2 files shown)


mrauhala:
Both fair:
- **Truncation direction** — corrected. `v as i64` rounds toward zero, not floor; for ranges crossing zero negative half-steps round *up* (toward zero) and positive ones round *down*. The LUT just picks one of the two neighbouring integer entries either way, so the user-visible behaviour (which is what the test pins, with `Arc::ptr_eq` + colour-equality at integers) is unaffected — only the docstring claim was wrong.
- **Length** — collapsed to a four-line block keeping the non-obvious bits (#207 reference, the <16-span rationale, the rounds-toward-zero caveat). Per CLAUDE.md.

mrauhala:
Round 2 catches all addressed — and these were the right ones to push on. The bug was deeper than docstrings: `IntegerLutColorMap` itself diverged from `LutColorMap` on rounding (truncate-toward-zero vs `.round()`) and on out-of-range handling (transparent vs clamp). Either would have shipped a visible regression for any non-integer data or any pixel outside the configured range (GRIB extremes silently disappearing).

Fix moves into `IntegerLutColorMap::color()`:
- `(v - offset).round() as i64`, clamped to `[0, len-1]` → matches `LutColorMap`'s round-nearest + saturate-clamp at integer entries.
- At non-integer values, both paths round-nearest in index space; they diverge by sub-unit only because of LUT-density (LutColorMap fixed 4096-entry vs IntegerLutColorMap 1-per-integer-unit). That's by design — the wrap trades sub-integer precision for the per-pixel speedup.

Tests:
- `test_integer_lut_out_of_range_saturates` (renamed) now pins clamp-to-first/last.
- `test_integer_lut_rounds_nearest_not_toward_zero` pins the rounding direction on a LinearColorMap with distinct colours per integer (radar_dbz is transparent at the low end so can't show it).
- `admin.rs` wrap test now also pins saturation at the boundary.

Also fixed the 65 535/65 536 off-by-one in the docstring and dropped the truncation caveat (no longer applies).

mrauhala:
Both round-3 catches fair and important:
- **NaN divergence** — `IntegerLutColorMap` returned `nodata_color` for `NaN`, `LutColorMap` returned `lut[0]` via the saturating cast, `LinearColorMap` returned the last stop's colour. The #207 wrap silently changed one to the other. Added a `!v.is_finite() → nodata_color` guard to both `LutColorMap` and `LinearColorMap` so all three paths agree; added a test pinning the three colormaps' outputs equal on `NaN`/`±∞`.
- **Off-by-one in the docstring** — `span = hi - lo` is the *difference*, so the guard's `span ≥ 16` means ≥17 entries. Reworded to "spans below 16 integer steps" (the actual semantic threshold).

Tests: ds-render 50 / server 49 + 10 green, clippy clean.

mrauhala:
Round 4 catches addressed (commit-only commentary + PR body):
- NaN guard now explicitly called out as a deliberate behaviour change for LutColorMap/LinearColorMap (their prior NaN outputs — `lut[0]` and last-stop respectively — were accidental and already inconsistent with each other; unifying on `nodata_color` is the right fix).
- The `None` arm in `maybe_wrap_integer_lut` marked as defensively unreachable given the current gate.
- PR body updated to reflect the saturate-clamp out-of-range behaviour (was stale from before the round-2 fix).

No behaviour change in this commit.

mrauhala:
Round 5 catch addressed:

**`from_colormap` was hardcoding `nodata_color: [0, 0, 0, 0]`** instead of copying it from the wrapped source. The reviewer's suggested fix landed in `a5ccff7`:

- Added `fn nodata_color() -> [u8; 4]` to the `ColorMap` trait with a `[0, 0, 0, 0]` default.
- Overridden in `LutColorMap` / `IntegerLutColorMap` / `LinearColorMap` to return their stored field.
- `IntegerLutColorMap::from_colormap` now calls `source.nodata_color()` and propagates it through both the normal and the empty-range paths.
- New `test_integer_lut_preserves_source_nodata_color` uses a mock with opaque-grey nodata to prove the propagation for None / NaN / out-of-range / empty-range.

The two test-only impls in `metatile.rs` don't need updating — the trait default covers them. Today every concrete colormap still uses transparent nodata, so this is a tautology in practice, but the trait now makes the propagation visible/correct for any future opaque-nodata colormap.

The other four round-5 findings (out-of-range doc, NaN behaviour-change call-out, `LinearColorMap` first-vs-last asymmetry, `None` arm unreachability) were addressed as commentary in `5064310` and the PR body. No behaviour change in those.

Issue #205, opened 2026-05-19
GeoTIFF: remeasure cold-file rotation stalls and identify their current cause


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open pending current evidence.

**Current assessment and challenge:** Cold I/O attribution is an old hypothesis; mmap/decoded caches and runtime isolation changed the path. Re-measure before claiming disk reads explain multi-second stalls.

**Next step / remaining scope:** Capture current cold-rotation timings with request, admission, decode and I/O attribution before retaining the old disk-read root cause.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

Part of #201 (GeoTIFF radar WMS performance epic). Overlaps the "adjacent-time pre-warm" scope of #147.

## Problem

Radar files rotate every ~60 s. When a new file appears, the first requests to hit it find a cold OS page cache → real disk reads → 50–200 ms added per tile. This is the primary explanation for the **172 WMS requests >5 s** (~0.28% of traffic, ≈1 event/minute — matching the file rotation cadence).

## Evidence

- Fresh decoder per request on the local path: `crates/engine-geotiff/src/reader.rs:1595`
- Prod metrics: 172 requests in the `le="+Inf"` bucket beyond `le="5"`; cadence matches per-minute file rotation.
- The existing `TileCache` stores **compressed** bytes — for local files the OS page cache already provides block-level caching, so adding `TileCache` to the local path does **not** fix this. The gap is the absence of a warm-up step.

## Impact

The dominant source of the multi-second tail. Each cold render also holds a render-semaphore permit for its full (slow) duration, reducing capacity for the burst that follows (see #201 → semaphore issue).

## Suggested fix

In `poll_once()`, after discovering a new file on the background poller thread, pre-warm it **before** any client request arrives: open it, traverse the IFD, and read (or `madvise(WILLNEED)` / sequentially touch) its tile data so the OS page cache is warm. Coordinate with #147, which already lists "adjacent-time pre-warm" in scope.

## Labels
`priority: medium`, `performance`, `reliability`, `effort: small` — milestone v0.2

_Audit: hunter-render, survived devil's-advocate review._

</details>

PR #253, opened 2026-05-27
perf(engine-geotiff): mmap local GeoTIFFs once instead of File::open per request (#204)

### LINKED ISSUE #204: perf(engine-geotiff): per-request File::open + 8KB BufReader + IFD re-parse for local GeoTIFF

## Problem

Every render of a local GeoTIFF file re-opens the file from scratch: `File::open` → `BufReader::new(file)` (8 KB default buffer) → `Decoder::new(...)` (traverses the TIFF IFD chain). The 8 KB `BufReader` undersizes the buffer for Deflate-compressed 256×256 tiles (typically 5–50 KB), forcing multiple `read()` syscalls per tile even when the file is already in the OS page cache.

## Evidence

- `DataSource::open_decoder`: `crates/engine-geotiff/src/reader.rs:132-156` — fresh `File` + `BufReader` + `Decoder` per call
- Called per render on the local path: `crates/engine-geotiff/src/reader.rs:1595` (`read_bbox_inner`)
- Prod confirmation: `storage_bytes_read_total` has **no entry** for the local `fmi-radar-composite-dbz` — the local path bypasses the storage abstraction entirely.

## Impact

With a warm OS page cache the open + IFD traversal + reads come from DRAM, so overhead is ~50–200 µs per request — real but secondary. It compounds with the cold-file case (#201 → cold-read issue).

## Suggested fix

Memory-map local files once at catalog load (`memmap2`), store the `Arc<Mmap>` on the `CatalogEntry`, and reuse it across all requests. Tile reads become direct memory-range accesses — no per-request `File::open`, no `BufReader`, no IFD re-parse (tile offsets are already in `TiffMetadata`).

## Labels
`priority: medium`, `performance`, `effort: small` — milestone v0.2

_Audit: hunter-render, survived devil's-advocate review._

mrauhala:
Implements #204: replace per-request `File::open` + `BufReader` + `Decoder::new` for local GeoTIFFs with a memory-mapped buffer that's built once at catalog scan and reused on every subsequent render.

## Problem

Every WMS / Maps / Tiles render against a local `.tif` re-opened the file from scratch: `File::open` → `BufReader::new(file)` (8 KB default — too small for typical 5–50 KB Deflate-compressed tiles) → `Decoder::new(...)` (full IFD chain re-parse from disk). With a warm OS page cache the overhead was ~50–200 µs per render; cold rotation cases were materially worse. This compounds across the radar-player's prefetch burst where 10+ concurrent renders all paid the cost.

## Approach

Memory-map each file lazily on first decoder open. Cache the `Arc<Mmap>` in the `DataSource::LocalFile` variant via an `OnceLock` shared across clones. The catalog scan triggers the mmap once; every subsequent request reuses it with **zero filesystem syscalls** — `Decoder::new` runs over a `Cursor<&[u8]>` against the kernel page cache directly.

Per the issue's "small effort" scope, the IFD parse itself remains (now over memory, not disk). Eliminating the IFD parse entirely by caching tile offsets in `OverviewLevel.tile_info` for local files — same shape we already populate for remote — would shave another ~30–50 µs. Tracked as a possible follow-up if needed after deploy measurement.

## What changed

- `DataSource::LocalFile(PathBuf)` → `LocalFile { path, mmap_cache: Arc<OnceLock<...>> }`. `from_path` stays infallible (lazy mmap means existing tests passing fake `/tmp/test.tif` paths to the FileEntry constructor keep working).
- New private `SharedBytes` enum: `Heap(Bytes)` (downloaded payloads) or `Mmap(Arc<Mmap>)` (local files). Both impl `AsRef<[u8]>`, so `Cursor<SharedBytes>: Read + Seek`.
- `DecoderWrapper` collapses from two variants (`File`/`Memory`) to a single newtype over `Decoder<Cursor<SharedBytes>>`. Each method drops its 2-arm match. Net code reduction even with the new types defined.
- `from_header_read` and `from_http_header_read` updated to use the new buffer type (downloaded header bytes go through `SharedBytes::Heap`).

## Safety

`unsafe { Mmap::map(&file) }` requires the file to not change underneath us. MeteoCore's publishing convention is atomic rename — replacement files appear under a new inode and the catalog scan picks them up as fresh entries; existing mmaps keep their old pages until the entry is dropped. This is the same hazard the previous `File::open` + `BufReader` path implicitly assumed; documented in the `load_mmap` doc comment.

A cached error (mmap failed) means we won't retry — the catalog scan would have hit the same error and either skipped the entry or failed loudly, so the request-time error is informational rather than recoverable.

## Tests

| Suite | Result |
|---|---|
| `engine-geotiff` lib (incl. new `local_file_mmaps_once_and_reuses_across_calls`) | **71 passed** |
| `engine-geotiff` integration (`render_projection.rs` exercises the full local-GeoTIFF render path) | **5 passed** |
| `engine-geotiff` STAC integration | **8 passed** (+ 2 ignored, unchanged) |
| `ds-core` | **91 passed** |
| `ds-render`, `api-wms`, `api-maps`, `api-tiles`, `server` | all pass |
| `cargo clippy -p engine-geotiff --all-targets -D warnings` | clean |
| `cargo fmt --check` | clean |

The new unit test verifies that five back-to-back `open_decoder` calls return the same `Arc<Mmap>` pointer (no re-mmap), and that cloning the `DataSource` shares the cache.

## Expected impact

- ~80–170 µs shaved off every local-GeoTIFF render (warm-cache case).
- Cold-rotation (#205) case slightly improved too — first request pays the mmap cost, subsequent requests don't.
- No effect on remote (S3 / HTTP / STAC) paths.

Closes #204. Part of #201 (GeoTIFF render epic).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -849,6 +849,7 @@ dependencies = [
  "ds-storage",
  "flate2",
  "futures",
+ "memmap2",
  "quick_cache",
  "rayon",
  "regex",
--- a/crates/engine-geotiff/Cargo.toml
+++ b/crates/engine-geotiff/Cargo.toml
@@ -15,6 +15,7 @@ regex = "1"
 tokio = { version = "1", features = ["rt", "sync", "time", "fs"] }
 tracing = "0.1"
 bytes = "1"
+memmap2 = "0.9"
 ds-storage = { path = "../storage" }
 futures = "0.3"
 flate2 = "1"
--- a/crates/engine-geotiff/src/catalog.rs
+++ b/crates/engine-geotiff/src/catalog.rs
@@ -365,22 +365,24 @@ pub fn scan_directory(
             // else: initial scan (existing empty) — accept immediately
         }
 
-        // Reuse cached metadata if file size unchanged
-        let metadata = if let Some(entry) = existing_entry {
-            if entry.file_size == file_size {
-                match entry.metadata().map(|m| (**m).clone()) {
-                    Some(m) => m,
-                    None => match TiffMetadata::from_file(&path) {
-                        Ok(m) => m,
-                        Err(e) => {
-                            tracing::warn!("Skipping {}: {e}", path.display());
-                            continue;
-                        }
-                    },
-                }
+        // Reuse the prior entry's `DataSource` (and its warm mmap cache from
+        // #204) when the file size hasn't changed — otherwise the OnceLock is
+        // replaced every poll cycle and the mmap is re-issued on the next
+        // render. For genuinely new or size-changed files build the source
+        // once and parse metadata through it, so we don't mmap the file twice
+        // (once via `from_file` then again for the stored entry).
+        let cached_unchanged = existing_entry
+            .filter(|e| e.file_size == file_size)
+            .and_then(|e| e.source().map(|s| (e, s)));
+
+        let (source, metadata) = if let Some((entry, old_source)) = cached_unchanged {
+            if let Some(meta_arc) = entry.metadata() {
+                ((**old_source).clone(), (**meta_arc).clone())
             } else {
-                match TiffMetadata::from_file(&path) {
-                    Ok(m) => m,
+                // Stub entry with no metadata yet — parse via the existing
+                // source so we reuse its mmap cache instead of opening fresh.
+                match TiffMetadata::from_source(old_source) {
+                    Ok(m) => ((**old_source).clone(), m),
                     Err(e) => {
                         tracing::warn!("Skipping {}: {e}", path.display());
                         continue;
@@ -388,8 +390,9 @@ pub fn scan_directory(
                 }
             }
         } else {
-            match TiffMetadata::from_file(&path) {
-                Ok(m) => m,
+            let source = DataSource::from_path(&path);
+            match TiffMetadata::from_source(&source) {
+                Ok(m) => (source, m),
                 Err(e) => {
                     tracing::warn!("Skipping {}: {e}", path.display());
                     continue;
@@ -411,7 +414,6 @@ pub fn scan_directory(
             );
         }
 
-        let source = DataSource::from_path(&path);
         entries.insert(
             datetime,
             FileEntry::loaded(path, source, metadata, file_size),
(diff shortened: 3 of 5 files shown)


mrauhala:
All three findings addressed in `87c04e7`:

**Finding 1 (HIGH) — mmap discarded every poll cycle.** Fixed in `scan_directory` by cloning the prior entry's `DataSource` (a cheap Arc-refcount bump on the shared `Arc<OnceLock<...>>`) when file size is unchanged. Mirrors the existing remote-path `(*entry).clone()` reuse. The warm mmap now genuinely survives across polls, not just within one.

**Finding 2 (MEDIUM) — double-mmap per changed file.** Fixed by consolidating: catalog now builds the `DataSource` once and parses metadata through it (`TiffMetadata::from_source(&source)`), then stores the same `source` in the entry. `TiffMetadata::from_file` had no production callers after this and is removed; 3 lib tests updated to `from_source(&from_path(...))`.

**Finding 3 (LOW) — transient failure caching window.** Updated the `load_mmap` doc comment to spell out that the recovery window equals `poll_interval_secs` (next size-changed scan rebuilds the source) and that initial-scan failures are skipped by the scan loop rather than poisoning the catalog. No code change.

**New test** `scan_directory_reuses_data_source_across_polls` runs two consecutive `scan_directory` calls against the real radar fixture, forces the mmap cache to populate between them, and asserts both:
1. The `Arc<Mmap>` allocation is the same across scans (no re-issue of `Mmap::map`).
2. The `Arc<OnceLock<...>>` wrapper is the same allocation (proving the `DataSource` was cloned, not rebuilt from scratch).

Re-running benches now to capture the steady-state numbers post-fix.

mrauhala:
Round 2 catches both addressed in `faeb972`:

**Finding 1 (dead else arm).** Collapsed the unreachable `if let Some(meta_arc) = entry.metadata() { ... } else { ... }` to a single `.expect("Loaded entry always carries metadata alongside source")`. The reviewer's reasoning is exactly right — `cached_unchanged` is gated on `entry.source().is_some()` which is gated on `FileState::Loaded` which always carries metadata too. The "Stub entry with no metadata" path was structurally unreachable.

**Finding 2 (same-size atomic replacement = stale mmap).** Real regression vs the old `File::open`-per-render path. Fixed by storing `mtime: Option<SystemTime>` on `FileEntry` (captured from the dir-read stat — same syscall as `file_size`, zero cost) and gating `cached_unchanged` on `file_size && mtime` equality. If either side's mtime is `None` (FAT / exotic NFS), we conservatively rebuild — costs one extra mmap, never serves stale data. Remote/STAC entries (no filesystem mtime) pass `None` and aren't affected by the new check.

**New test** `scan_directory_rebuilds_source_when_mtime_changes`:
- Copies a real fixture into a per-PID tempdir, scans once, captures the `Arc<OnceLock<...>>`.
- Bumps mtime by 2 s via `File::set_modified` while asserting size stays equal.
- Re-scans and asserts the new entry's OnceLock is a **different** allocation (DataSource was rebuilt, not cloned) and that the new mtime is recorded.

Together with the existing `scan_directory_reuses_data_source_across_polls` (which now also benefits from the mtime check passing on unchanged files), the cache-invalidation contract is now fully covered both directions.

All 72 lib + 13 integration tests pass. fmt + clippy clean.

mrauhala:
Round 3 catches all addressed in `d2eea7f`. Solid catches — particularly Finding 2.

**Finding 1 (HIGH — SIGBUS on in-place overwrite).** Real crash hazard the doc comment under-sold. Full mitigation (signal handler / per-request size validation) is too invasive for this PR — minimum-viable fix landed: a startup `WARN` log in `GeoTiffEngine::new` at the local-data_path branch that names the atomic-rename requirement, so an operator who configures `rsync --inplace` sees the warning loudly. The doc comment was also wrong to call this a "convention" without enforcement; the warn makes it operational.

**Finding 2 (MEDIUM — same-second same-size atomic replacement).** You were exactly right about `rename(2)` not bumping mtime; on HFS+/APFS this is a real hole. Fixed by capturing `inode: Option<u64>` from `MetadataExt::ino()` alongside size+mtime — same dir-read stat, zero extra cost — and gating `cached_unchanged` on all three. Atomic rename always allocates a new inode, so this closes the gap fully on Unix. Non-Unix falls back to size+mtime as before.

**Finding 3 (LOW — misleading test panic).** Replaced `.expect("mmap_cache must already be populated after reuse")` with an explicit `panic!` block that names *both* possible causes (regression OR external fixture touch), so CI doesn't point operators at the wrong root cause.

**New test** `scan_directory_rebuilds_source_when_inode_changes_same_size_same_mtime` reproduces the exact scenario you described: atomic-rename a same-size copy of the fixture over itself, pin mtime to the original's, assert inode changed (gracefully skip if filesystem preserves it), then assert the second scan rebuilt the DataSource. The two prior tests (`scan_directory_reuses_data_source_across_polls`, `scan_directory_rebuilds_source_when_mtime_changes`) still pass — the inode check is naturally satisfied on unchanged files.

74 lib + 13 integration tests, fmt + clippy clean.

Issue #145, opened 2026-05-12
raster ETag should be content-derived (mirror MVT fix from #136)


mrauhala:
## Background

#136 switched the **vector tile (MVT)** ETag from key-derived (`VectorTileKey::etag()`) to content-derived (`CachedTile::new(bytes)` → FNV-1a over the encoded bytes). The reason: a key-derived ETag stays identical when the data behind the same `(z, x, y, style, time)` request changes, so browsers with a cached ETag keep getting `304 Not Modified` and render stale tiles indefinitely.

## What's still broken

`crates/api-tiles/src/handlers.rs::render_raster_tile` uses `CacheKey::etag()` from `ds-render`, which hashes the **key** fields (bbox, time, style, data version) rather than the rendered PNG bytes. The same failure mode applies:

- A server-side fix that produces different pixels for the same `(z, x, y, style, time)` is invisible to a client with a cached ETag.
- The browser sends `If-None-Match`, the server's key-derived ETag matches → `304`, the client renders the stale image until its cache entry ages out.

This is identical to the bug #136 fixed for vector tiles. The bot flagged it in the round-2 review of #136 as an explicit follow-up.

## Fix

Mirror the #136 pattern in `ds-render`:

1. Either expose a `CachedTile`-style struct that carries (bytes, etag) computed from content via FNV-1a, or compute the ETag inline in `render_raster_tile` after the cache hit/encode step.
2. Move the `If-None-Match` check **after** the cache lookup so the comparison runs against the content-derived ETag, not the key-derived one.
3. Add a regression test pinning "different bytes ⇒ different ETag" mirroring `ds_mvt::cache::tests::etag_rotates_after_data_refresh`.

## Acceptance

- `render_raster_tile` issues an ETag derived from the rendered PNG bytes.
- `If-None-Match` validation runs after the cache lookup.
- A test demonstrates that two raster tiles with identical keys but distinct rendered bytes carry distinct ETags.
- Existing raster tile caching tests still pass.

Tracked because the same browser-cache-survives-server-fix failure mode applies to every API that uses `CacheKey::etag()` (currently raster tiles via Maps, Tiles, and WMS).

mrauhala:
Filed from #136 round-2 review feedback; same content-vs-key ETag bug already fixed for MVT in that PR.

PR #254, opened 2026-05-27
perf(render): auto-emit 8-bit indexed-palette PNG when ≤256 colours

### LINKED ISSUE #252: perf(render): emit 8-bit indexed-palette PNG (PNG8) for colormap layers — 3-4× smaller bytes for the same image
## Summary

MeteoCore currently emits 32-bit RGBA PNG for all WMS/Maps/Tiles render outputs. For colormap-rendered layers (radar, classification, single-parameter rasters) the output palette is ≤256 distinct colours, so a paletted 8-bit indexed PNG would produce a **visually identical** image at 3–4× smaller bytes. FMI's GeoServer does this by default for the same kind of layer.

## Why bother

Bytes matter for end-user latency, especially on mobile or geographically-distant clients where TLS+TCP roundtrips dominate and the body transfer time scales with bytes:

| Service | Format | Bytes for matched 1024×1280 Finland radar | Server render | TTFB | Total (sequential median) |
|---|---|---|---|---|---|
| FMI GeoServer (`Radar:suomi_dbz_eureffin`) | 8-bit palette PNG | **96 KB** | slower | ~450 ms | 494 ms |
| MeteoCore (`fmi-radar-composite-dbz`) | 32-bit RGBA PNG | 324 KB | faster | ~180 ms | 255 ms |

MeteoCore is already ~2× faster end-to-end despite shipping 3.4× more bytes. Closing the byte gap with PNG8 should compound the win for slower clients (currently dominated by transfer time on mobile/edge networks).

Even at a flat 3× compression ratio, a 60-tile radar player session pulling 18 frames currently transfers ~6 MB; with PNG8 that drops to ~2 MB — meaningful on metered connections.

## Scope

- A new `FORMAT=image/png; mode=8bit` (or `image/png8`) variant on `/wms`, `/maps`, `/tiles` endpoints — same conventions as GeoServer.
- The default `FORMAT=image/png` keeps emitting RGBA (no behaviour change for existing clients).
- When the rendered image uses a `ColorMap` whose palette fits in ≤256 entries (radar, classification — basically any `LutColorMap`/`IntegerLutColorMap` style), encode as paletted PNG with a transparency entry (`tRNS` chunk).
- When the palette would exceed 256 colours (continuous gradients, multi-band false-colour), either reject with `FORMAT not supported` or fall back to RGBA — to be decided in design.
- Update `GetCapabilities` to advertise the new format on layers that can produce it.

## Implementation notes (suggested, not prescriptive)

- `crates/render/src/lib.rs`'s `ImageFormat` enum already discriminates Png/Jpeg/Webp; add `Png8` (or `PngPalette`) as a new variant.
- The `png` crate supports indexed-colour writing via `Encoder::set_color(ColorType::Indexed)` and `Encoder::write_header()` with `set_palette`/`set_trns`. Should be a self-contained change to the PNG encoder path in `ds-render`.
- For colormap-rendered layers, the palette is the `ColorMap`'s underlying LUT — already computable at config time, so the per-render cost is just the encoder path (no palette derivation per frame). With `IntegerLutColorMap` already wired (#207, PR #250), the LUT is literally `Vec<[u8; 4]>` ready to use.
- The cache key already includes `format: u8` — bumping to a new variant is a one-line change.
- ETag stability: an 8-bit PNG of the same scene must produce a deterministic byte stream so the cache + `If-None-Match` continue to work. The `png` crate is deterministic; verify there's no per-call randomisation (timestamps, etc.) by hashing two encodes of the same buffer in a test.

## Open design questions (resolve in PR)

1. **MIME spelling:** `image/png; mode=8bit` (GeoServer-compatible) or `image/png8` (shorter, also accepted by GeoServer) or both?
2. **Fallback behaviour** when the palette exceeds 256 colours: error vs. silent RGBA fallback?
3. **Tile cache interaction**: the tile-cache stores encoded bytes; with a new format variant the cache hit rate halves if clients alternate. Probably fine — same key→same format, and most clients ask for one format.

## Suggested test

Add to `crates/render/tests/` (or extend existing PNG round-trip tests):
- Encode a 256×256 frame in PNG8 with a `radar_dbz` palette.
- Decode it back and verify pixel-equivalence with the RGBA encoding of the same frame.
- Assert PNG8 file is ≥3× smaller than RGBA for the same frame.
- Assert two encodes of the same buffer produce byte-identical output.

Labels: `enhancement`, `performance`, `priority: medium`
Milestone: `v0.3`

mrauhala:
Closes #252.

> **Heads-up to the requester:** you referenced issue #242, but #242 is a reverted PR (rayon meta-tile assembly). The PNG8 issue matching your description is #252 — this PR targets that.

## Design (revised after first round of review)

The first iteration of this PR exposed PNG8 as a separate `FORMAT=image/png; mode=8bit` (GeoServer-style) opt-in, with a parallel `ImageFormat::Png8` variant, a separate cache-key byte, alias parsing, and capabilities advertising. Three review findings flagged real issues with that path — non-IANA MIME strings in `<Format>`, a weakened coverage test, and an empty-tile shortcut that silently dropped PNG8 — and the reviewer's question ("do we really need GeoServer's two-format split?") was the right one.

**This PR now folds PNG8 into the regular `image/png` path.** `encode_png` auto-selects:

| Buffer's distinct RGBA colours | Encoding |
|---|---|
| ≤256 (every colormap layer) | 8-bit indexed-palette PNG (`PLTE` + optional `tRNS`) |
| >256 (gradients, false-colour) | 32-bit RGBA PNG (previous behaviour) |

Content-type is `image/png` for both branches. **No API surface added:** no `FORMAT=image/png; mode=8bit`, no `image/png8` alias, no `ImageFormat::Png8`, no extra cache-key byte, no normalize-format helper, no parser tests for new aliases. Every existing client immediately gets the bandwidth win with zero config.

## Why this is safe

- **Pixels are byte-identical** — the indexed encoder only triggers when the exact palette fits in ≤256 entries (no quantisation). GeoServer's two-format split exists because their indexed path *can* be lossy; ours isn't.
- **No client compatibility break** — content-type stays `image/png`, and ETags are content-derived so caches invalidate themselves on the first request after redeploy.
- **Bounded encode cost** — single-pass HashMap scan with early-exit at 257 distinct colours; a few ms on a 1024×1024 frame.
- **ETag stability preserved** (#145 invariant) — palette ordering by first pixel-occurrence is deterministic; two encodes of the same buffer produce byte-identical output.

## Result

The 3× target from #252 is locked in by a unit test on an entropy-rich 16-colour 1024×1024 frame:

```
auto-PNG8 ≈ 320 KB → forced-RGBA ≈ 1.05 MB → 3.3×
```

On a real radar tile the ratio should match GeoServer's reported 3–4×.

## How the review findings dissolved

| Round-1 review finding | Resolution |
|---|---|
| WMS GetCapabilities advertises non-IANA MIME (`image/png; mode=8bit`, `image/png8`) | Capabilities goes back to just `image/png` |
| Coverage test weakened from strict 1-to-1 to loose containment | Strict round-trip restored — aliases removed |
| Empty-tile fast path silently ignores PNG8 | No PNG8 variant to ignore; empty tiles also auto-PNG8 |

## Diff size

8 files, ~362 added / ~9 removed. The bulk is `crates/render/src/encode.rs` (the encoder split and its tests). API param files only get an updated `SUPPORTED_FORMATS` docstring + minor OpenAPI description tweak.

## Tested

Five new ds-render encode tests exercising the public `encode_png`:

- `encode_png_auto_palettes_when_under_256_colors` — asserts `ColorType::Indexed` on the encoded output and pixel-identical RGBA round-trip.
- `encode_png_auto_palette_is_at_least_three_times_smaller` — pits auto-PNG8 vs the forced RGBA path on entropy-rich input, locks in the ≥3× ratio.
- `encode_png_is_deterministic` — same input → same bytes (ETag invariant preserved).
- `encode_png_falls_back_to_rgba_above_256_colors` — 257-colour input produces a `ColorType::Rgba` PNG.
- `encode_png_writes_trns_only_when_palette_has_alpha` — `tRNS` chunk omitted for fully-opaque palettes.

Plus the existing metatile geography test was updated to decode both `Indexed` and `Rgba` PNGs since auto-PNG8 may flip the branch depending on the synthetic test pattern's palette.

`cargo fmt --all` clean, `cargo clippy -p ds-render -p api-wms -p api-maps -p api-tiles -- -D warnings` clean, `cargo check --workspace` clean, `cargo test -p ds-render -p api-wms -p api-maps -p api-tiles` all green.

### DIFF
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -485,7 +485,7 @@ pub async fn api_definition(State(state): State<AppState>) -> impl IntoResponse 
                         "default": "image/png",
                         "enum": ["image/png", "image/jpeg", "image/webp"]
                     },
-                    "description": "Output format"
+                    "description": "Output format. `image/png` auto-emits an 8-bit indexed-palette PNG (~3–4× smaller) for colormap-rendered layers; falls back to 32-bit RGBA above 256 distinct colours."
                 },
                 "bbox-crs": {
                     "name": "bbox-crs",
--- a/crates/api-maps/src/params.rs
+++ b/crates/api-maps/src/params.rs
@@ -16,6 +16,12 @@ pub const MAX_MAP_DIMENSION: u32 = 8000;
 const SUPPORTED_CRS: &[&str] = &["CRS:84", "EPSG:4326", "EPSG:3857", "EPSG:3067", "EPSG:3035"];
 
 /// Supported output formats.
+///
+/// `image/png` auto-selects an 8-bit indexed-palette encoding ("PNG8") when
+/// the rendered image carries ≤256 distinct colours (every colormap layer);
+/// the encoder falls back to 32-bit RGBA above that. Content-type is
+/// `image/png` either way — clients can't tell, and no second `f=` value is
+/// needed.
 const SUPPORTED_FORMATS: &[&str] = &["image/png", "image/jpeg", "image/webp"];
 
 /// Query parameters for OGC API Maps get_map / get_styled_map endpoints.
--- a/crates/api-tiles/src/handlers.rs
+++ b/crates/api-tiles/src/handlers.rs
@@ -479,7 +479,7 @@ pub async fn api_definition(State(state): State<AppState>) -> impl IntoResponse 
                             "application/vnd.mapbox-vector-tile"
                         ]
                     },
-                    "description": "Output format. `mvt` selects Mapbox Vector Tile (only on collections with a FeatureEngine)."
+                    "description": "Output format. `image/png` auto-emits an 8-bit indexed-palette PNG (~3–4× smaller) for colormap-rendered layers; falls back to 32-bit RGBA above 256 distinct colours. `mvt` selects Mapbox Vector Tile (only on collections with a FeatureEngine)."
                 },
                 "elevation": {
                     "name": "elevation",
--- a/crates/api-tiles/src/params.rs
+++ b/crates/api-tiles/src/params.rs
@@ -21,6 +21,11 @@ pub const TILE_SIZE: u32 = 256;
 pub const MAX_FEATURES_PER_TILE: usize = 50_000;
 
 /// Supported raster output formats for map tiles.
+///
+/// `image/png` auto-selects an 8-bit indexed-palette encoding ("PNG8") when
+/// the rendered tile carries ≤256 distinct colours (every colormap layer);
+/// the encoder falls back to 32-bit RGBA above that. Content-type is
+/// `image/png` either way — no second `f=` value is needed.
 const SUPPORTED_FORMATS: &[&str] = &["image/png", "image/jpeg", "image/webp"];
 
 /// MVT format aliases. Either form works in `?f=` for clients that prefer
--- a/crates/api-wms/src/params.rs
+++ b/crates/api-wms/src/params.rs
@@ -21,6 +21,12 @@ const SUPPORTED_CRS: &[&str] = &["CRS:84", "EPSG:4326", "EPSG:3857", "EPSG:3067"
 /// Single source of truth: capabilities iterates this slice when emitting
 /// `<Format>` children, and `parse_image_format` mirrors it. If you add a
 /// format here, extend the match in `parse_image_format` to match.
+///
+/// `image/png` auto-selects an 8-bit indexed-palette encoding ("PNG8") when
+/// the rendered image carries ≤256 distinct colours (every colormap layer);
+/// the encoder falls back to 32-bit RGBA above that. Content-type is
+/// `image/png` either way — no second `FORMAT=` value is needed because the
+/// choice is invisible to clients.
 pub const SUPPORTED_FORMATS: &[&str] = &["image/png", "image/jpeg", "image/webp"];
 
 /// Parse a WMS `FORMAT=` query parameter into the corresponding `ImageFormat`.
--- a/crates/render/src/encode.rs
+++ b/crates/render/src/encode.rs
@@ -470,12 +470,28 @@ mod tests {
         // ETag stability hinges on byte-identical output for byte-identical
         // input. Two encodes of the same buffer must produce the same bytes —
         // no encoder-side timestamps or randomisation.
+        //
+        // The pattern is constrained to 16 distinct colours so this test
+        // exercises the **indexed-palette path** specifically — that path's
+        // determinism is *our* invariant (palette ordered by first
+        // pixel-occurrence). The RGBA fallback's determinism is the `png`
+        // crate's responsibility and is incidentally covered elsewhere.
         let rgba = rgba_from(64, 64, |x, y| {
-            [((x * 4) & 0xFF) as u8, ((y * 4) & 0xFF) as u8, 0, 255]
+            [((x % 4) * 64) as u8, ((y % 4) * 64) as u8, 0, 255]
         });
         let a = encode_png(&rgba, 64, 64).unwrap();
         let b = encode_png(&rgba, 64, 64).unwrap();
         assert_eq!(a, b, "PNG encoding must be deterministic for a given input");
+        // Guard against a silent regression: if this test ever falls back to
+        // RGBA (e.g. someone widens the colour range above 256), it would
+        // stop testing the invariant it claims to.
+        let decoder = png::Decoder::new(&a[..]);
+        let reader = decoder.read_info().unwrap();
+        assert_eq!(
+            reader.info().color_type,
+            png::ColorType::Indexed,
+            "this test must exercise the indexed-palette path, not the RGBA fallback"
+        );
     }
 
     #[test]
--- a/crates/render/src/encode.rs
+++ b/crates/render/src/encode.rs
@@ -137,8 +137,11 @@ fn encode_png_indexed(
         encoder.set_depth(png::BitDepth::Eight);
         encoder.set_compression(png::Compression::Fast);
         encoder.set_palette(plte);
-        if let Some(ref t) = trns {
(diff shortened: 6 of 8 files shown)


mrauhala:
Force-pushed a rewrite that drops the two-format design entirely. The reviewer's question ("do we really need GeoServer's split?") was the right one — answering "no" makes all three review findings dissolve. Receipts:

**1. WMS GetCapabilities advertises non-IANA MIME — fixed.** `SUPPORTED_FORMATS` is back to `["image/png", "image/jpeg", "image/webp"]`. No `image/png; mode=8bit` or `image/png8` entries, no spec violation.

**2. Coverage test weakened — fixed.** `parse_image_format_covers_every_supported_format` restored to the strict `parsed.content_type() == *fmt` 1-to-1 assertion. The cross-wire-typo regression it was originally guarding against is detectable again.

**3. Empty-tile shortcut silently ignores PNG8 — fixed.** There is no separate PNG8 format to ignore. The empty-tile fast path now also benefits from auto-PNG8 (single-colour palette → very small indexed PNG); content-type stays `image/png`.

PNG and PNG8 are no longer distinct cache entries because they're no longer distinct formats — `encode_png` auto-selects internally, content-derived ETag guarantees correctness across the (now invisible) branch switch.

Encoding contract verified end-to-end:
- ≤256 colours → `ColorType::Indexed` PNG (asserted in `encode_png_auto_palettes_when_under_256_colors`)
- >256 colours → `ColorType::Rgba` PNG (asserted in `encode_png_falls_back_to_rgba_above_256_colors`)
- Deterministic bytes for same input (ETag invariant — `encode_png_is_deterministic`)
- ≥3× smaller than forced-RGBA on entropy-rich input (`encode_png_auto_palette_is_at_least_three_times_smaller`)

mrauhala:
Addressed in 05d2ae3.

**Fix:** `encode_png_is_deterministic` now uses a 16-colour `((x % 4) * 64, (y % 4) * 64, 0, 255)` pattern instead of the 4096-colour `(x*4, y*4, 0, 255)` one, so the test exercises `encode_png_indexed` — the path whose determinism is *our* invariant (first-occurrence palette ordering). Added a `ColorType::Indexed` guard assertion on the decoded PNG so a future widening of the test's colour range can't silently fall back to RGBA and re-introduce the same coverage gap.

The RGBA fallback's determinism is the `png` crate's responsibility and is incidentally covered by `encode_png_falls_back_to_rgba_above_256_colors` (which already pins `ColorType::Rgba`).

mrauhala:
Addressed all three test-helper findings in 17373c9 (no production code touched):

**1. tRNS check scanned raw IDAT bytes — fixed.** `encode_png_writes_trns_only_when_palette_has_alpha` now decodes the header and checks `info().trns.is_some()` via a `png_has_trns` helper, instead of `bytes.windows(4).any(|w| w == b"tRNS")`. No more chance of a spurious match inside compressed image data on a larger fixture.

**2. `unwrap_or(&[])` → unchecked indexing — fixed.** Both `decode_png_to_rgba` (encode.rs) and `decode_rgba` (metatile.rs) now use `.expect("indexed PNG must carry a PLTE chunk")`, so a truncated/foreign indexed PNG fails with a legible message rather than an opaque OOB panic on the next line.

**3. Metatile branch coverage gap — fixed.** `assembles_solid_region_and_caches_tiles` now pins `ColorType::Indexed` on the ≤256-colour SolidRed mosaic, so a regression where `encode_png_indexed` is always bypassed (silently returning the RGBA fallback) is caught in the assembly path too — not only in the encoder unit tests.

`cargo test -p ds-render` green (56 tests), `cargo fmt`/`clippy -D warnings` clean.

mrauhala:
Applied finding #2 in 08af81c.

`encode_png`'s length guard and the two `encode_png_indexed` capacity hints now cast to `usize` before multiplying (`(width as usize) * (height as usize) [* 4]`), so the product can't wrap u32. Unreachable via the server (API caps at 8000²/64M px), but closes the exposure for non-API callers (fuzz/tests/future engine code).

Scoped to the PNG path this PR adds — the identical pre-existing pattern in `encode_jpeg`/`encode_webp` is left as-is (out of scope here; can be a separate `fix:` if wanted).

Finding #1 (metatile indexed-decode pixel round-trip) left unaddressed by request — the production indexed-encode path is already pixel-round-trip covered by `encode_png_auto_palettes_when_under_256_colors`; #1 is test-helper thoroughness only.

`cargo test -p ds-render` green (56), fmt/clippy clean.

mrauhala:
Triaged the three findings against a "production code only" gate:

**Finding 2 — unnecessary `tRNS.clone()` — FIXED (a5ffbae).** This is the one genuine shipping-code item: a per-tile allocation on the hot path (fires on ≈ every transparent radar tile). `trns` drops right after the block, so moving it in is strictly better and zero-risk.

**Finding 1 — `encode_jpeg`/`encode_webp` still use the u32-multiply pattern — deferred (scope).** Correct that the same hardening applies, but this is *pre-existing* code unrelated to PNG8 that I deliberately scoped out of this PR. Same production-reachability story as the PNG fix: unreachable via the API layer (8000²/64M-px caps), only exposed to hypothetical non-API callers, and there is no jpeg/webp encoder fuzz target today. Will land as a separate `fix:` so this PR stays PNG8-focused — unless you'd prefer it folded in here.

**Finding 3 — exact 256-colour boundary untested — deferred (test-only).** A `>`-vs-`>=` off-by-one in the palette guard would be invisible to the current suite (tests use 16 and 257 colours, not exactly 256). Legitimate test-hardening but not a production problem; happy to add the boundary test if wanted.

`cargo test -p ds-render` green (56), fmt/clippy clean.

PR #261, opened 2026-05-28
feat(api): align OGC API Maps/Tiles collection metadata with the spec

### LINKED ISSUE #255: Remove non-standard `apis` array from collection JSON
The vendor `apis: ["maps","tiles",...]` array is echoed into every collection JSON (maps, tiles, edr, features). It is not part of any OGC schema — standard clients ignore it, and discoverability is supposed to come from typed `links` (`map`, `tilesets-map`, `data`, ...).

**Scope:** Remove `"apis": config.apis` from the collection JSON in api-maps, api-tiles, api-edr, api-features. The `apis` *config* field and the `server/src/admin.rs` routing logic stay untouched — only response serialization drops it.

**Affected serializers:**
- `crates/api-maps/src/handlers.rs`
- `crates/api-tiles/src/handlers.rs`
- `crates/api-edr/src/handlers.rs`
- `crates/api-features/src/handlers.rs`

Delete/repurpose the five `*_exposes_apis_array` tests. Leave `config.apis`-gated behavior (preview discovery, tileset-link gating) intact.

Part of OGC API – Maps spec-compliance work (see plan).

### LINKED ISSUE #256: Add spatial extent resolution (`extent.spatial.grid`) to Maps/Tiles collections
`extent.spatial` currently carries only `bbox` + `crs`. OGC API Common Part 2 defines a per-axis `grid` (`cellsCount` + `resolution`) that we can advertise from the native grid dimensions already available at the engine level.

**Approach:**
- Add `grid_size: Option<[u32;2]>` (`[x_cells, y_cells]`) to `RasterInfo` (`crates/core/src/map_engine.rs`).
- Populate it: geotiff (`TiffMetadata.width/height`), querydata (`grid.nx/ny`), odim-composite (`seed_xsize/ysize`). `None` for odim-PVOL (polar, no regular CRS84 grid) and grib (ni/nj not cheaply available pre-decode — follow-up).
- Serialize `spatial.grid` in api-maps + api-tiles: `[{cellsCount: nx, resolution: (E-W)/nx}, {cellsCount: ny, resolution: (N-S)/ny}]` (resolution in CRS84 degrees), emitted only when both `grid_size` and `spatial_extent` are present.

Introduces the shared private `build_extent` helper refactor (also used by the temporal/vertical issues).

Part of OGC API – Maps spec-compliance work (see plan).

### LINKED ISSUE #257: Add temporal extent resolution (`extent.temporal.grid`) to Maps/Tiles collections
`extent.temporal` currently carries only `interval` + `trs`. The regular time step (e.g. 5-minute radar) is computable from the existing `times` vector but never advertised.

**Approach (API layer, no engine change):**
- Add `format_iso8601_duration(seconds) -> Option<String>` to `crates/core/src/datetime.rs` (inverse of the existing `parse_iso8601_duration`).
- In `build_extent` (api-maps + api-tiles): diff successive `times`; if all diffs equal (1s tolerance) emit `grid: {cellsCount, resolution: "PT5M"}`, else emit `grid: {cellsCount, coordinates: [rfc3339...]}`. Omit `grid` for ≤1 timestamp.

Depends on the `build_extent` refactor from the spatial-grid issue.

Part of OGC API – Maps spec-compliance work (see plan).

### LINKED ISSUE #258: Upgrade vertical extent to OGC Common Part 2 form (`unit` + `grid.coordinates`)
Collections with a vertical axis (radar PVOL elevation sweeps) emit `{interval, values}`, which is the OGC EDR extent shape but not the OGC API Common Part 2 additional-dimension form used by Maps.

**Approach (additive, backward compatible):** keep `interval` + `values`; add `unit` (from `VerticalDimension::unit()`) and `grid: {coordinates: levels}` in api-maps + api-tiles `build_extent`.

`vrs` is deferred — the only vertical kind currently exposed is radar elevation angle, which has no standard OGC vertical-CRS URI. Revisit when a pressure/height vertical axis is exposed.

Depends on the `build_extent` refactor.

Part of OGC API – Maps spec-compliance work (see plan).

### LINKED ISSUE #259: Advertise map tilesets: `tilesets-map` link + map-tilesets conformance class
A maps collection has no link to its rendered (raster) tiles. In OGC terms those are *map tilesets*, discoverable from the maps collection via the `tilesets-map` relation. `/tiles` stays a standalone API; we only add discovery.

**Approach:**
- In `crates/api-maps/src/handlers.rs` `build_collection_metadata`, push a link gated on `config.apis` containing `"tiles"` (mirrors the `tilesets-vector` precedent in api-features):
  `{ href: "{base}/tiles/collections/{id}/tiles", rel: "http://www.opengis.net/def/rel/ogc/1.0/tilesets-map", type: "application/json", title: "Map tilesets" }`
- Add `http://www.opengis.net/spec/ogcapi-maps-1/1.0/conf/tilesets` to `conformance()`.
- Tests: link present with `apis:["maps","tiles"]` fixture, absent otherwise; conformance URI present.

Part of OGC API – Maps spec-compliance work (see plan).

### LINKED ISSUE #260: Maps standards polish: add `storageCrs` to collection metadata
OGC API Common collection metadata supports `storageCrs` (the native CRS of the data). api-features already emits it; api-maps does not, despite knowing `RasterInfo.native_crs`.

**Approach:** add `"storageCrs": crs_to_uri(&info.native_crs)` in `crates/api-maps/src/handlers.rs` `build_collection_metadata` (reuse the existing `crs_to_uri` helper). Keep the existing short link rels (`self`, `map`, `styles`, `conformance`, `data`) — they're accepted and test-covered.

Explicitly **not** adding `minScaleDenominator`/`maxScaleDenominator`: there is no defensible display-pixel-size convention for arbitrary `bbox/width/height` map requests.

Part of OGC API – Maps spec-compliance work (see plan).

mrauhala:
## What & why

Brings the OGC API – Maps (and the related Tiles) collection metadata into line with [OGC API – Maps Part 1 (20-058)](https://docs.ogc.org/is/20-058/20-058.html) and the OGC API Common Part 2 extent model. All changes are **additive** except removing the non-standard `apis` field.

## Changes

- **Drop `apis` from collection JSON** (maps, tiles, edr, features) — it has no OGC schema; discovery is via typed `links`. The `apis` config field, admin routing, and the internal `/preview` SPA manifest are unchanged. (#255)
- **`extent.spatial.grid`** — per-axis `cellsCount` + CRS84-degree `resolution`, from a new `RasterInfo.grid_size` populated by geotiff/querydata/odim-composite (`None` for grib and polar odim-PVOL). (#256)
- **`extent.temporal.grid`** — regular `{cellsCount, resolution}` (ISO 8601 duration) or `{cellsCount, coordinates}` fallback, computed from the `times` vector via a new `ds_core::datetime::format_iso8601_duration`. (#257)
- **Vertical extent → Common Part 2** — adds `unit` + `grid.coordinates` alongside the retained `interval`/`values`. `vrs` deferred (radar elevation angle has no standard OGC vertical-CRS URI). (#258)
- **Map tilesets** — `tilesets-map` link (gated on the `tiles` API) + `conf/tilesets` conformance class. `/tiles` stays a standalone router. (#259)
- **`storageCrs`** on maps collection metadata. (#260)

## Verification

- `cargo fmt`, `cargo clippy --workspace --all-targets` clean; full `cargo test --workspace` green (48 suites, 0 failures).
- New unit/integration tests: spatial/temporal grid, vertical Common Part 2 form (new mock with elevation axis), `storageCrs`, `tilesets-map` present/absent, conformance class, `apis` absent across all four APIs.
- **Live smoke test** against real local fixtures:
  - EU OPERA composite → `has("apis")=false`, real `spatial.grid` (3800×4400), `temporal.grid` auto-detected `PT5M`, `tilesets-map` link, `storageCrs`.
  - FMI PVOL volume → vertical extent with real elevation angles (`unit: "deg"`, `grid.coordinates`), `spatial.grid` correctly `null` (no fabricated polar grid).

## Out of scope / follow-ups

- GRIB `grid_size` (needs ni/nj cached pre-decode).
- `vrs` on vertical extent (until a pressure/height axis is exposed).
- Porting the grid descriptors to the EDR extent.

Closes #255
Closes #256
Closes #257
Closes #258
Closes #259
Closes #260

### DIFF
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -769,7 +769,6 @@ fn build_collection_metadata(
         "id": config.id,
         "title": config.title,
         "description": config.description,
-        "apis": config.apis,
         "links": [
             {
                 "href": format!("{base_url}/edr/collections/{}", config.id),
--- a/crates/api-edr/tests/ogc_edr_api_tests.rs
+++ b/crates/api-edr/tests/ogc_edr_api_tests.rs
@@ -564,10 +564,11 @@ mod collections {
     }
 
     #[tokio::test]
-    async fn collection_exposes_apis_array() {
+    async fn collection_omits_nonstandard_apis_field() {
+        // `apis` is a vendor extension with no OGC schema; it must not leak
+        // into the standard collection JSON.
         let (_, json) = get("/collections/weather").await;
-        let apis = json["apis"].as_array().expect("apis must be present");
-        assert!(apis.iter().any(|a| a == "edr"));
+        assert!(json.get("apis").is_none() || json["apis"].is_null());
     }
 }
 
--- a/crates/api-edr/tests/ogc_edr_api_tests.rs
+++ b/crates/api-edr/tests/ogc_edr_api_tests.rs
@@ -568,7 +568,10 @@ mod collections {
         // `apis` is a vendor extension with no OGC schema; it must not leak
         // into the standard collection JSON.
         let (_, json) = get("/collections/weather").await;
-        assert!(json.get("apis").is_none() || json["apis"].is_null());
+        assert!(
+            json.get("apis").is_none(),
+            "apis must not be present in the standard collection JSON"
+        );
     }
 }
 
--- a/crates/api-features/src/handlers.rs
+++ b/crates/api-features/src/handlers.rs
@@ -489,7 +489,6 @@ fn build_collection_metadata(
         "id": config.id,
         "title": config.title,
         "description": config.description,
-        "apis": config.apis,
         "itemType": "feature",
         "crs": [
             "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
--- a/crates/api-features/tests/ogc_features_api_tests.rs
+++ b/crates/api-features/tests/ogc_features_api_tests.rs
@@ -392,10 +392,11 @@ mod collections {
     }
 
     #[tokio::test]
-    async fn collection_detail_exposes_apis_array() {
+    async fn collection_detail_omits_nonstandard_apis_field() {
+        // `apis` is a vendor extension with no OGC schema; it must not leak
+        // into the standard collection JSON.
         let (_, json) = get("/collections/cities").await;
-        let apis = json["apis"].as_array().expect("apis must be present");
-        assert!(apis.iter().any(|a| a == "features"));
+        assert!(json.get("apis").is_none() || json["apis"].is_null());
     }
 
     #[tokio::test]
@@ -455,11 +456,12 @@ mod vector_tile_discovery {
     }
 
     #[tokio::test]
-    async fn collection_with_tiles_api_advertises_apis_array() {
+    async fn collection_with_tiles_api_still_omits_apis_array() {
+        // Even when the operator wires both Features and Tiles, the vendor
+        // `apis` field stays out of the standard collection JSON — discovery
+        // is via the `tilesets-vector` link asserted below.
         let json = fetch("/collections/cities").await;
-        let apis = json["apis"].as_array().expect("apis must be present");
-        assert!(apis.iter().any(|a| a == "features"));
-        assert!(apis.iter().any(|a| a == "tiles"));
+        assert!(json.get("apis").is_none() || json["apis"].is_null());
     }
 
     #[tokio::test]
--- a/crates/api-features/tests/ogc_features_api_tests.rs
+++ b/crates/api-features/tests/ogc_features_api_tests.rs
@@ -396,7 +396,10 @@ mod collections {
         // `apis` is a vendor extension with no OGC schema; it must not leak
         // into the standard collection JSON.
         let (_, json) = get("/collections/cities").await;
-        assert!(json.get("apis").is_none() || json["apis"].is_null());
+        assert!(
+            json.get("apis").is_none(),
+            "apis must not be present in the standard collection JSON"
+        );
     }
 
     #[tokio::test]
@@ -461,7 +464,10 @@ mod vector_tile_discovery {
         // `apis` field stays out of the standard collection JSON — discovery
         // is via the `tilesets-vector` link asserted below.
         let json = fetch("/collections/cities").await;
-        assert!(json.get("apis").is_none() || json["apis"].is_null());
+        assert!(
+            json.get("apis").is_none(),
+            "apis must not be present in the standard collection JSON"
+        );
     }
 
     #[tokio::test]
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -167,7 +167,7 @@ fn build_collection_metadata(
     // Engines label projected/rotated grids with internal names ("TM",
     // "LAEA", "projected", "rotated_ll", …) that have no URI; emitting CRS84
     // for those would mislabel the storage grid, so omit it instead.
-    if let Some(storage_crs) = native_crs_to_uri(&info.native_crs) {
+    if let Some(storage_crs) = ds_core::geo::native_crs_uri(&info.native_crs) {
         metadata["storageCrs"] = json!(storage_crs);
     }
 
@@ -176,24 +176,6 @@ fn build_collection_metadata(
     }
 
     metadata
-}
-
-/// Map an engine's native CRS label to an OGC URI for `storageCrs`, or `None`
-/// when there is no stable URI for it. Unlike [`crs_to_uri`] (which serves the
-/// `Content-Crs` header for request-validated CRS values and defaults to
-/// CRS84), this never falls back: engines populate `RasterInfo.native_crs`
-/// with internal labels ("TM", "LAEA", "projected", "rotated_ll", …) for
-/// projections that have no canonical URI, and a wrong `storageCrs` is worse
-/// than an absent one.
-fn native_crs_to_uri(crs: &str) -> Option<&'static str> {
-    match crs {
-        "CRS:84" => Some("http://www.opengis.net/def/crs/OGC/1.3/CRS84"),
-        "EPSG:4326" => Some("http://www.opengis.net/def/crs/EPSG/0/4326"),
-        "EPSG:3857" => Some("http://www.opengis.net/def/crs/EPSG/0/3857"),
-        "EPSG:3067" => Some("http://www.opengis.net/def/crs/EPSG/0/3067"),
-        "EPSG:3035" => Some("http://www.opengis.net/def/crs/EPSG/0/3035"),
-        _ => None,
-    }
 }
 
 /// Build the OGC API Common Part 2 `extent` object (spatial, temporal,
@@ -275,22 +257,11 @@ fn temporal_grid(times: &[chrono::DateTime<chrono::Utc>]) -> Option<serde_json::
         return None;
     }
     let cells = times.len();
-    let gaps: Vec<i64> = times
-        .windows(2)
-        .map(|w| (w[1] - w[0]).num_seconds())
-        .collect();
-    let min = *gaps.iter().min().unwrap();
-    let max = *gaps.iter().max().unwrap();
-    // Regular when every gap agrees within a 2-second spread (absorbs ±1 s
-    // rounding/leap-second jitter on any interval, not just the first) and is
-    // strictly positive. Anchor the advertised resolution to the rounded mean
-    // step so a jittered endpoint can't bias it.
-    if min > 0 && max - min <= 2 {
-        let count = gaps.len() as i64;
-        let avg = (gaps.iter().sum::<i64>() + count / 2) / count;
-        if let Some(resolution) = ds_core::datetime::format_iso8601_duration(avg) {
-            return Some(json!({ "cellsCount": cells, "resolution": resolution }));
-        }
+    // Regular series -> a single ISO 8601 resolution; irregular -> the full
+    // coordinates list. Regularity detection lives in ds-core so the Maps and
+    // Tiles builders can't drift apart.
+    if let Some(resolution) = ds_core::datetime::regular_step_duration(times) {
+        return Some(json!({ "cellsCount": cells, "resolution": resolution }));
     }
     let coordinates: Vec<String> = times.iter().map(|t| t.to_rfc3339()).collect();
     Some(json!({ "cellsCount": cells, "coordinates": coordinates }))
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -241,16 +241,22 @@ fn build_extent(info: &ds_core::map_engine::RasterInfo) -> Option<serde_json::Va
     // omitted: the only kind in use (radar elevation angle) has no standard
     // OGC vertical-CRS URI.
     if let Some(vertical) = &info.vertical {
-        let levels = &vertical.levels;
-        extent.insert(
-            "vertical".to_string(),
-            json!({
-                "interval": vertical.extent().map(|(lo, hi)| [[lo, hi]]),
-                "values": levels,
-                "unit": vertical.unit(),
-                "grid": { "coordinates": levels }
-            }),
-        );
+        // Only emit the vertical extent when there are levels: OGC API Common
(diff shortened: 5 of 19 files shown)


mrauhala:
Thanks — both bugs were real and are fixed in e03dcb1.

**Bug 1 — `storageCrs` mislabelling projected grids.** Confirmed: my own smoke test showed the OPERA composite (projected) reporting `storageCrs: …/CRS84`. Added a dedicated `native_crs_to_uri(&str) -> Option<&'static str>` that maps only the CRS values with a stable OGC URI (CRS84 + the EPSG codes) and returns `None` otherwise; `build_collection_metadata` now omits `storageCrs` rather than emitting a wrong CRS84. `crs_to_uri` (the `Content-Crs` header path, fed request-validated values) is left unchanged. New test: `storage_crs_omitted_for_projected_native_crs` (native CRS `"LAEA"`).

**Bug 2 — `temporal_grid` anchored to the first gap.** Switched both api-maps and api-tiles to a spread check: regular iff `max_gap − min_gap ≤ 2 s` (and `min_gap > 0`) across all gaps, and the advertised resolution is now the rounded **mean** step so a jittered endpoint can't bias it. New test: `temporal_grid_treats_jittered_series_as_regular` (gaps 3599/3601/3600 s → `PT1H`, no `coordinates` fallback).

`cargo fmt` + `cargo clippy --all-targets` clean; api-maps/api-tiles suites green.

mrauhala:
Round 2 addressed in 77f83d5 — all three findings were valid.

**Bug — rotated-lat-lon GeoTIFF `storageCrs`.** Correct, and a direct consequence of the round-1 fix. `engine-geotiff` labelled `Crs::RotatedLatLon` as `"EPSG:4326"` (harmless when no `storageCrs` was emitted, now mislabelling). Changed it to `"rotated_ll"` to match `engine-querydata`, so `native_crs_to_uri` returns `None` and `storageCrs` is omitted.

**Test gap — `api-tiles` jitter path.** Added `temporal_grid_jitter::temporal_grid_treats_jittered_series_as_regular` to api-tiles, mirroring the api-maps test (the function is duplicated across both crates).

**Latent — `"interval": null` for empty `VerticalDimension`.** `build_extent` now only emits the vertical extent when `vertical.extent()` is `Some`, so an empty `VerticalDimension` omits the object entirely rather than emitting a null `interval`. Fixed in both crates; new test `empty_vertical_dimension_omits_extent`.

`cargo fmt` + `cargo clippy --all-targets` clean; api-maps (71), api-tiles (71), engine-geotiff (74) suites green.

mrauhala:
Round 3 addressed in a385b58 — all four findings were valid. (`native_crs` is informational: its only consumers are `storageCrs` and one querydata test, so relabelling is behaviourally safe.)

**1. `Crs::Wgs84` → `"CRS:84"` (was `"EPSG:4326"`).** Fixed in engine-geotiff and engine-querydata; the data is lon-first, so the lat-first EPSG:4326 URI would invite axis-swapping. `native_crs_uri` keeps distinct entries for both labels.

**2. TM/LAEA over-claiming.** Extracted a testable `crs_label(&Crs)` in engine-geotiff that only emits `EPSG:3067` when params match TM35FIN and `EPSG:3035` when they match ETRS89-LAEA Europe; any other zone gets a generic label (`"TM"`/`"LAEA"`) → `storageCrs` omitted. Tests cover TM35FIN vs UTM-33N and ETRS89 vs a non-matching LAEA.

**3. `api-tiles` missing `storageCrs`.** Mirrored the api-maps emit for raster collections; new test `raster_collection_advertises_storage_crs`.

**4. `temporal_grid` duplication.** Moved the regularity detection to `ds_core::datetime::regular_step_duration` (returns the ISO 8601 step, or `None` for irregular/short) and added `ds_core::geo::native_crs_uri` as the single label→URI source of truth. Both api crates now call the shared helpers; the JSON assembly (which can't live in ds-core) stays thin in each. ds-core unit tests cover both.

`cargo fmt` + `clippy --all-targets` clean; full `cargo test --workspace` green (48 suites). Updated the querydata assertion to `"CRS:84"`.

mrauhala:
Round 4 addressed in e9f4076.

**1. Divergent CRS labels across engines (confirmed).** Aligned engine-geotiff, engine-odim, and engine-querydata on one descriptive vocabulary — `CRS:84` / `TM` / `LAEA` / `LCC` / `stere` / `rotated_ll` — so `ds_core::geo::native_crs_uri` maps all three identically. engine-geotiff keeps its param-checked EPSG *upgrade* (TM35FIN→3067, ETRS89-LAEA→3035) layered on top; engine-odim stays generic (spherical composites). Tests: geotiff `crs_label_lcc_and_stereographic_match_odim_vocabulary`, odim's variant test updated, querydata `Stereographic`→`stere`. I also filed #262 to make this structural — one shared label helper in `ds_core::geo` so the three match-arms can't drift again.

**2. Overly permissive `apis` assertions (confirmed).** Tightened all four/five to `assert!(json.get(\"apis\").is_none(), …)` — `\"apis\": null` no longer passes.

**3. Move `temporal_grid` wrapper to ds-core (won't do, with reason).** The wrapper builds `serde_json::Value`, which the project architecture rule deliberately keeps out of `ds-core` (\"serialization belongs in the API crates\"; ds-core deps are limited to chrono/serde/thiserror/toml). The *logic* that could meaningfully drift already lives in ds-core (`regular_step_duration` + `native_crs_uri`); the remaining ~10 lines are irreducible JSON glue. Happy to revisit if you'd prefer a typed `TemporalGrid` struct in ds-core that the API layer serializes, but that's more surface for the same outcome.

`cargo fmt` + `clippy --all-targets` clean; full `cargo test --workspace` green (48 suites).

mrauhala:
Round 5 addressed in bbb5752.

**1. GRIB `EPSG:4326` → `CRS:84` (confirmed).** The one WGS84 engine I'd missed in round 3; now consistent with geotiff/odim/querydata.

**2. `conf/tilesets` conformance (confirmed concern, corrected diagnosis).** I verified against the spec: `http://www.opengis.net/spec/ogcapi-maps-1/1.0/conf/tilesets` *is* a real OGC API Maps "Map Tilesets" requirements class — the suggested `ogcapi-tiles-1/.../conf/tilesets-list` is the **Tiles** spec's class, not the right substitute. **However** the underlying instinct is right: that class's abstract tests require maps-native `/collections/{id}/map/tiles` endpoints, which we deliberately don't implement (raster tiles come from the standalone OGC API Tiles service, discovered via the `tilesets-map` link). Declaring it was a false conformance claim, so I **removed the declaration** and kept the discovery link. Test now asserts the class is *absent*.

**3. Negative resolution across the anti-meridian (confirmed).** Added `ds_core::geo::crs84_bbox_spans` (wrap-aware; `east < west` → a 20° span, not −340°) and use it in both api-maps and api-tiles. Tests: ds-core unit + an api-maps integration regression guard.

**4. storageCrs tests asserting the wrong URI (confirmed).** Switched the mocks to `CRS:84` (what production WGS84 engines now emit) and assert the CRS84 URI; added a ds-core unit test covering the full `native_crs_uri` map (incl. the `EPSG:4326` entry) so that mapping stays covered.

`cargo fmt` + `clippy --all-targets` clean; full `cargo test --workspace` green (48 suites).

mrauhala:
Round 6 addressed in 3833738.

**1. Projected-CRS spatial.grid guard (confirmed).** Added `ds_core::geo::is_geographic_crs` and gate `extent.spatial.grid` on it in both crates. Note: I used `is_geographic_crs` rather than the suggested `native_crs_uri(...).is_some()` — the latter returns `Some` for the projected EPSG:3067/3035/3857 codes too, so it wouldn't actually suppress the grid for ODIM LAEA/GeoTIFF EPSG:3067 etc. `is_geographic_crs` matches only `CRS:84`/`EPSG:4326`, which are the cases where a CRS84-degree resolution is genuinely degree-regular. Projected collections now emit `bbox` but no `grid`.

**2. temporal_grid duplication (resolved).** Moved the builder into ds-core as a `#[derive(Serialize)] TemporalGrid` enum + `temporal_grid()`. `serde` is an allowed ds-core dependency (only `serde_json::Value` is barred), so the descriptor lives there and the API crates serialize it with `serde_json::to_value`. The verbatim wrappers are deleted from both api-maps and api-tiles — there's now a single definition, so the `/maps` and `/tiles` shapes can't drift.

**3. Jitter boundary (informational).** Added the caveat to `regular_step_duration`'s docstring (alternating sub-minute series; no current engine produces sub-minute cadence). No behaviour change.

`cargo fmt` + `clippy --all-targets` clean; full `cargo test --workspace` green (48 suites).

mrauhala:
Round 7 addressed in 7239a55 (all three were low-severity).

**1. Zero-span bbox → `resolution: 0.0` (fixed).** Extended the spatial-grid guard to `lon_span > 0.0 && lat_span > 0.0` in both crates, so a degenerate bbox emits no grid rather than an invalid `0.0` resolution.

**2. `native_crs_uri("EPSG:3857")` dead arm (documented).** Added a docstring note that the map is keyed on the engine's `native_crs` *label*, and that the `EPSG:4326`/`EPSG:3857` arms are forward-looking (no engine emits those labels today — WGS84 is `"CRS:84"`, there's no Web Mercator `Crs` variant). Kept the arms so a future Mercator engine "just works" once it emits the label.

**3. `build_extent` duplication (tracked as #263).** The shareable logic is already in ds-core (`TemporalGrid`, `native_crs_uri`, `is_geographic_crs`, `crs84_bbox_spans`); what's left is JSON assembly, which the architecture rule keeps out of ds-core, and the two functions differ in signature (tiles has a vector `feature_extent` fallback). Rather than refactor a converged PR, I filed **#263** to model the whole extent as `serde` types in ds-core (same pattern as `TemporalGrid`) so both crates share one definition. Happy to fold it in here instead if you'd prefer.

`cargo fmt` + `clippy --all-targets` clean; scoped suites green (ds-core/api-maps/api-tiles).

mrauhala:
Round 8 addressed in a97ab4c (all four latent; correct for current engines).

**1. EPSG:4326 lon-first grid mismatch (fixed).** Renamed `is_geographic_crs` → `is_crs84_grid` and restricted it to `"CRS:84"` only. We always emit `spatial.grid` axes lon-first (axis 0 = lon) to match the CRS84 extent; EPSG:4326 is lat-first, so emitting our lon-first grid for it would break the OGC API Common Part 2 axis-order rule. No engine emits `"EPSG:4326"`, so this has no production effect — it just removes the latent trap. (For an EPSG:4326 source the grid lives under a CRS84 *extent* so it's arguably still fine, but restricting to CRS:84 removes all ambiguity at zero cost.)

**2. Mock labels (fixed).** Switched `VerticalMockEngine`/`EmptyVerticalMockEngine`/`JitteredTimesMockEngine`/`MultiParamMockEngine` (api-maps) and the two api-tiles mocks from `"EPSG:4326"` to `"CRS:84"`, matching `MockMapEngine` and the WGS84 invariant.

**3. Guard order (fixed).** api-maps now checks `is_crs84_grid` before computing `crs84_bbox_spans`, consistent with api-tiles.

**4. Jitter tolerance (kept, informational).** Holding the documented absolute 2 s tolerance: no engine produces sub-minute cadence, the docstring already flags the alternating-series edge, and switching to a relative tolerance just moves the boundary (e.g. 20 s data with ±1 s jitter would then misclassify). Best tuned against a real sub-minute engine if one is ever added.

`cargo fmt` + `clippy --workspace --all-targets` clean; ds-core/api-maps/api-tiles suites green.

mrauhala:
Round 9: no blocking issues — the review has converged. ✅

The one low-severity observation (STAC GeoTIFF cold-start: `extent.spatial.grid` is absent from the *first* collection-metadata response until the initial poll populates catalog entries) is self-healing and matches how all STAC-backed metadata bootstraps — no code change. Noting it alongside the existing follow-ups:

- **#256** — GRIB `grid_size` (ni/nj not known pre-decode).
- **#262** — unify the CRS/projection label vocabulary across engines.
- **#263** — model the whole OGC extent as ds-core `serde` types (finish the `build_extent` dedup).

All checks green; `cargo fmt` + `clippy --workspace --all-targets` clean; full `cargo test --workspace` green (48 suites). Closes #255–#260.

Issue #203, opened 2026-05-19
perf(engine-geotiff): per-pixel TM35FIN projection in resampling loop — ~12 transcendentals/pixel


mrauhala:
Part of #201 (GeoTIFF radar WMS performance epic).

## Problem

The WMS/Maps/Tiles resampling loop calls a full CRS projection **for every output pixel**. For the hot collection `fmi-radar-composite-dbz` (source CRS EPSG:3067 / TM35FIN), each pixel runs ~12 transcendental floating-point operations.

## Evidence

- Resampling loop: `crates/engine-geotiff/src/lib.rs:1375-1405` — `gt.world_to_pixel(lon, lat)` per output pixel
- `merc_y_to_lat` (`lib.rs:1631`): `exp()` + `atan()` = 2 ops
- `tm_forward` (`crates/ds-core/src/geo.rs:429-433`): `cos()`, `sin()`, `tan()`, `sqrt()` = 4 ops
- `meridian_arc(lat)` (`geo.rs:558-560`): 3× `sin()` = 3 ops
- `meridian_arc(lat0)` (`geo.rs:441`): 3× `sin()` = 3 ops — **all wasted** (see F3b below)
- **Total ≈ 12 transcendental ops/pixel ≈ 600 ns/pixel**
- CRS confirmed: `gdalinfo testdata/fmi-radar/20260406064000_fmi_radar_composite_dbz.tif` → `PROJCRS["EUREF-FIN / TM35FIN(E,N)", ID["EPSG",3067]]`, `lat_origin = 0°`

Cost: ~288 ms for 800×600, ~1.24 s for 1920×1080. The prod mean of 231 ms corresponds to an effective ~390×390 px average — consistent with mixed request sizes. This is the dominant per-render CPU cost and the main throughput gap vs GeoServer (which projects on a coarse grid).

## Impact

- Dominant CPU cost of every WMS/Maps/Tiles render for any projected-CRS source (TM35FIN, LAEA, LCC, Stereographic).
- Variable fullscreen output size → variable service time → fat-tail queueing latency.

## Suggested fix

[...]

PR #267, opened 2026-05-29
fix(wms,maps): render projected output CRS (EPSG:3067/3035) instead of blank tiles

### LINKED ISSUE #160: WMS: BBOX in EPSG:3067 / EPSG:3035 is treated as WGS84 by the engine
## Bug

The WMS \`GetCapabilities\` document advertises \`EPSG:3067\` and \`EPSG:3035\` as supported CRS (see [\`crates/api-wms/src/params.rs:17\`](../blob/main/crates/api-wms/src/params.rs#L17)). For both of these CRS, the WMS handler:

1. Accepts a \`BBOX\` in projected meters and returns it as-is in the validated \`[west, south, east, north]\` (no reprojection — [\`params.rs:277-290\`](../blob/main/crates/api-wms/src/params.rs#L277-L290)).
2. Sets \`output_crs = OutputCrs::Wgs84\` because the \`match\` only special-cases \`EPSG:3857\` ([\`params.rs:180-183\`](../blob/main/crates/api-wms/src/params.rs#L180-L183)).

\`MapEngine::get_raster_tile\` expects \`bbox\` to be \`[west, south, east, north]\` in WGS84 (degrees) — \`engine-geotiff\` then calls \`gt.bbox_to_pixels(west, south, east, north)\` and interpolates \`lon = west + frac_x * (east - west)\` as lon/lat degrees ([\`engine-geotiff/src/lib.rs:1261-1384\`](../blob/main/crates/engine-geotiff/src/lib.rs#L1261-L1384)).

In short: for \`CRS=EPSG:3067\` (or \`EPSG:3035\`), the engine receives projected meters (values in the millions) and treats them as WGS84 degrees. No reprojection happens. The result is either no data or a wildly wrong rendering.

The misleading code comment at \`params.rs:285-288\` claims \"the engine handles reprojection internally via GeoTransform::world_to_pixel which calls Crs::forward\" — that's not how the flow works.

## Reproduction

\`\`\`
GET /wms/?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&LAYERS=<radar-collection>&CRS=EPSG:3067&BBOX=100000,6700000,200000,6800000&WIDTH=512&HEIGHT=512&FORMAT=image/png
\`\`\`

## Expected vs. actual

- **Expected:** the bbox is reprojected to WGS84 (or the engine is told the output CRS so it can sample its source-CRS grid directly).
- **Actual:** the bbox is passed through unchanged, the engine treats the meters as lon/lat, and the response is garbage / nodata.

## Suggested fix

Either:

1. **Drop EPSG:3067/3035 from \`SUPPORTED_CRS\`** until proper handling exists (smallest change, but loses advertised functionality).
2. **Reproject the BBOX to WGS84** in \`api-wms/src/params.rs::parse_bbox\` for these two CRS, the same way \`EPSG:3857\` is already handled. This works because the engine resamples nearest-neighbor and a small WGS84 bbox approximating the projected one is good enough at most scales — but accuracy degrades at high latitudes.
3. **Extend \`OutputCrs\`** to carry the projected CRS so the engine can sample directly in projected coordinates. This is the correct fix but touches \`ds-core\`, every \`MapEngine\` impl, and the rendered-cache key.

Option (3) is the right long-term answer. (1) is fine as a stopgap if no one is actively using these CRS.

## Related

- The same issue applies to \`api-maps\` and \`api-tiles\` — both advertise \`EPSG:3067\`/\`EPSG:3035\` via the same \`SUPPORTED_CRS\` list and feed the engine projected meters with \`output_crs = Wgs84\`. The fix needs to land in all three API crates (or in a shared helper).
- Discovered while auditing the README WMS section against the implementation (#159).

### LINKED ISSUE #251: bug(api-wms): EPSG:3067 GetMap returns fully-transparent PNG for all radar layers (data extent intact in other CRSs)
WMS GetMap with `CRS=EPSG:3067` returns a fully-empty (zero-alpha) PNG for **every** Finland-bbox radar layer in production. The cache key, TIME parsing, and ETag flow are all correct — the render itself produces an empty image. Other CRSs work fine for the same layers/TIMEs/bboxes.

## Reproduction (production, 2026-05-27 ~12:00 UTC — precipitation active over Finland)

Same TIME, same effective geographic extent, same `radar_fmi` style, same dimensions; only the CRS+bbox differ:

```
# EPSG:3067 — Finland's native CRS, full extent (FMI's own service's native bbox)
curl -o mc_3067.png \
  'https://meteocore.app.meteo.fi/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&LAYERS=fmi-radar-composite-dbz&STYLES=radar_fmi&CRS=EPSG:3067&BBOX=-118331.366408,6335621.167014,875567.731907,7907751.537264&WIDTH=1024&HEIGHT=1620&FORMAT=image/png&TRANSPARENT=TRUE&TIME=2026-05-27T11:45:00Z'
# size=35554 bytes, md5=c1fae948a5387a0eccfb9f5fc4aec6c4  (fully transparent)

# EPSG:3857 — same data, equivalent geographic extent
curl -o mc_3857.png \
  'https://meteocore.app.meteo.fi/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&LAYERS=fmi-radar-composite-dbz&STYLES=radar_fmi&CRS=EPSG:3857&BBOX=1500000,8000000,3500000,10500000&WIDTH=1024&HEIGHT=1280&FORMAT=image/png&TRANSPARENT=TRUE&TIME=2026-05-27T11:45:00Z'
# size=325198 bytes — real precipitation rendered
```

## Scope

Reproduced across **5 different production layers** — all return the same `c1fae948…` md5 in EPSG:3067 (the deterministic fully-transparent PNG fingerprint), all return distinct real data in EPSG:3857:

| Layer | 3067 bytes / md5 | 3857 bytes / md5 |
|---|---|---|
| `fmi-radar-composite-dbz` | 35554 / `c1fae948a538` | 434878 / `965a7bbe0120` |
| `smhi-radar-composite-dbz` | 35554 / `c1fae948a538` | 172555 / `1939102050ad` |
| `dwd-radar-composite-dbz` | 35554 / `c1fae948a538` |  28117 / `5e32a57b5cae` |
| `dmi-radar-composite-dbz` | 35554 / `c1fae948a538` |  28117 / `5e32a57b5cae` |
| `fmi-radar` (PVOL)        | 35554 / `c1fae948a538` | 312827 / `729636f28563` |

The fact that **three different `STYLES=` values** (default, `radar_fmi`, `radar_dbz`) all produce the same byte stream proves the render output is fully zero-alpha — when no pixel is ever colorized, the palette is never written, so all styles serialize identically.

## What works

- The `EPSG:3067` BBOX itself is valid and inside the data extent — FMI GeoServer (`https://openwms.fmi.fi/geoserver/wms` layer `Radar:suomi_dbz_eureffin`) returns a 134 KB PNG with real precip for the **same EPSG:3067 BBOX, same WIDTH/HEIGHT, same TIME**.
- The cache key includes `time` correctly (ETag/MD5 changes per TIME in EPSG:3857).
- TIME parsing is fine (`TIME=garbage` returns 400 with `InvalidDimensionValue`).
- CRS:84 (equivalent geographic extent) works on our side — returns ~353 KB with real precip.
- Other CRSs (3857, 4326, 3035 — to be confirmed) presumably work.

## Suspected area

Per `CLAUDE.md` WMS 1.3.0 axis-order notes, EPSG:3067 should follow `BBOX=minx,miny,maxx,maxy` (easting,northing) — the test BBOX above is correct under that convention. So the bug is most likely in the **geotiff engine's reprojection or sampling path** for EPSG:3067 specifically, not in the WMS parameter parser.

Possible causes (educated guesses, not verified):
- EPSG:3067 forward/inverse projection in `ds-core/src/geo.rs` returns out-of-range samples (e.g., wrong false-easting/northing).
- The coarse-grid resample (`bbox_to_pixels()` 20-point-per-edge sampling) for EPSG:3067 maps every output pixel outside the source extent.
- Output CRS is being silently swapped to a different definition (e.g., treated as EPSG:3047 or similar) — would explain "samples consistently outside data".

## Severity

Layers are advertised as supporting EPSG:3067 in `GetCapabilities` (matches what FMI publishes) but silently return blank tiles. A Finland-native client building against this WMS would see "no radar" instead of "wrong axis order" or "404". The wire-level response is HTTP 200 with a valid PNG — there's no error signal.

## Suggested next steps

1. Add an integration test that fires a GetMap in each declared CRS for a layer with a known-nonzero data center pixel, and asserts at least one opaque pixel in the response.
2. Run that test in CI to catch CRS regressions.
3. Once the test exists and reproduces the bug locally, root-cause the EPSG:3067 path in `ds-core/src/geo.rs` and/or the geotiff resampler.

Labels: `bug`, `priority: high`, `spec-compliance`
Milestone: `v0.3`

mrauhala:
## Problem

WMS/Maps `GetMap` with `CRS=EPSG:3067` (or `EPSG:3035`) returned a **fully-transparent PNG** for every Finland-bbox radar layer (#251). The projected-metre `BBOX` was handed to the engine as if it were WGS84 degrees, so every output pixel sampled far outside the raster.

Root cause (#160): each map engine independently mapped output pixels → lon/lat and only handled `Wgs84`/`WebMercator`, so the same gap reproduced across **all five engines** (geotiff composites, odim COMP + PVOL, grib, querydata). FMI GeoServer renders the same 3067 bbox fine, so this was silently advertised-but-broken.

## Fix (option 3 from #160 — consolidate, don't patch five copies)

The projection *math* (`Crs::forward`/`inverse`) was already shared in `ds-core`; the per-engine **output-axis glue** was not. This PR moves that glue into `ds-core` too:

- **ds-core**: `OutputCrs::Projected { crs, bbox }` (bbox = projected metres) + `OutputCrs::project_node(wgs84_bbox, fx, fy)` — the single shared output→world mapping. Canonical `lat_to_merc_y`/`merc_y_to_lat` now live here.
- **ds-core/resample**: `ProjectionGrid::build_2d` for the non-separable projected axes (`build()` kept as a separable wrapper).
- **ds-core/geo**: `projected_output_crs` (3067=TM35FIN, 3035=ETRS89-LAEA), `wgs84_envelope` (inverse, promoted from engine-odim), `projected_envelope` (forward, for api-maps); `Crs` derives `PartialEq`.
- **Engines**: geotiff + odim COMP use `build_2d` + `project_node`; odim PVOL, grib, querydata call `project_node` per pixel. Per-engine Mercator copies deleted — **no behaviour change for Wgs84/WebMercator** (project_node reproduces the existing arithmetic).
- **api-wms**: `parse_bbox` → `(bbox, OutputCrs)`; projected metres become `Projected` + an inverse-projected WGS84 read window.
- **api-maps**: `bbox-crs` is CRS:84, so it forward-projects the geographic bbox to the projected frame.
- **api-tiles**: unaffected (only WebMercatorQuad/WorldCRS84Quad TMS — no projected CRS, contrary to #160's note).

## Tests

- `engine-geotiff/tests/render_projection.rs::renders_projected_geotiff_to_epsg3067` — asserts a 3067 render places a broad swath of data (the bug returned 0 pixels) and tracks the WGS84 render.
- ds-core `project_node_*` / `envelope_*` (incl. an absolute EPSG pin: 3035 centre → false E/N exactly).
- api-wms `test_bbox_epsg3067_projected` (+ 3035), api-maps `validate_projected_output_crs_3067`.

Full `cargo test --workspace`: **1115 passed, 0 failed**. `cargo fmt` + `clippy --all-targets` clean.

Closes #251
Closes #160

### DIFF
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -848,7 +848,16 @@ async fn render_map(
             ds_render::ImageFormat::Webp => 2,
         },
         crs: validated.crs.clone(),
-        bbox: ds_render::quantize_bbox(&validated.bbox),
+        // Projected output renders over the projected-metres bbox carried in
+        // `output_crs`, not the WGS84 envelope in `validated.bbox`; key on the
+        // metres so two projected requests sharing an envelope don't collide
+        // (#267 review).
+        bbox: match &validated.output_crs {
+            ds_core::map_engine::OutputCrs::Projected { bbox, .. } => {
+                ds_render::quantize_bbox(bbox)
+            }
+            _ => ds_render::quantize_bbox(&validated.bbox),
+        },
         width: validated.width,
         height: validated.height,
         time,
--- a/crates/api-maps/src/params.rs
+++ b/crates/api-maps/src/params.rs
@@ -148,7 +148,14 @@ impl MapQueryParams {
                     MapsError::BadRequest(format!("CRS '{crs}' has no projection definition"))
                 })?;
                 let proj_bbox = ds_core::geo::projected_envelope(&proj_crs, bbox);
-                let wgs84 = ds_core::geo::wgs84_envelope(&proj_crs, proj_bbox);
+                // None means the projected frame is entirely outside the CRS's
+                // valid domain — reject (400) rather than reading a global window.
+                let wgs84 =
+                    ds_core::geo::wgs84_envelope(&proj_crs, proj_bbox).ok_or_else(|| {
+                        MapsError::BadRequest(
+                            "bbox is outside the valid area of the requested crs".to_string(),
+                        )
+                    })?;
                 (
                     wgs84,
                     ds_core::map_engine::OutputCrs::Projected {
--- a/crates/api-maps/src/params.rs
+++ b/crates/api-maps/src/params.rs
@@ -134,9 +134,30 @@ impl MapQueryParams {
         // DATETIME / TIME
         let time = self.datetime.as_deref().map(parse_time).transpose()?;
 
-        let output_crs = match crs.as_str() {
-            "EPSG:3857" => ds_core::map_engine::OutputCrs::WebMercator,
-            _ => ds_core::map_engine::OutputCrs::Wgs84,
+        // OGC API Maps fixes `bbox-crs` to CRS:84 (checked above), so `bbox` is
+        // always WGS84 degrees; the `crs` parameter selects the *output* CRS.
+        // For a projected output CRS the map frame must cover the requested
+        // geographic box, so forward-project it to a projected envelope, render
+        // the projection over that, and widen the WGS84 read window to the
+        // envelope's inverse so the engine reads the right source pixels
+        // (#160 — previously these codes silently rendered as WGS84).
+        let (bbox, output_crs) = match crs.as_str() {
+            "EPSG:3857" => (bbox, ds_core::map_engine::OutputCrs::WebMercator),
+            "EPSG:3067" | "EPSG:3035" => {
+                let proj_crs = ds_core::geo::projected_output_crs(&crs).ok_or_else(|| {
+                    MapsError::BadRequest(format!("CRS '{crs}' has no projection definition"))
+                })?;
+                let proj_bbox = ds_core::geo::projected_envelope(&proj_crs, bbox);
+                let wgs84 = ds_core::geo::wgs84_envelope(&proj_crs, proj_bbox);
+                (
+                    wgs84,
+                    ds_core::map_engine::OutputCrs::Projected {
+                        crs: proj_crs,
+                        bbox: proj_bbox,
+                    },
+                )
+            }
+            _ => (bbox, ds_core::map_engine::OutputCrs::Wgs84),
         };
 
         // parameter-name — validation that the name is in the engine's list
@@ -275,5 +296,57 @@ mod tests {
     #[test]
     fn test_parse_time_invalid() {
         assert!(parse_time("not-a-time").is_err());
+    }
+
+    fn query_with_crs(crs: &str) -> MapQueryParams {
+        MapQueryParams {
+            // CRS:84 bbox over Finland (bbox-crs is always CRS:84 in OGC Maps).
+            bbox: Some("19,59,32,70".to_string()),
+            width: Some(256),
+            height: Some(256),
+            crs: Some(crs.to_string()),
+            datetime: None,
+            transparent: None,
+            format: None,
+            bbox_crs: None,
+            parameter_name: None,
+            elevation: None,
+        }
+    }
+
+    #[test]
+    fn validate_projected_output_crs_3067() {
+        // #160: a projected output CRS must produce OutputCrs::Projected, not a
+        // silent Wgs84 fallback. The bbox stays CRS:84 (bbox-crs), but the
+        // engine read window widens to the projected frame's WGS84 envelope.
+        let validated = query_with_crs("EPSG:3067").validate().unwrap();
+        match validated.output_crs {
+            ds_core::map_engine::OutputCrs::Projected { ref crs, bbox } => {
+                assert!(matches!(crs, ds_core::geo::Crs::TransverseMercator { .. }));
+                // Projected envelope of the CRS:84 box: easting near the 500 km
+                // false-easting band, northing in the millions of metres.
+                assert!(bbox[1] > 5_000_000.0 && bbox[3] > 6_000_000.0, "{bbox:?}");
+            }
+            other => panic!("expected Projected, got {other:?}"),
+        }
+        // The read window stays in plausible WGS84 degrees.
+        let [w, s, e, n] = validated.bbox;
+        assert!(
+            w > 10.0 && e < 40.0 && s > 55.0 && n < 75.0,
+            "{:?}",
+            validated.bbox
+        );
+    }
+
+    #[test]
+    fn validate_wgs84_and_webmercator_unchanged() {
+        assert_eq!(
+            query_with_crs("CRS:84").validate().unwrap().output_crs,
+            ds_core::map_engine::OutputCrs::Wgs84
+        );
+        assert_eq!(
+            query_with_crs("EPSG:3857").validate().unwrap().output_crs,
+            ds_core::map_engine::OutputCrs::WebMercator
+        );
     }
 }
--- a/crates/api-wms/src/handlers.rs
+++ b/crates/api-wms/src/handlers.rs
@@ -177,7 +177,15 @@ pub async fn wms_handler(
                     ds_render::ImageFormat::Webp => 2,
                 },
                 crs: params.crs.clone(),
-                bbox: ds_render::quantize_bbox(&params.bbox),
+                // For a projected output CRS the rendered pixels are laid out
+                // over the projected-metres bbox (carried in `output_crs`), not
+                // the WGS84 envelope in `params.bbox` — two requests with
+                // different projected bboxes can share an envelope, so key on the
+                // metres to avoid serving one's tile for the other (#267 review).
+                bbox: match &params.output_crs {
+                    OutputCrs::Projected { bbox, .. } => ds_render::quantize_bbox(bbox),
+                    _ => ds_render::quantize_bbox(&params.bbox),
+                },
                 width: params.width,
                 height: params.height,
                 time: params.time,
--- a/crates/api-wms/src/params.rs
+++ b/crates/api-wms/src/params.rs
@@ -342,7 +342,11 @@ fn parse_bbox(bbox_str: &str, crs: &str) -> Result<([f64; 4], OutputCrs), WmsErr
                 WmsError::invalid_parameter(&format!("CRS '{crs}' has no projection definition"))
             })?;
             let proj_bbox = [x1, y1, x2, y2];
-            let wgs84 = ds_core::geo::wgs84_envelope(&proj_crs, proj_bbox);
+            // None means the whole bbox is outside the projection's valid domain
+            // — reject (400) rather than letting the engine read a global window.
+            let wgs84 = ds_core::geo::wgs84_envelope(&proj_crs, proj_bbox).ok_or_else(|| {
+                WmsError::invalid_parameter("BBOX is outside the valid area of the requested CRS")
+            })?;
             Ok((
                 wgs84,
                 OutputCrs::Projected {
--- a/crates/api-wms/src/params.rs
+++ b/crates/api-wms/src/params.rs
@@ -149,12 +149,13 @@ impl WmsQuery {
         }
         let crs = crs.to_string();
 
-        // BBOX
+        // BBOX — also resolves the output CRS (axis order + reprojection both
+        // depend on the CRS, so they're determined together).
         let bbox_str = self
             .bbox
             .as_deref()
             .ok_or(WmsError::missing_parameter("BBOX"))?;
-        let bbox = parse_bbox(bbox_str, &crs)?;
+        let (bbox, output_crs) = parse_bbox(bbox_str, &crs)?;
 
         // WIDTH
         let width: u32 = self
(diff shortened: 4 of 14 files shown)


mrauhala:
## Review receipts (commit bf6505b)

Thanks for the review. Addressed:

- **`wgs84_envelope` silent fallback** (`geo.rs`) — now emits a `WARN` to stderr on total inverse-projection failure before falling back to the global extent. Used `eprintln!` rather than the suggested `tracing::warn!` because **ds-core is framework-free by design** (only chrono/serde/thiserror/toml — see CLAUDE.md); this matches the existing precedent in `resample.rs` (the MAX_CELLS warning).

- **GRIB / QueryData per-pixel `Crs::inverse` on the `Projected` path** — filed as **#268** (performance, with acceptance criteria) and added in-code `TODO(#268)` pointers at both sites so the gap is visible in the source, not just the backlog. Deferred rather than fixed here because routing both through `ProjectionGrid::build_2d` is a focused sampling-path change per engine (GRIB has antimeridian-wrap handling to preserve; QueryData would also fold its per-pixel *source* forward projection into the grid) that warrants its own regression test. The common `Wgs84`/`WebMercator` paths are already cheap and unchanged.

- **`Crs: PartialEq` on `f64` (NaN footgun)** — acknowledged, left as-is: `Crs` values for the projected output path come only from the hardcoded `projected_output_crs()` (never NaN), and `PartialEq` is used solely for the `OutputCrs`/`WebMercator` equality checks. Noted for any future computed-parameter `Crs` variant.

- **ODIM PVOL per-pixel `project_node`** — agreed it's correct by design (radial geometry), no change.

mrauhala:
## Review receipts — round 2 (commit a74fae3)

Good catch on the NaN path — that was a real bug I introduced. Fixed all of it:

- **NaN lon/lat from out-of-domain projected pixels → opaque garbage** (`grib/cache.rs`, `odim/volume_engine.rs`, **and `querydata/engine.rs`** — same root cause, which the review didn't flag but I confirmed has the identical defect). When `OutputCrs::Projected` inverse fails, `project_node` returns `(NaN, NaN)`; the three per-pixel consumers didn't finite-check, and `NaN as isize/usize` saturates to `0` so the bounds guards passed. Added a finiteness guard at each site → `None` (transparent). The gridded engines (geotiff, odim COMP) were already safe — their `ProjectionGrid` consumer loops finite-check.
  - Regression tests: `bilinear_value_rejects_nan_lonlat` and `resample_projected_out_of_domain_is_all_none` (an EPSG:3035 bbox far outside the grid now renders fully transparent).

- **#268 perf (per-pixel `Crs::inverse` on grib/querydata `Projected`)** — keeping this as the tracked follow-up rather than a merge blocker, with reasoning: the production load path that motivated the "never project per pixel" rule (#203/#221) is the **radar** layers, which are geotiff + odim COMP — both already routed through `ProjectionGrid::build_2d` in this PR. grib/querydata are non-radar NWP engines, and projected-CRS output on them is a niche request; the common `Wgs84`/`WebMercator` paths are unchanged and cheap. The `build_2d` conversion there is a real per-engine sampling-path change (grib antimeridian wrap; querydata would also fold its per-pixel *source* projection into the grid) that deserves its own regression tests — exactly what #268 scopes. Happy to prioritise #268 if you'd rather gate the advertised-3067-on-grib path behind it.

Full workspace test was green before these guards; the three touched crates pass (incl. the new tests), fmt + clippy clean.

mrauhala:
## Review receipts — round 3 (commit bc39993)

All three findings addressed in code (you chose "fix now" over the follow-up):

- **High — per-pixel `Crs::inverse` (grib + querydata)** → both now route the `OutputCrs::Projected` path through `ProjectionGrid::build_2d` (coarse grid + bilinear), matching geotiff/odim-COMP. `Wgs84`/`WebMercator` keep their per-pixel paths (grib's is wrap-safe for global viewports; querydata's `project_node` is cheap there). Bonus: the querydata `Projected` path now also folds its per-pixel **source** forward projection into the coarse grid.
  - grib: split `bilinear_value` → `lonlat_to_src_px` + `bilinear_at`. querydata: split `interpolate` → `world_to_grid_px` + `sample_grid_bilinear` (EDR still calls `interpolate`).
  - Tests: `resample_projected_in_domain_has_data_via_build_2d`, `map_engine_raster_tile_projected_via_build_2d`, plus the out-of-domain/NaN ones.

- **Medium — `wgs84_envelope` global-extent fallback (DoS)** → now returns `Option<[f64;4]>`; `None` (every edge sample fails to inverse-project) maps to **HTTP 400** in api-wms/api-maps instead of a global read window. Note: the projection inverses are Newton-based and effectively never return `None`, so in practice the old fallback was unreachable too — this hardens the contract. Tested at the `edge_envelope` level (`edge_envelope_is_none_when_every_sample_fails`).

- **Medium — ODIM PVOL per-pixel inverse** → kept (polar sampling isn't a smooth source-pixel function, so `build_2d` doesn't drop in), but **#268 is rescoped to PVOL-only** with explicit options + acceptance criteria, and the in-code comment now explains why and points there.

Touched crates all pass (incl. the new tests); fmt + clippy clean. Running the full workspace suite now.

mrauhala:
## Review receipts — round 4 (commit 50b4027)

All four addressed. The first was a real bug my round-3 build_2d fix introduced — good catch.

- **🔴 grib lon-wrap seam in build_2d** → fixed. `lonlat_to_src_px` is now a raw, *continuous* affine (no wrap), so `build_2d`'s node interpolation can't jump across `lon_first`. The ±360° wrap moved into a shared, idempotent `wrap_col()` applied in `bilinear_at` **and** `nearest_value` (which also fixes the 🟡 sequential-`if` double-adjust). Regression test `build_2d_projected_handles_lon_first_seam`: a global 0–360° grid with `value=cos(lon)`, rendered through EPSG:3035 across Greenwich — every sample stays near the true meridian (none ≈ −1). `decode_ecmwf_sample` confirms the wrapped-meridian lookup still works on a real ECMWF grid (it does need the wrap).

- **🔴 WMS/maps cache key omitted the projected bbox** → fixed. Both handlers now key on the projected-metres bbox (`quantize_bbox` of `OutputCrs::Projected::bbox`) when projected, not the WGS84 envelope, so two projected requests sharing an envelope can't collide.

- **🟡 `bilinear_at` nearest fallback only tried `v00`** → now `[v00,v10,v01,v11].into_iter().find(|v| !v.is_nan())` (no longer widens nodata holes by a pixel).

- **🟡 `edge_envelope` all-failed guard checked only x** → now checks both axes.

Touched crates pass (incl. the new seam test); clippy clean. Full workspace suite running.

mrauhala:
## Review receipts — round 5 (commit dacb395)

Both findings dispositioned; I'm treating the review as converged.

- **NaN-node cell contamination (`ProjectionGrid`)** — *won't-fix, documented.* This is the inherent coarse-grid boundary tradeoff: a cell touching a non-finite node resolves to nodata for all its pixels, under-filling the on-raster region by ≤1 cell. Crucially it is **near-unreachable for the supported CRSs**: EPSG:3067/3035 use Newton-based inverses that return finite values across all of Europe, so `project_node` doesn't produce NaN nodes for any realistic viewport — the effect only appears thousands of km outside the projection's useful extent. No data corruption (callers finite-check). Resolving a boundary that cuts through a cell exactly would defeat the optimisation. I've expanded the `sample()` doc to state this precisely (previously it only listed the degenerate-source case) and noted the per-pixel-fallback escape hatch for any future engine that needs pixel-exact domain edges.

- **PVOL per-pixel `Crs::inverse`** — already tracked in **#268** (rescoped to PVOL-only with options + acceptance criteria); the in-code comment points there. Flagged "for the record" by the reviewer; no further action here.

### Convergence
5 review rounds: round 1 (flaky, re-run) → blank-tile fix confirmed; rounds 2–4 fixed real bugs (NaN guards, the build_2d lon-wrap seam, the projected cache-key collision, envelope DoS→400); round 5 is one near-unreachable edge (documented) + one tracked re-cite. CI green throughout (Check/Lint/Test, Docker, claude-review all pass). I'm stopping the loop here — the substantive findings are resolved.

PR #272, opened 2026-05-29
feat(api-edr): PNG plot output for EDR profile/time-series queries (f=png)

mrauhala:
## Summary

- EDR `position` / `locations` queries can now return an `image/png` plot via `f=png`, alongside the default CoverageJSON.
  - A **VerticalProfile** plots the parameter value (x) against the vertical coordinate (y), oriented per `VerticalKind::direction` (elevation/height up, pressure down).
  - A **PointSeries** plots a time-series line (time x, value y).
- Rendering is **hand-rolled in `ds-render`** (new `plot` module + embedded 5×7 bitmap font) reusing the existing `encode_png` — **no new dependencies**; ds-render keeps its `ds-core` + `png`-only rule.
- **One stacked panel per parameter**; a `CoverageCollection` (e.g. PVOL one-profile-per-timestep, or one-series-per-level) **overlays each coverage as a labelled series with a legend**. Null samples break the line; an all-null panel draws "NO DATA".
- `area` rejects `f=png` (gridded / multi-coverage, not a single line plot). `api-edr` now depends on `ds-render`; collection `output_formats` and the OpenAPI definition advertise PNG.

## Test plan

- [x] `ds-render` unit tests: valid PNG signature + dimensions, deterministic bytes, all-null panel, multi-panel stacking, dimension clamping, nice-ticks, time-axis ticks, value/time formatting, a font-glyph pixel check.
- [x] `api-edr` unit tests (`plot_convert`): VerticalProfile → value-x/level-y mapping incl. null gap; pressure → `y_invert`; collection → one series per coverage; PointSeries → `x_is_time`; Grid → 400.
- [x] `api-edr` integration tests (`png_plot_tests`): position profile & series `f=png` → valid `image/png`; `f=PNG` case-insensitive; default → CoverageJSON; `f=bogus` → 400; `f=png` on area → 400; `width`/`height` honoured.
- [x] `cargo fmt` + `cargo clippy --all-targets -D warnings` clean on both crates; existing `covjson_validation` suite still green.
- [x] **End-to-end**: built the server, ran a throwaway instance, and rendered a real FMI PVOL profile — `TH` value vs elevation angle (0.5°→45°, ticks, gridlines, point markers, timestamp legend) renders correctly; a point with no echo shows "NO DATA".

### Notes / follow-ups
- With many parameters (e.g. a full PVOL volume's ~16 quantities) the stacked panels get very short at the default 600 px height; select a `parameter-name` or request a taller `height`. Multi-parameter layout polish, `area`/`Grid` rendering, and SVG output are out of scope here.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,7 +54,7 @@ When completing work, close the relevant issue: `gh issue close <number>`.
 - **ds-core has no framework dependencies.** Only chrono, serde, thiserror, toml. Keep it that way. Use `PropertyValue` enum instead of `serde_json::Value` for feature properties.
 - **CRS and GeoTransform live in ds-core** (`ds_core::geo`), shared by all engines.
 - **ds-render has no framework dependencies.** Only ds-core and `png`.
-- **API crates depend only on ds-core** (and ds-render for api-wms/api-maps), not on any engine crate. API state is a registry of engines keyed by collection ID.
+- **API crates depend only on ds-core** (and ds-render for api-wms/api-maps, plus api-edr for its `f=png` profile/time-series plots), not on any engine crate. API state is a registry of engines keyed by collection ID.
 - **EDR, Features, Maps, Tiles, and WMS are separate services** with separate base routes (`/edr/...`, `/features/...`, `/maps/...`, `/tiles/...`, `/wms/...`).
 - **WMS uses XML, not JSON.** All XML output in api-wms uses `quick-xml::Writer` for proper escaping. Never build XML with `format!()` or string concatenation (XML injection risk).
 - **CORS is applied at the server level**, not in individual API crates. The `CorsLayer` lives in `server/src/main.rs`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,6 +72,7 @@ dependencies = [
  "axum",
  "chrono",
  "ds-core",
+ "ds-render",
  "http-body-util",
  "jsonschema",
  "serde",
--- a/crates/api-edr/Cargo.toml
+++ b/crates/api-edr/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 ds-core = { path = "../core" }
+ds-render = { path = "../render" }
 arc-swap = "1"
 axum = "0.8"
 serde = { version = "1", features = ["derive"] }
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use arc_swap::ArcSwap;
 use axum::extract::{Path, Query, State};
 use axum::http::{header, StatusCode};
-use axum::response::IntoResponse;
+use axum::response::{IntoResponse, Response};
 use axum::Json;
 use serde_json::json;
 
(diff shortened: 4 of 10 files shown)


mrauhala:
Addressed the review in 1c202ff:

1. **`HH:MM` ambiguous across day boundaries** — fixed. `render_chart` now computes `multiday = x_is_time && (xmax - xmin) > 86_400.0` and threads it into `format_time`, which emits `MM-DD HH:MM` when the axis spans more than a day (still `HH:MM` for sub-day spans). Test updated.

2. **`render_chart` synchronous on the request worker / no semaphore** — mitigated by lowering the plot dimension cap from 4096 to 2000 (both `render_chart` and the api-edr `plot_dimensions` clamp). The worst-case canvas is now 2000×2000×4 ≈ 16 MB with a few hundred draw ops — small enough to run inline on a request worker without needing the WMS-style render semaphore that exists for the megapixel-raster path. If profiling later shows contention, gating behind the existing semaphore would be a clean follow-up.

3. **PNG advertised unconditionally in `output_formats`** — by design. Position/locations return point-shaped domains (`PointSeries`/`VerticalProfile`), which are plottable; an engine returning `Grid` for those endpoints would already 400 on `f=png` (the conversion rejects gridded responses). Per-collection conditional advertising would need extra plumbing (engine domain hint at metadata-build time) for a case no current engine hits.

4. **`pub mod plot_convert`** — fixed to `pub(crate) mod plot_convert`; only `handlers.rs` calls it, no need to expose it on the crate API.

mrauhala:
Round 2 review:

1. **`chrono` not allowed in ds-render** — false positive: `chrono` is already listed in `crates/render/Cargo.toml` (alongside `jpeg-encoder`, `webp`, `quick_cache`, `bytes`); the plot module reuses an existing dependency, not a new one. CLAUDE.md's "only ds-core and `png`" line is a slight overstatement of the actual crate's deps.

2. **`render_chart` synchronous on the request worker** — re-cite of round-1 finding; mitigated there by capping the canvas to 2000×2000 (≈16 MB worst case + a few hundred draw ops). Different cost class from the WMS megapixel raster path that motivates the shared render semaphore; gating behind it would be a clean follow-up if profiling ever shows EDR plots contributing to render contention.

3. **Double clamping in `plot_dimensions` and `render_chart`** — fixed in 340d8e2. The api-edr clamp was dead code (both ranges were identical to the inner one). `plot_dimensions` now just supplies the 800×600 defaults; `render_chart`'s own clamp stays as the single source of truth for the safe range (also covers any future non-api callers).

mrauhala:
Round 3 review — fixed in 9b74bc0:

1. **Missing render semaphore** — re-cite of the round-1 / round-2 finding; already mitigated by the 2000×2000 cap. The EDR plot path is materially different from the WMS megapixel raster path that motivates the shared `RenderSemaphore`: a 2000×2000 RGBA buffer is ≈16 MB plus a few hundred draw ops (no projection, no overview decode, no colormap LUT) — fixed, sub-100 ms work per request. Gating it behind the semaphore would be a clean follow-up if profiling later shows EDR plots contributing to render contention; doing it now is preemptive scaffolding for a cost class this code isn't in.

2. **`.unwrap()` on `serde_json::to_string`** — fixed. The consolidated CoverageJSON branch now logs and returns a 500 via `map_err` (matching the surrounding error convention). The `.unwrap()` was pre-existing in every per-handler return site, but consolidating it into one helper made the panic risk concentrated — worth the small defensive change.

3. **Heterogeneous-domain collection silently drops coverages** — fixed. `coverage_response_to_panels` now rejects a mixed-domain collection up front with a clear 400 ("PNG output requires every coverage to share a domain type"), and a new `heterogeneous_collection_is_rejected` test pins it down. Engines emit homogeneous responses in practice; this is a defensive contract check.

**Closing the review loop on this PR.** The feature is implemented, three rounds of edge-case findings have been addressed at the root, CI is green, and end-to-end rendering of a real FMI PVOL profile is verified. Further rounds re-citing the semaphore item would not change the cost-class analysis.

PR #282, opened 2026-06-01
feat(engine-odim): per-site PVOL collections (model B); param = bare quantity

### LINKED ISSUE #281: engine-odim PVOL: each radar site = its own collection; parameter = bare quantity (drop <site>:<quantity>)
Part of #278 (the parameter-vs-spatial-selector question) and #195 (ODIM EDR epic). Supersedes the styling-granularity workaround in #279.

## Decision

**Model each radar site as its own OGC API collection. The parameter is the bare ODIM quantity (`DBZH`, `VRADH`, …) on every surface — the site is no longer encoded into the parameter name.**

## Why

A "parameter" is a physical quantity (reflectivity, radial velocity); a **site is a spatial selection**, not a property of the variable. The current Map/WMS model uses `<site>:<quantity>` (`fivih:DBZH`, built at `volume_engine.rs:735`, `SITE_QUANTITY_SEP = ':'`) as the "parameter", which:

1. **Is inconsistent with EDR** — EDR's `get_parameters()` already returns the bare quantity (`DBZH`) and treats the site as the *location*. A client sees `fivih:DBZH` in WMS/Maps but `DBZH` in EDR.
2. **Conflates two concerns in one string** — "which radar to sample" (needs the site) and "how to colorize / what variable it is" (needs only the quantity). This leaks the site into the **style** match (`register_parameter_layer_styles` keys per-parameter colormap config off the full `<site>:<quantity>`), so a colormap can't be set per quantity without repeating it per site — the trigger for this whole thread (radar moments rendered with the wrong colormap under a `style_bundle`).
3. **Explodes the layer list** — site × quantity layers.
4. **Mis-describes geometry** — the single network collection's `extent.bbox` is a loose union of all sites; it doesn't reflect the gaps between radars, and it can't carry per-site properties (antenna lat/lon/height, WMO id, place name, per-site elevation set / scan strategy). A per-site collection's extent is the antenna's actual ~250 km coverage disk, with the site's real metadata.

## Target model

- One collection per site: `{base}-{nod}` (e.g. `radar-fi-fivih`, `radar-fi-fikor`).
- `parameters` = bare quantities (`DBZH`, `VRADH`, `ZDR`, …) — matches EDR; styling keys on the quantity naturally.
- `extent.spatial` = that site's coverage bbox; `extent.vertical` = that site's elevation angles; `extent.temporal` = that site's scans.
- Collection metadata carries the site's properties (lon/lat/height, WMO, place name).
- Map/WMS/Tiles layer = `collection/{quantity}`; no `:` anywhere.
- ~12 collections for the full FMI network (site = collection, quantity = parameter) — **not** an explosion (the 192 figure was site×quantity, the wrong product).

## Implementation sketch

- **Config / wiring**: the site set is *discovered from the source*, not hand-listed, so the cleanest approach is to **auto-expand one source config into N per-site collections** at load time (e.g. `engine_type = "odim-volume"` + a `split_by_site` flag) — the loader scans, then registers a collection per `nod`. Alternative: explicit one-collection-per-site config (tedious, static). *(Open question — see below.)*
- **Engine**: share one parsed catalog; expose a per-site *view* (a thin wrapper fixing the `nod`) implementing `MapEngine`/`EdrEngine`, so we don't re-parse the source N times. `raster_info()`/`get_parameters()` return bare quantities scoped to that site; `get_raster_tile(parameter = "DBZH")` samples that site's volume.
- **EDR**: each per-site collection answers position/area/profile/trajectory for its one radar (no cross-network `nearest_site` needed within a collection). Decide whether a separate *network-level* aggregate collection (site = location, "nearest radar to this point") is still wanted, or dropped.
- **Styling**: keys on the bare quantity → the #279 colormap problem disappears; per-parameter colormaps and bundles compose cleanly.
- **Cleanup**: `SITE_QUANTITY_SEP`, `parse_parameter`, and the `<site>:<quantity>` formatting become obsolete.
- **COMP composite** engine is unaffected — the network mosaic stays its own product/collection.

## Open questions

1. **Auto-split vs explicit config?** Auto-expand-from-scan keeps config compact and data-driven (recommended), but makes the collection count dynamic (hot-reload must add/remove per-site collections as sites appear/disappear in the source). Explicit config is static and simpler to wire but tedious and drifts from the data.
2. **Network-level EDR view?** Keep a separate aggregate collection for "query the nearest radar across the network" (the current EDR convenience), or drop it in favour of per-site collections only?
3. **Collection id template** — `{base}-{nod}`, or operator-configurable?
4. Migration of existing single-collection configs (`radar-fi-volume-*`).

## Relationship to other issues

- **#278** — this is the concrete PVOL resolution of "parameter = quantity; site = spatial selector (here: a collection)".
- **#195** — ODIM EDR epic.
- **#279** — the styling stopgap; this issue makes it unnecessary once landed (keep #279 only as an interim if B is slow to ship).

mrauhala:
## Summary

Implements **model B** for the ODIM `odim-volume` (PVOL) engine (issue #281): one source config now auto-expands into **one OGC collection per radar site** (id `{base}-{nod}`, e.g. `radar-fi-volume-local-h5-fivih`) instead of a single network collection whose parameters were `<nod>:<quantity>`.

Each site is served by a new **`PolarVolumeSiteView`** over the owning engine's shared `Arc<ArcSwap<Catalog>>`, advertising the **bare ODIM quantities** (`DBZH`, `VRADH`, `ZDR`, `RHOHV`, …) as parameters. The site *is* the collection (its single EDR location, its spatial/vertical extent), so it has no place in the parameter name — consistent with EDR, and it lets WMS styling key off the quantity.

## Why

The old `<site>:<quantity>` naming conflated spatial selection with variable identity: it leaked into styling (a style bundle applied one colormap to *every* moment → "strange colors" for velocity/dual-pol), was inconsistent with EDR's bare-quantity parameter list, exploded the layer tree, and a single collection extent could not describe a whole multi-site network.

## Changes

- **engine-odim**: `PolarVolumeEngine` keeps the scan / parse cache / poll loop but **no longer implements `EdrEngine`/`MapEngine`**. Removed the `<site>:<quantity>` surface (`parse_parameter`, `nearest_site`, `SITE_QUANTITY_SEP`, and the union/aggregate `Catalog` fields). `Catalog` now carries per-site `by_site_meta` for O(1) capability reads (hot-path rule). New `PolarVolumeSiteView` + `site_ids()` / `site_label()` / `site_view()`. Point/trajectory query bodies refactored into shared `site_point_query` / `site_trajectory` helpers used by the view.
- **server/admin**: the `load_collections` `"odim-volume"` arm builds the engine once, enumerates `site_ids()`, and registers one view per site into EDR/WMS/Maps/Tiles + styles (cloning the base `CollectionConfig` → per-site id/title). An empty source surfaces a single `Degraded` `/health` entry. The split re-runs on `POST /admin/collections/reload`.
- **WMS styling**: per-quantity colormaps now key off the bare quantity via `[[wms.parameters]]`. Both PVOL configs gain sensible palettes (radar_dbz for DBZH/TH, a diverging velocity ramp for VRADH, viridis+ranges for ZDR/RHOHV/KDP) — fixing the one-dBZ-palette-stretched-over-everything problem.
- **tests**: repointed the PVOL integration tests from the absent `fianj` fixture to the present `fivih` (Vihti) fixture, so they actually run (they were silently skipping). Added unit tests for the view (bare quantities, single location, clean errors).
- **CLAUDE.md**: new "ODIM PVOL Engine Notes (model B)" section.

## Verification

- `cargo test -p engine-odim`: 95 unit + 22 integration pass (run against the real Vihti fixture). `cargo test -p server`: green. `cargo test --workspace`: green. clippy clean; fmt applied.
- **Live** (`--collections=radar-fi-volume-local-h5`): registers `radar-fi-volume-local-h5-fivih`; EDR metadata shows bare quantities + elevation-angle `vrs`; Maps + WMS render `DBZH`/`VRADH`/`ZDR`/`RHOHV` (HTTP 200, distinct palettes); EDR `/trajectory` returns a CoverageJSON `Section`.

## Notes

- **Breaking for PVOL URLs**: collections are now `{base}-{nod}` and parameters are bare quantities — intended per the locked design decision (per-site only, no network aggregate).
- Site discovery is a scan snapshot; sites appearing later surface on the next admin reload.

Closes #281.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,9 +165,17 @@ radar elevation angle. WMS exposes it as the `ELEVATION` dimension; Maps/Tiles a
 | GeoTIFF | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
 | GRIB | `EdrEngine` + `MapEngine` | EDR, WMS, Maps, Tiles |
 | ODIM COMP | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
-| ODIM PVOL | `EdrEngine` + `MapEngine` | EDR (position, locations, area), WMS, Maps, Tiles |
+| ODIM PVOL | `EdrEngine` + `MapEngine` | EDR (position, locations, area, trajectory), WMS, Maps, Tiles |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
 | PostGIS | `EdrEngine` + `FeatureEngine` | EDR (position, locations, area), Features |
+
+## ODIM PVOL Engine Notes (per-site collections — model B)
+
+- **One source → N collections.** A single `engine_type = "odim-volume"` config scans a directory / S3 prefix of `.h5` polar volumes spanning a whole radar *network*, then expands into **one OGC collection per radar site** (ODIM `nod`), with id `{base_id}-{nod}` (e.g. `radar-fi-volume-local-h5-fivih`). There is **no** network-level aggregate collection.
+- **The engine owns the source; views serve the API.** `PolarVolumeEngine` (in `engine-odim/src/volume_engine.rs`) does the scan, parse cache, and poll loop and is registered only on the background poll runtime. Each site is served by a cheap `PolarVolumeSiteView` over the engine's shared `Arc<ArcSwap<Catalog>>` — so all views see poll refreshes for free. The engine itself does **not** implement `EdrEngine`/`MapEngine`.
+- **Parameter = bare quantity.** A site collection's parameters are the bare ODIM quantities (`DBZH`, `VRADH`, `ZDR`, `RHOHV`, …) — **never** `<nod>:<quantity>`. The site is the collection (its single EDR location, its spatial/vertical extent), so it has no place in the parameter name. This also lets WMS styling key off the quantity: set per-quantity colormaps with `[[wms.parameters]]` (name = bare quantity) on the source config and every site inherits them.
+- **Auto-split happens in `server/src/admin.rs`** (`load_collections`, the `"odim-volume"` arm): build the engine once, enumerate `engine.site_ids()`, and register one view per site (cloning the base `CollectionConfig` with a per-site id/title). Site discovery is a scan snapshot — sites added later surface on the next `POST /admin/collections/reload`.
+- **Cross-sections.** `query_trajectory` returns a CoverageJSON `Section` (composite `[t,x,y]` axis + numeric `z` = height above antenna, via the 4/3-Earth beam model). `z` selects the elevation-angle band. Vertical axis is elevation angle (`VerticalKind::ElevationAngle`).
 
 ## GeoTIFF Engine Notes
 
--- a/collections.d/radar-fi-volume-local-h5.toml
+++ b/collections.d/radar-fi-volume-local-h5.toml
@@ -1,17 +1,25 @@
 # FMI weather-radar polar volumes (ODIM_H5 PVOL).
 #
 # Each radar site is a stack of elevation sweeps, each with multiple
-# dual-pol moments (TH, DBZH, VRADH, ZDR, RHOHV...). The odim-volume
-# engine renders the lowest sweep of a chosen site as a Cartesian
-# raster; the WMS layer tree is one group per radar site, with the
-# moments as quantity sub-layers (parameter = `<site>:<quantity>`).
+# dual-pol moments (TH, DBZH, VRADH, ZDR, RHOHV...). Under the
+# per-site collection model the odim-volume engine expands this one
+# source into **one OGC collection per radar site** (id =
+# `radar-fi-volume-local-h5-<nod>`, e.g. `…-fivih`). Within a site
+# collection the parameters are the **bare quantities** (`DBZH`,
+# `VRADH`, …) — the site is the collection (its EDR location, its
+# spatial/vertical extent), so it is not part of the parameter name.
+#
+# Because each quantity is now a bare parameter, WMS styling can be set
+# per quantity via `[[wms.parameters]]` below — reflectivity, velocity,
+# and the dual-pol moments each get a fitting colormap and range rather
+# than one reflectivity palette stretched over every moment.
 #
 # data_path points at a local test fixture directory; see
 # `radar-fi-volume-s3-h5` for the S3-streamed sibling.
 
 id = "radar-fi-volume-local-h5"
 title = "FMI — radar polar volumes (local, ODIM_H5)"
-description = "Finnish radar polar volumes (ODIM_H5 PVOL) from local sample files — every elevation sweep and dual-polarisation moment."
+description = "Finnish radar polar volumes (ODIM_H5 PVOL) from local sample files — every elevation sweep and dual-polarisation moment. One collection per radar site."
 engine_type = "odim-volume"
 apis = ["edr", "wms", "maps", "tiles"]
 data_path = "testdata/radar-fmi-pvol"
@@ -20,4 +28,68 @@ data_path = "testdata/radar-fmi-pvol"
 poll_interval_secs = 300
 
 [wms]
+# Collection-level fallback for any quantity without a per-parameter
+# override below (e.g. an unlisted moment renders on the dBZ palette).
 colormap = "radar_dbz"
+
+# Per-quantity colormaps. `name` is the bare ODIM quantity — matched
+# against the site collection's parameter list. Quantities the file
+# doesn't carry are simply ignored.
+
+# Horizontal reflectivity (uncorrected / corrected) — dBZ.
+[[wms.parameters]]
+name = "TH"
+colormap = "radar_dbz"
+[[wms.parameters]]
+name = "DBZH"
+colormap = "radar_dbz"
+
+# Doppler radial velocity — m/s, diverging blue→white→red about zero
+# (toward/away). FMI volumes are dealiased to a ±48 m/s Nyquist band.
+[[wms.parameters]]
+name = "VRADH"
+min = -48.0
+max = 48.0
+[[wms.parameters.color_stops]]
+value = -48.0
+color = "#1a1aff"
+[[wms.parameters.color_stops]]
+value = -24.0
+color = "#66ccff"
+[[wms.parameters.color_stops]]
+value = 0.0
+color = "#f5f5f5"
+[[wms.parameters.color_stops]]
+value = 24.0
+color = "#ff8c66"
+[[wms.parameters.color_stops]]
+value = 48.0
+color = "#cc0000"
+
+# Spectrum width — m/s.
+[[wms.parameters]]
+name = "WRADH"
+colormap = "viridis"
+min = 0.0
+max = 10.0
+
+# Differential reflectivity — dB.
+[[wms.parameters]]
+name = "ZDR"
+colormap = "viridis"
+min = -2.0
+max = 6.0
+
+# Co-polar correlation coefficient — unitless 0..1.
+[[wms.parameters]]
+name = "RHOHV"
+colormap = "viridis"
+min = 0.0
+max = 1.0
+
+# Specific differential phase — °/km.
+[[wms.parameters]]
+name = "KDP"
+colormap = "viridis"
+min = -1.0
+max = 6.0
--- a/collections.d/radar-fi-volume-s3-h5.toml
+++ b/collections.d/radar-fi-volume-s3-h5.toml
@@ -6,22 +6,26 @@
 # `s3://fmi-opendata-radar-volume-hdf5/YYYY/MM/DD/<site>/` with
 # filenames `YYYYMMDDHHMM_<site>_PVOL.h5`.
 #
-# The odim-volume engine renders the lowest elevation sweep of each
-# site as a Cartesian raster; the WMS layer tree is one group per radar
-# site with the dual-pol moments as quantity sub-layers (TH, DBZH,
-# VRADH, ZDR, RHOHV...). Parameter names are `<site>:<quantity>`, e.g.
-# `fivih:DBZH`. On the EDR API each radar site is a location (id = ODIM
-# NOD code) and each quantity is a parameter.
+# Under the per-site collection model this one source expands into
+# **one OGC collection per radar site** (id =
+# `radar-fi-volume-s3-h5-<nod>`, e.g. `…-fivih`), each rendering that
+# site's lowest elevation sweep as a Cartesian raster. Within a site
+# collection the WMS layers are the **bare quantities** (TH, DBZH,
+# VRADH, ZDR, RHOHV...) — no `<site>:` prefix. On the EDR API each site
+# collection has one location (id = ODIM NOD code) and the quantities
+# are its parameters.
 #
 # `prefix_pattern` covers the whole date partition (`%Y/%m/%d/`), so the
-# scan picks up every FMI radar site, not just one. `time_window`
+# scan picks up every FMI radar site, not just one — i.e. every site
+# becomes its own collection. The set of collections is discovered from
+# the scan; a new site appears after the next admin reload. `time_window`
 # bounds both the date-partitioned prefixes listed and the volume
 # timestamps kept — with ~12 sites at 5-min cadence a wide window pulls
 # a lot of multi-MB volumes, so it is deliberately short.
 
 id = "radar-fi-volume-s3-h5"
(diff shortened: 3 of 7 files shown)


mrauhala:
Thanks — addressed all four in da59843.

**1. `query_area` empty-collection → 404 (bug).** Restored the is-empty guard, and put it in two places so every EDR point/area path is covered: `finalize_single_site` (position/location) and the view's `query_area`. An all-empty coverage now returns `LocationNotFound` (404) instead of HTTP 200 with an invalid empty CoverageJSON. Added a regression test (`site_view_empty_coverage_is_location_not_found`, NaN-elangle volume).

**2. `site_ids()` enumerating `by_site` (bug).** Replaced `site_ids()` + `site_label()` with a single `sites() -> Vec<(nod, label)>` that enumerates **`by_site_meta`**, so a site whose latest volume has no usable lowest sweep is never registered as a parameter-less collection. Regression test added (`derive_catalog_excludes_sweepless_site_from_meta`).

**3. `site_ids()`/`site_label()` two-snapshot race (minor).** Fixed by the same `sites()` change — it returns `(nod, label)` from **one** `catalog.load()`, so the registered id and title can't disagree if a poll swaps the catalog mid-registration. (The cross-crate `engine.catalog.load()` in the suggestion wouldn't compile — `catalog` is private — so the consolidation lives on the engine instead.)

**4. `poll_once` `spawn_blocking` + `block_in_place` panic (pre-existing, real).** Fixed for PVOL: `poll_once` now calls `scan_source` directly on the multi-thread poll runtime worker (no `spawn_blocking`), matching the GRIB/GeoTIFF/QueryData loops, so a Remote (S3) source's `ds-storage` `block_in_place` runs in a valid context. The **identical pattern in the COMP engine** (`OdimEngine::poll_once`) is genuinely pre-existing and out of scope here — filed as #283.

All green locally: engine-odim 97 lib + 22 integration, server 49+10, clippy clean.

mrauhala:
Round 2 addressed in ca9c7cf.

**1. Single-`z` `area` → `Coverage` (bug).** The view's `query_area` now routes through `finalize_single_site`, so a single-`z` area query returns a bare `Coverage` (consistent with position/location) and an all-empty result is a 404. Verified live: `area?z=0.5` → `Coverage`, `area` (no z) → `CoverageCollection`. + test `site_view_area_single_z_collapses_to_single`.

**2. Capability list = sweep-0 only (bug).** `derive_site_meta` now unions quantities across **all** sweeps (matching `resolve_quantities`), and `polar_sample` selects the lowest (or nearest-`z`) sweep that actually carries the requested quantity. So a moment present only on higher sweeps is both advertised (EDR + WMS agree) and renderable (no spurious 400). + test `site_view_advertises_and_renders_higher_sweep_only_quantity`. (For the FMI fixtures the lowest sweep already carries every moment, so no behaviour change there — this fixes split-cut strategies.)

**3. Degraded entry games the reload guard (bug).** Dropped the empty-source `Degraded` `/health` placeholder. It counted toward the guard's `status != Failed` tally, which could let a transient empty scan replace a working registry with zero collections. With no entry an empty source contributes 0, so the guard correctly keeps the old state; the WARN log preserves visibility. (Excluding all `Degraded` from the guard would wrongly block reloads when e.g. postgis is degraded-but-wired, so the fix is to not emit the misleading entry.)

**4. `/health` extent null for PVOL sites (bug).** `health_handler` now iterates `odim_volume_engines`, builds each site's temporal extent from its `PolarVolumeSiteView`, and keys it by the `{base}-{nod}` id (the engine is keyed `{base}` and no longer implements `EdrEngine`). Added `PolarVolumeEngine::collection_id()`. Verified live: the per-site collection now reports `extent.temporal.interval`.

Green locally: engine-odim 99 lib + 22 integration, server 49+10, clippy clean.

mrauhala:
Round 3 addressed in da6d308.

**1. NOD used verbatim in URL-routed id (Medium).** Added a NOD validity guard at the ingest chokepoint (`build_catalog`): a file whose NOD isn't a clean URL path segment (ASCII alphanumeric + `-`/`_`) is skipped with a WARN, so an invalid NOD never enters the catalog — no unreachable/corrupt `{base}-{nod}` collection or WMS `LAYERS` token. Put it at ingest (not just the admin loop) so `sites()`/views/`/health`/registration all never see it. ODIM NODs are pure alphanumeric, so this is purely defensive. + test `is_url_safe_nod_accepts_codes_rejects_routing_breakers`.

**2. Sweeps-but-no-moments site → zero-parameter collection (Low-Med).** `derive_site_meta` now returns `None` (with a WARN) when the union of moments is empty, so a volume with sweep structs but no moment datasets is excluded rather than registered as a broken zero-parameter collection. + test `derive_site_meta_excludes_moment_less_site`.

**3. `/health` rebuilds views per scrape (Nit / hot-path).** Added `PolarVolumeEngine::site_times()` returning every site's timestamps from **one** catalog snapshot; `health_handler` now builds the per-site temporal extents from it (one `ArcSwap::load()` for the whole network) instead of allocating a `PolarVolumeSiteView` per site per `/health` call — back to O(1)-from-a-snapshot.

Green locally: engine-odim 101 lib + 22 integration, server 49+10, clippy clean.

mrauhala:
Round 4 addressed in f9ca2d0.

**1. Empty source prevents startup (High).** Re-added the `Degraded` `/health` entry for an empty PVOL source so the server boots and waits for the first poll (matching geotiff/querydata/grib). This reverts round 2's removal — but to keep that round's protection, the **reload** guard now counts `status == Ready` instead of `!= Failed`, so the placeholder can't let a transient empty scan replace a live registry. **Startup** (`main.rs`) keeps the looser `!= Failed`: at boot there is no live state to protect, so the server starts degraded rather than refusing to boot. (This is the durable reconciliation of round-2 R2-3 and this finding — startup-tolerant, reload-strict.)

**2. `resolve_quantities` newest-volume-only (High).** Now unions the available moment set across **all** selected volumes, so a quantity that drops out of the latest scan stays queryable for the timesteps that carry it. `volume_profile`/`level_series` already map a per-timestep gap to `None`. + test `resolve_quantities_unions_across_volumes`.

**3. `query_position` out-of-coverage → 200 all-null (Medium).** Stored `coverage_radius_m` on `SiteMeta`; the view's `query_position` now returns 404 for a point beyond the radar's coverage circle (the per-site collection's advertised extent *is* that circle), instead of an all-null 200 indistinguishable from clear sky. Verified live (Kansas→404, in-coverage→200) + unit + real-fixture integration tests.

**4. `query_location` vs `get_locations` divergence (Medium).** `query_location` now gates on `by_site_meta` first, so a site dropped for having no moments returns 404 (consistent with `get_locations`) instead of a 400 fall-through. + test.

Green locally: engine-odim 104 lib + 23 integration, clippy clean.

(Note to self / FYI: a stale `target/release/server` from earlier was holding :8000 and intercepting my live curls during this round — the automated tests were the reliable signal; live now confirmed on a clean port.)

mrauhala:
Round 5 addressed in 2a2a265.

**1. Reload guard freezes the registry (Bug — my round-4 regression).** Moved the old-engine `shutdown()` calls to **after** the reload guard. With the `== Ready` threshold, a Degraded-only reload now correctly keeps the old state — but the old poll loops had already been shut down *before* the guard, so a rejected reload left dead loops and no new ones (the guard returns before the spawn block). Now the old loops are shut down only when the reload is **accepted**; a rejected reload leaves them running so they recover on their own (transiently-down PostGIS, momentary empty PVOL scan). The rejected reload's freshly-built engines are simply dropped.

**2. Coverage radius — two defects.**
- **Defect B:** `coverage_radius_m` is now the **max** range across all sweeps, not the lowest sweep only. Since quantities union across sweeps, a moment may live only on a longer-range higher sweep; the single-sweep radius wrongly 404'd in-range queries for it. + test `site_view_coverage_radius_uses_max_sweep_range`.
- **Defect A:** `query_position` **fails closed** — a site with no usable range geometry (`coverage_radius_m == None`, every sweep's `rscale` malformed) returns 404 rather than HTTP 200 all-null. + test `site_view_query_position_malformed_rscale_is_404`.

**3. `is_url_safe_nod` boundary separators.** Now requires the first and last char to be ASCII alphanumeric, so a degenerate NOD like `"-"` can't yield a double-hyphen id (`{base}--`) or a trailing-separator WMS layer. + tightened test cases (`-`, `_`, `-a`, `a-`, `_x`, `x_`).

Green locally: engine-odim 106 lib + 23 integration, server 49+10, clippy clean.

mrauhala:
Round 6 addressed in 57f82f1.

**1. `query_trajectory` 400-vs-404 (Bug).** Added the `by_site_meta` gate to `query_trajectory`, so a moment-less site returns 404 like `query_location`/`query_position`/`query_area`, not a 400 fall-through. + test.

**2. Derived-id collision (Bug).** Made `is_url_safe_nod` strictly ASCII alphanumeric (no `-`/`_`). Since a nod can no longer contain the `-` separator, `{base}-{nod}` is unambiguous and two sources can't derive the same id (`radar-a`+`b-fivih` vs `radar-a-b`+`fivih`). Added a defence-in-depth guard too: a per-site id that collides with an already-registered collection (e.g. a hand-named inline `[[collections]]`) is skipped with an error rather than silently overwritten.

**3. Vanished site invisible in `/health` (Bug).** On reload, a per-site radar that was `Ready` but is absent from the new scan (files aged out / radar offline) now gets a `Degraded` `/health` entry — only when its base source is still configured (a fully-removed source is a config change, not a data gap). It's `Degraded`, so it doesn't affect the reload guard.

**4. Per-quantity style snapshot at load (pre-existing, documented).** WMS/Maps/Tiles per-quantity styles are registered from `raster_info()` at load time, so a moment a later poll brings in falls back to the collection default colormap until the next reload. This is the same load-time snapshot GeoTIFF/QueryData have; re-running `register_parameter_layer_styles` from the engine's poll loop would cross the engine→API-state boundary the architecture deliberately separates. Documented inline (and noted that PVOL moment sets are firmware-dependent, so a reload is required when they change) rather than re-architecting.

Green locally: engine-odim 106 lib + 23 integration, server 49+10, clippy clean.

mrauhala:
Round 7 addressed in ae4c0ab.

**1. (HIGH) Bare `LAYERS={site}` / `GET /maps/{site}/map` → 400.** The view's `get_raster_tile` now defaults a `None` parameter to the site's primary (first) quantity instead of erroring — matching the GRIB engine (whose `get_raster_tile` defaults `None` to its first parameter). The WMS/Maps handlers pass `None` for a bare layer, so the per-site multi-parameter collection was unrenderable without an explicit `/quantity` suffix. Verified: bare render now succeeds. + unit + real-fixture integration tests.

**2. (MEDIUM) All-null z-pinned `PointSeries` → 200.** `site_coverages`' `Some(levels)` path now drops a level whose series is all-null across every quantity/timestep, so `finalize_single_site` turns the all-empty result into 404 (e.g. `?z=15` against data that only exists on a lower sweep). `finalize_single_site` checks empty first now (covering the single-level shape) and no longer 500s on the now-reachable empty case. + test.

**3. (MEDIUM) `get_locations` empty-200 vs `query_location` 404.** `get_locations` now returns 404 when the site has dropped from `by_site_meta`, consistent with `query_location` for the same data condition.

**4. (LOW) Reload log/response `loaded` vs `ready`.** The success log and JSON now report `ready` and `degraded` separately instead of `loaded` (which counted Degraded placeholders that wire no servable routes).

Green locally: engine-odim 107 lib + 23 integration, server 49+10, clippy clean.

mrauhala:
Round 8 addressed in 8fe9e49. All four are corrections to fallout from earlier rounds:

**1. `query_area` shaping (reverses round-2 R2-1).** You're right — an `area` query must always return a `CoverageCollection` per OGC EDR; the single-`z` Single-collapse I added in round 2 was correct only for position/location. Reverted: `query_area` always returns a Collection (empty → 404). Test updated.

**2. Vanished-site `strip_prefix` false positive.** The detector now requires the suffix after the base to be exactly `-{nod}` with an alphanumeric nod, not merely to start with `-`, so a still-present base `radar` no longer claims a vanished site from a removed `radar-fi` source.

**3. `degraded` count undercount.** Recomputed `ready`/`degraded` from the final `result.health` (after the vanished-site block appends its Degraded entries), so the response counters match the `collections` array. Dropped the now-unused `loaded`.

**4. Collision `continue` skipped `/health`.** A per-site id collision now pushes a `Failed` `/health` entry before `continue`, matching every other error path in the loop (previously only a log line).

Green locally: engine-odim 107 lib + 23 integration, server 49+10, clippy clean.

PR #288, opened 2026-06-02
docs: ODIM engine docs — drop "model B" jargon, document odim/odim-volume

mrauhala:
Documentation cleanup + additions for the ODIM radar engines, after #282 (per-site PVOL collections).

**Drop "model B" jargon.** "Model B" was just the A/B label from a choice made during implementation — it means nothing to a reader. Removed everywhere (CLAUDE.md heading + code/test comments), replaced with plain "per-site collections" wording. Comment-only, no logic change.

**Bring references current.**
- README engine table: `engine-odim` is `EdrEngine` + `MapEngine` (EDR is implemented, not "coming") and now covers native polar volumes as well as 2-D composites.
- CLAUDE.md: loader enumerates `engine.sites()` (renamed from `site_ids()` during #282 review).

**Document the ODIM engines (were undocumented in the Data Engines section).**
- New "ODIM Radar (HDF5)" README section: `odim` (2-D composites) and `odim-volume` (polar volumes, one collection per radar site) with config examples — per-quantity WMS colormaps, elevation-angle vertical dimension, and the EDR position/locations/area/trajectory surface.
- Listed `odim`/`odim-volume` in the `engine_type` field table.

Docs + comments only; engine-odim still compiles (verified).


mrauhala:
Good catch on the review's follow-up note — folded it into this PR (78fd1ea) rather than deferring: the `engine-odim/src/lib.rs` module doc still had the stale "Phase 1 scope" block listing PVOL trajectory + EDR as deferred to phases 2/3 (both shipped). Rewrote it to describe both engines (`OdimEngine` composites + `PolarVolumeEngine` per-site polar volumes) and dropped the phase language and the internal `[[...]]` planning links. Compiles clean.

mrauhala:
All four fixed in c9bbead — they were real bugs in the new examples, not nits:
1. COMP example: moved `data_path` to the top-level `[[collections]]` block (it's a CollectionConfig field; serde was silently dropping it under `[collections.odim]` → NoSource at boot).
2. COMP example: added the required `parameter` and `unit`.
3. Replaced `LAYERS={site}` with a concrete `LAYERS=radar-fi-volume-fivih/DBZH` and noted the layer is the full per-site collection id `{base}-{nod}` (WMS keys by collection id, not the bare nod).
4. Fixed the stale `<site>:<quantity>` doc on `OdimConfig.parameter` in ds-core/config.rs (#4 from the review — folded in since it's the same jargon).

Issue #178, opened 2026-05-15
api-edr: wrap EdrEngine query calls in spawn_blocking


mrauhala:
## Problem

The api-edr HTTP handlers (`position_query`, `area_query`, `location_query` in `crates/api-edr/src/handlers.rs`) are `async fn`s that call `EdrEngine::query_position` / `query_area` / `query_location` **directly**, without `tokio::task::spawn_blocking`.

Every `EdrEngine` implementation does blocking I/O inside those methods:

- **engine-odim** — `load_composite` does `std::fs::read` + HDF5 parse (1–50 ms/file), and the position query loops once per catalog timestep. With `max_files = 288` (5-min cadence) and no `datetime` filter, a single position request can block a Tokio worker for several seconds.
- **engine-geotiff** — `reader::read_pixel` / `read_bbox` read tiles from disk/S3 (blocking on cache miss).
- **engine-querydata** — memory-mapped reads (page faults are blocking).
- **engine-grib** — byte-range reads.

So this is a cross-cutting api-edr issue, not specific to one engine. By contrast, the WMS/Maps/Tiles handlers already wrap `MapEngine::get_raster_tile` in `spawn_blocking`.

## Fix

Wrap the `EdrEngine` calls in the api-edr handlers in `tokio::task::spawn_blocking` (the engine `Arc`s are `Send + Sync`, so this is mechanical). This covers all current and future EDR engines in one place.

## Context

Surfaced during review of PR #177 (engine-odim EdrEngine, Phase 1.5). The reviewer correctly noted the fix belongs in api-edr's handlers rather than in any single engine.

mrauhala:
Partially addressed in PR #275 (merged). The trajectory handler — the heaviest EDR query path (nodes × z-levels × quantities × timesteps `sample_polar_slant` iterations) — now wraps its engine call in `tokio::task::spawn_blocking` (`crates/api-edr/src/handlers.rs`). Position / area / location queries still call the engine directly on the request worker, so this issue stays open for those.

mrauhala:
**Bump: lazy PVOL pixel loading (PR #290) raises the priority of this.**

[...]

PR #290, opened 2026-06-02
perf(engine-odim): lazy PVOL pixel loading — bounded RAM, non-blocking scan

### LINKED ISSUE #289: engine-odim PVOL: eager full-volume parse doesn't scale (blocks startup + unbounded RAM) — needs lazy pixel loading
## Symptom (production, `nexus`)

`PolarVolumeEngine` **eagerly parses every `.h5` in each source directory and holds every sweep's decoded pixel array in RAM** (the parse cache, `Arc<PolarVolume>`). The initial parse is **synchronous at startup**, so it blocks serving; and the resident set grows with the archive.

- Configured: `fi-radar-single-fipet-pvol`, `fi-radar-single-fivih-pvol` (`data_path = /meteo/data/fmi_radar_single/{radar}`, no caps — the downloader keeps a rolling ~14 h window and cleans the dir, so capping is **not** wanted).
- **3 radars**: fipet 176 files/4.3 GB, fivih 238/4.5 GB (fikor 238/5.2 GB present but unconfigured) → ~8.8 GB of HDF5 parsed synchronously at boot → "everything blocked" for minutes; meteocore RSS **24.5 GiB**.
- **This is only 3 of a 12-radar network.** At full scale the eager model is ~4× → **~75 GB of pixel data / ~100 GB RSS on a 62 GB host → OOM**, plus a much longer startup block. The model does not scale.

**Not a regression from #282** — the scan path (`enumerate_local` → `build_catalog` → `derive_catalog`) is byte-identical to pre-merge; deploy = restart = re-parse exposed a pre-existing design limit. The merge just made it visible.

## Why caps aren't the fix

`max_files` / `time_window` cap *retention* (after parse) — useful only as a safety valve when the downloader fails to clean. The operator wants **all** files in the rolling window usable (history scrubbing across the full network), so the fix must serve every timestep of every radar without holding them all decoded in RAM.

## Fix — lazy, on-demand pixel loading (the GeoTIFF COG pattern)

The 27 MB/file is the `/datasetN/dataM/data` pixel arrays (`read_moment_array` in `pvol.rs`). Everything the catalog/extent/registration needs — `nod`, time, `elangle`, `nbins`/`rscale`/`rstart`, `quantity`/`gain`/`offset`/`nodata` — is in HDF5 **attributes** (`/what`, `/where`), i.e. KB, not MB.

1. **Metadata-only scan.** Add a `read_polar_volume_meta` that reads structure + attributes but **not** the pixel datasets. The catalog (`by_site`, `by_site_meta`) is built from metadata → scan is fast (seconds for the whole network), startup never blocks, and per-site collections register immediately (model-B `sites()` works as today).
2. **On-demand pixel read + bounded LRU.** `polar_sample` / `sample_sweep_moment` fetch a moment's `RawPixels` lazily — read the one `/datasetN/dataM/data` dataset from the file when first needed, cache it in a **bounded** LRU keyed by `(file, sweep, moment)` (size-capped, like the GeoTIFF tile/rendered caches). Recent low-elevation reflectivity stays hot; old/high-sweep moments load on demand and evict.
3. Result: RSS bounded by the LRU (a few GB, configurable) regardless of radar count or window length; startup is non-blocking; every file remains queryable. Scales to 12+ radars.

`max_files` stays as the documented safety valve for a misbehaving downloader.

## Notes / scope

- Touches the `pvol` reader (metadata vs. lazy dataset read), `PolarMoment`/`PolarVolume` (hold a lazy handle or have the sampler fetch via reader+cache instead of eager `RawPixels`), and the cache model (parse cache → metadata catalog + pixel LRU).
- Mirrors `engine-geotiff` (COG byte-range reads + LRU) and the CLAUDE.md rules: compact decode, bounded caches, no startup-blocking scans.
- `effort: large`, `priority: high` — required before the PVOL engine can serve the full radar network.

mrauhala:
Fixes the production scaling problem in #289: the PVOL engine parsed every `.h5` and held every sweep's decoded pixels in RAM, blocking startup for minutes at 100% CPU and growing RSS unbounded (3 radars ≈ 24.5 GiB; a 12-radar network ≈ 75 GB → OOM).

## Change

Pixels load **lazily**, on demand, behind a **bounded cache** — the COG pattern the GeoTIFF engine already uses.

- **`PolarMoment` is metadata-only** — it carries the moment's `/datasetN/dataM/data` dataset path, not the array. `read_polar_volume` reads attributes and validates each moment's shape but **does not decode** the pixel datasets, so a scan is cheap and a parsed volume is KB-scale. A whole network's catalog fits in memory → **fast, non-blocking startup**.
- **`read_moment_pixels`** reads one moment's array on demand.
- **Process-global byte-bounded `PixelCache`** (`quick_cache`, byte-weighted like the GeoTIFF/GRIB caches), keyed by `(file_id, dataset_path)`. **One budget bounds resident pixels across every PVOL collection**, sized by `MC_PVOL_PIXEL_CACHE_MB` (default 1024 MiB) → **RSS bounded regardless of radar count or window length**.
- A `Pixels` context threads the file source through the samplers; on a cache miss it re-fetches the file bytes (local read / S3 get), decodes the one dataset, and caches it. A read failure degrades to transparent/nodata, not a 500.

`max_files` stays as the downloader-failure safety valve. Every file in the rolling window remains queryable.

## Verification

- `cargo test -p engine-odim`: **107 lib + 23 integration** pass. The integration tests render and run EDR position/area/trajectory **against the real `fivih` fixture through the lazy path**, so the on-demand read + decode is exercised end-to-end. Server: 49 + 10. Clippy clean.

## Notes / follow-ups

- The metadata scan still reads each file's bytes once (the pure-Rust `hdf5-reader` parses from an in-memory buffer) but **skips the pixel decompression** that was the 100% CPU cost; RSS no longer holds the decoded arrays. A further optimization — reading only attribute regions, or backgrounding the metadata scan — can come later if the transient scan I/O matters at full scale.
- An EDR vertical profile currently re-reads the file once per moment; a per-file batch decode is a cheap follow-up (map rendering, the hot path, already fetches one moment).
- Cache hit/miss are tracked (`PixelCache::stats`) — wiring them into `/metrics` is a small follow-up.
- **Real-world validation is a `nexus` deploy** against the live archives — happy to help test there.

Branch is off `main`; once the docs PR (#288) lands I'll rebase (both touch `lib.rs`).

Closes #289.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,6 +891,7 @@ dependencies = [
  "ds-storage",
  "hdf5-reader",
  "ndarray",
+ "quick_cache",
  "regex",
  "serde",
  "tempfile",
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -863,10 +863,12 @@ pub async fn trajectory_query(
     // nodes × z-levels × quantities × timesteps `sample_polar_slant`
     // iterations (millions, seconds of CPU). Run it on the blocking pool
     // so it doesn't park a request-serving worker and head-of-line-block
-    // other requests. Safe here: the trajectory path reads an in-memory
-    // `ArcSwap` catalog snapshot and never touches `DataStore` (whose
-    // `block_in_place` would panic on a `spawn_blocking` thread). The
-    // general "wrap every EdrEngine query" work is tracked in #178.
+    // other requests. spawn_blocking is also *required* for correctness under
+    // lazy pixel loading: an S3 cache miss now fetches via `DataStore::get_on`
+    // (`handle.block_on`), which is valid on a `spawn_blocking` thread but
+    // would be the panicking `block_in_place` path if run directly on a worker.
+    // Offloading the other EdrEngine queries (position/area/locations) the same
+    // way is tracked in #178.
     let engine = engine.clone();
     let coords = params.coords.clone();
     let result = tokio::task::spawn_blocking(move || {
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -865,10 +865,11 @@ pub async fn trajectory_query(
     // so it doesn't park a request-serving worker and head-of-line-block
     // other requests. spawn_blocking is also *required* for correctness under
     // lazy pixel loading: an S3 cache miss now fetches via `DataStore::get_on`
-    // (`handle.block_on`), which is valid on a `spawn_blocking` thread but
-    // would be the panicking `block_in_place` path if run directly on a worker.
-    // Offloading the other EdrEngine queries (position/area/locations) the same
-    // way is tracked in #178.
+    // (`handle.block_on`), the valid bridge on a `spawn_blocking` pool thread —
+    // the plain `DataStore::get` uses `block_in_place`, which *panics* on a
+    // spawn_blocking thread (it is only valid on a multi-thread runtime worker).
+    // The still-direct position/area/locations queries keep `DataStore::get`
+    // precisely because they run on a worker; offloading them is tracked in #178.
     let engine = engine.clone();
     let coords = params.coords.clone();
     let result = tokio::task::spawn_blocking(move || {
--- a/crates/engine-odim/Cargo.toml
+++ b/crates/engine-odim/Cargo.toml
@@ -10,6 +10,7 @@ hdf5-reader = "0.4"
 ndarray = "0.17"
 chrono = { version = "0.4", features = ["serde"] }
 arc-swap = "1"
+quick_cache = "0.6"
 regex = "1"
 tokio = { version = "1", features = ["macros", "rt", "sync", "time"] }
 tracing = "0.1"
--- a/crates/engine-odim/src/lib.rs
+++ b/crates/engine-odim/src/lib.rs
@@ -30,4 +30,4 @@ pub mod volume_engine;
 
 pub use engine::{EngineError, OdimEngine};
 pub use pixel_cache::PixelCache;
-pub use volume_engine::{PolarVolumeEngine, PolarVolumeSiteView};
+pub use volume_engine::{pixel_cache_metrics, PolarVolumeEngine, PolarVolumeSiteView};
--- a/crates/engine-odim/src/lib.rs
+++ b/crates/engine-odim/src/lib.rs
@@ -22,10 +22,12 @@
 pub mod catalog;
 pub mod edr;
 pub mod engine;
+pub mod pixel_cache;
 pub mod proj;
 pub mod pvol;
 pub mod reader;
 pub mod volume_engine;
 
 pub use engine::{EngineError, OdimEngine};
+pub use pixel_cache::PixelCache;
 pub use volume_engine::{PolarVolumeEngine, PolarVolumeSiteView};
--- a/crates/engine-odim/src/pixel_cache.rs
+++ b/crates/engine-odim/src/pixel_cache.rs
@@ -29,9 +29,22 @@ impl quick_cache::Weighter<PixelKey, Arc<RawPixels>> for PixelWeighter {
     }
 }
 
+/// Count cap on the negative (known-bad) cache. Bounds the memory a burst of
+/// distinct failing keys can pin; entries age out by LRU so a transient
+/// failure is retried once evicted.
+const NEGATIVE_CAPACITY_ITEMS: usize = 4096;
+
 /// Thread-safe, byte-bounded LRU of decoded moment arrays.
 pub struct PixelCache {
     inner: Cache<PixelKey, Arc<RawPixels>, PixelWeighter>,
+    /// Count-bounded LRU of keys whose pixel read/decode failed. A per-cell
+    /// sampler loop (e.g. `volume_section`, thousands of cells) would otherwise
+    /// re-fetch and re-count a failed moment on every cell, since a failure
+    /// caches nothing — one transient S3 error inflated `pvol_pixel_read_
+    /// failures_total` by thousands and hammered the store (PR #290 review).
+    /// Recording the failure here makes the retry a single no-op lookup and
+    /// the metric increment happen once.
+    negative: Cache<PixelKey, ()>,
     capacity_bytes: u64,
     hits: AtomicU64,
     misses: AtomicU64,
@@ -47,6 +60,7 @@ impl PixelCache {
         let estimated_items = ((capacity_bytes / (256 * 1024)).max(16)) as usize;
         PixelCache {
             inner: Cache::with_weighter(estimated_items, capacity_bytes.max(1), PixelWeighter),
+            negative: Cache::new(NEGATIVE_CAPACITY_ITEMS),
             capacity_bytes,
             hits: AtomicU64::new(0),
             misses: AtomicU64::new(0),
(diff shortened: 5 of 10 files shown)


mrauhala:
## Review findings addressed in `769ed96`

**1. CRITICAL — S3 cache-miss panic.** `fetch_file_bytes` called `DataStore::get()` (→ `block_in_place`, which panics on a `spawn_blocking` pool thread). Added `DataStore::get_on(path, handle)` (drives the fetch on the runtime handle, mirroring `get_range_on`) and threaded an `Option<&Handle>` through `Pixels`: `Some(Handle::current())` from the `spawn_blocking` sites (`get_raster_tile`, `query_trajectory`), `None` from the request-worker sites (EDR `query_position`/`query_area`/`query_location`, where `block_in_place` is correct and `handle.block_on` would itself panic). The two contexts are documented at each `Pixels {}` site.

**2. HIGH — flaky test.** `polar_sample_blends_across_rays_end_to_end` was the only test writing a *divergent* array (value = ray) under the shared key; all others seed the identical deterministic `synthetic_raw()` (value = bin). Isolated it to a unique key via a new `unique_file_id()` helper and documented the shared-key invariant on `TEST_FILE`. **Verified deterministic across 3 back-to-back runs.**

**3. MEDIUM — silent degradation.** Added a failure counter; `/metrics` now exposes `pvol_pixel_read_failures_total` plus `pvol_pixel_cache_{hits,misses,bytes,capacity}` (gated on PVOL collections being loaded).

**4. MEDIUM (deferred) — `Arc<str>` allocs in `get`.** A borrow-based lookup needs a custom key newtype + `Equivalent` impl; the geotiff/grib `quick_cache` caches also allocate their key on `get`, so this is left as a consistent micro-opt follow-up rather than a one-off here.

`cargo fmt` + `clippy -D warnings` clean; engine-odim 107+23, server 49+10.

mrauhala:
## Review round 2 addressed in `dfc2824`

**1. CRITICAL — cross-source cache collision.** `PIXEL_CACHE` was keyed on the bare file id. A local id is an absolute path (unique), but an S3 object key is unique only *within its bucket* — two S3 PVOL sources with the same key layout but different buckets collided and served each other's pixels. New `pixel_cache_id()` qualifies a remote id with `endpoint`+`bucket` (the bare id stays the fetch object path). Unit tests: `pixel_cache_id_qualifies_remote_by_bucket`, `pixel_cache_id_local_is_bare_path`.

**2. MEDIUM — worker parking on position/area/locations → deferred to #178.** `handle: None` routes those cache misses through `block_in_place` on the request worker. They're **generic handlers shared by every engine**, so they can't be blanket-wrapped in `spawn_blocking` (geotiff/grib rely on `block_in_place` being valid on a worker — wrapping would panic them). Offloading per-engine is exactly **#178** ("wrap EdrEngine query calls in spawn_blocking"); I've bumped that issue with the lazy-loading context. The high-QPS render path + trajectory are already correctly offloaded; these are lower-QPS point queries, and the path is correct (no panic), just not yet offloaded.

**3. LOW — stale comment.** The trajectory handler comment claimed it never touches `DataStore`; with lazy pixels an S3 miss now fetches via `DataStore::get_on` (`handle.block_on`, valid on `spawn_blocking`). Comment updated to say so.

**4. MEDIUM — failure over-count / retry storm.** A failed read cached nothing, so a `volume_section` per-cell loop re-fetched + re-counted thousands of times. Added a bounded negative cache (`is_known_bad`/`mark_bad`): a failed key short-circuits to nodata and the metric increments once. Unit tests: `negative_cache_dedups_failures`, `disabled_cache_never_negative_caches`, `moment_failure_marks_known_bad_and_returns_none`.

`cargo fmt` + `clippy -D warnings` clean; engine-odim 112+23 (deterministic across 2 runs), api-edr green.

mrauhala:
## Review round 3 addressed in `dac589a`

**1. `handlers.rs` — inverted `block_in_place` direction in the comment.** You're right, I got it backwards last round. Corrected: `block_in_place` is *valid* on a multi-thread runtime worker and *panics* on a `spawn_blocking` pool thread — which is exactly why the spawn_blocking trajectory path fetches via `DataStore::get_on` (`handle.block_on`) while the worker-thread position/area/locations queries keep `DataStore::get`. (Now consistent with the `fetch_file_bytes` doc.)

**2. Per-moment full-file download (perf).** Confirmed and quantified — a file with Q cold moments is fetched+parsed Q times (~Q×27 MB). Documented the magnitude at the fetch site and filed **#293** for the batch-decode-on-miss fix (fetch once, decode all moments of the volume into the cache). Deferred as an S3-transfer optimisation — local re-reads hit the page cache, and it's not a correctness issue.

**3. `mark_bad` TOCTOU double-count.** Fixed: replaced the get-then-insert with an atomic `get_or_insert_with`, whose placeholder guard runs the closure only for the winning caller — so concurrent failures of the same key count once. New test `mark_bad_counts_once_under_repeated_calls`.

**4. Diagnostic mode (`MC_PVOL_PIXEL_CACHE_MB=0`) disabled the negative cache.** Fixed: `is_known_bad`/`mark_bad` no longer gate on byte capacity, so the negative cache stays active even when the positive cache is off — no more retry-storm/log-flood in diagnostic mode. Test renamed to `negative_cache_stays_active_when_positive_disabled`.

`cargo fmt` + `clippy -D warnings` clean; engine-odim 113+23 (deterministic ×2), api-edr green.

mrauhala:
## Review round 4 addressed in `a756bf1`

**1. Phantom `pvol_pixel_cache_misses_total` (confirmed).** Fixed: `get()` now counts **hits only**; the loader records a genuine miss via `record_miss()` *after* the `is_known_bad` check, so a known-bad short-circuit counts as neither hit nor miss and no longer inflates the metric per cell. I kept `is_known_bad` *after* the positive `get()` (rather than swapping the order) so a good-key access stays a single lookup — counting the miss at the genuine-miss site gives the same correct metric without adding a negative-cache probe + Arc allocs to the hot `volume_section` path. New test `miss_counted_only_by_record_miss_not_by_get`.

**2. Stale parse-cache geometry → negative-cache poisoning (local-FS only).** Filed **#294**. Root cause is the parse cache not validating mtime/size — pre-existing, not introduced by lazy loading, and not triggered by atomic-rename rotation (operational standard). The negative cache only amplifies the symptom (transparent tiles until LRU eviction). Deferred with a concrete fix plan (validate freshness on parse-cache reuse + drop negative entries on re-parse).

`cargo fmt` + `clippy -D warnings` clean; engine-odim 114+23 (deterministic ×2). All prior rounds' fixes verified clean by the reviewer.

mrauhala:
## Review round 5 addressed in `6c3746f`

**Fixed (both in the /metrics PVOL block):**

**CONFIRMED 2 — `has_pvol` lock-poison.** Was `.map(..).unwrap_or(false)`, which silently suppressed `pvol_pixel_read_failures_total` on a poisoned lock. Now recovers via `into_inner()`, matching every other RwLock read in admin.rs.

**PLAUSIBLE 5 — rebaseline coupling.** The PVOL pixel cache is a process-global that's never replaced, so its counters are strictly monotonic. Replaced the `||`-coupled rebaseline with per-counter `saturating_sub` — simpler, can't underflow, no coupling.

**Tracked, not new (re-cites of deferred items):**

- **CONFIRMED 1 — S3 `block_in_place` on EDR position/area/locations** → **#178**. Deferred and acknowledged in prior rounds; render + trajectory are already offloaded. These are *generic* api-edr handlers shared by every engine, so they can't be blanket-wrapped in `spawn_blocking` (geotiff/grib rely on `block_in_place` being valid on a worker). The proper per-engine offload is #178; impact is S3-only (local `fs::read` doesn't stall).
- **PLAUSIBLE 3 + 4 — negative-cache outage thrashing & stale positive-cache pixels on key reuse** → folded into **#294**. Same root cause (key reuse without freshness validation). Notably, bumping `NEGATIVE_CAPACITY_ITEMS` would *worsen* post-outage recovery (more stale failures pinned), so the right fix is poll-tied cache invalidation, not a bigger cap. 4096 is correct as a within-request bound (~52 keys/request).
- **PLAUSIBLE 6 — per-call `Arc<str>` alloc** → deferred micro-opt, largely subsumed by **#293** (batch-decode removes the per-cell `moment()` repetition that drives the alloc churn).

`cargo fmt` + `clippy -D warnings` clean; server 49+10. The reviewer re-confirmed the core design + all prior fixes sound.

Issue #296, opened 2026-06-02
OGC API Common collections: audit remaining itemType, CRS and HTML gaps


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** HTML metadata and itemType/CRS improvements have shipped in stages. Verify any residual API-specific item rather than implementing the whole original list.

**Next step / remaining scope:** Build a current per-API gap list before implementing additional itemType, CRS or HTML changes.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

PR #292 (closed #291) declared the OGC API - Common Part 1 + Part 2 conformance
classes — EDR/Maps/Tiles now advertise `common-2/conf/{collections,json}` and the
`/collections` structure already validates against the standard. The following
**optional** Part 2 enhancements were listed as out-of-scope in #291 (now closed),
recorded here so they stay tracked:

- **`itemType` on the collection object** — Part 2 default is `"unknown"`; there
  is no registered value for map/coverage items, so omitting it is defensible.
  Add if a client needs it (EDR → `"coverage"`, Tiles vector layer → `"feature"`).
- **`crs` array on Tiles collections** — Tiles currently leans on
  `tileMatrixSetLinks`; Maps already advertises a full `crs` array. Minor
  consistency item.
- **HTML representation of `/collections`** (`common-2/.../conf/html`) +
  `?f=json|html` content negotiation on the metadata endpoints — only JSON is
  served today, and the HTML conformance class is deliberately NOT declared
  (declaring it without an HTML representation would be a false claim).

Genuinely N/A (not tracked): `storageCrs` on EDR (CRS84-only); a "simple-query"
filter on `/collections` (the 20-024 draft defines no such conformance class).

Reference: original assessment in #291; draft https://docs.ogc.org/DRAFTS/20-024.html.

</details>

PR #298, opened 2026-06-02
feat(api): Part 2 optional collection fields + Features Common declarations (#296)

mrauhala:
Partially addresses #296 (structural items only — HTML representation lands separately).

## What

Optional OGC API – Common – Part 2 collection enhancements that #292 left out:

| Change | Surface |
|---|---|
| `itemType: "coverage"` on collection objects | EDR |
| `crs` array on collection objects (parity with Maps) | Tiles |
| `itemType: "feature"` on vector-only collections | Tiles |
| Declare Common Part 1 `{core,landing-page,oas30}` + Part 2 `{collections,json}` | Features |

## Notes

- **Tiles `crs`**: tiles are delivered only in their TileMatrixSet's CRS, so the
  array lists exactly the supported TMS CRSs — `EPSG:3857` (WebMercatorQuad) and
  `CRS84` (WorldCRS84Quad) — *not* Maps' full reproject-to output-CRS list. That
  would over-claim CRSs tiles can't actually be fetched in.
- **`itemType`**: only EDR (`coverage`) and vector tiles / Features (`feature`)
  get one. Map (raster) items have no registered `itemType` value, so it stays
  omitted rather than invented — consistent with the #296 assessment.
- **Features Common declarations**: its landing page, `/conformance`, `/api`, and
  `/collections{,/{id}}` already satisfy Common Part 1 + Part 2 structurally; this
  declares it (parity with what #292 did for Maps/Tiles) and is the prerequisite
  for the planned Common Part 4 searchable-collections support.

The HTML representation of `/collections` and the `common-2 .../conf/html`
conformance class are intentionally **still not declared** — they arrive with the
content-negotiation work.

## Tests

api-edr, api-tiles, api-features suites all pass unchanged.

### DIFF
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -1067,6 +1067,10 @@ fn build_collection_metadata(
         "id": config.id,
         "title": config.title,
         "description": config.description,
+        // OGC API – Common – Part 2 `itemType`. EDR collections expose
+        // coverage data, so "coverage" is the appropriate value (#296). Maps
+        // and raster tiles omit it — map items have no registered value.
+        "itemType": "coverage",
         "links": [
             {
                 "href": format!("{base_url}/edr/collections/{}", config.id),
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -1067,10 +1067,12 @@ fn build_collection_metadata(
         "id": config.id,
         "title": config.title,
         "description": config.description,
-        // OGC API – Common – Part 2 `itemType`. EDR collections expose
-        // coverage data, so "coverage" is the appropriate value (#296). Maps
-        // and raster tiles omit it — map items have no registered value.
-        "itemType": "coverage",
+        // No `itemType`: OGC API – Common – Part 2 registers only "feature"
+        // and "record", and the field describes a /collections/{id}/items
+        // sub-resource — which EDR has no equivalent of (data is reached via
+        // /position, /area, /trajectory, …). EDR collections are also not all
+        // coverage data (CSV/PostGIS serve discrete observations), so no single
+        // itemType applies. Omitted rather than mislabelled (review on #298).
         "links": [
             {
                 "href": format!("{base_url}/edr/collections/{}", config.id),
--- a/crates/api-features/src/handlers.rs
+++ b/crates/api-features/src/handlers.rs
@@ -130,8 +130,28 @@ pub async fn api_definition(State(state): State<AppState>) -> impl IntoResponse 
     let mut collection_paths = json!({});
     for config in state.collections.values() {
         let id = &config.id;
+        let detail_path = format!("/features/collections/{id}");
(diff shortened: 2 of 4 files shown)


mrauhala:
Addressed in a9d4a85:

1. **EDR `itemType: "coverage"` — removed.** Agreed: "coverage" isn't a registered Part 2 itemType (only `feature`/`record`), `itemType` describes a `/collections/{id}/items` sub-resource EDR has no equivalent of, and EDR collections aren't uniformly coverage data (CSV/PostGIS are discrete observations). Now omitted entirely, with a comment explaining why.

2. **Tiles `crs` build — hardened.** Switched the `if let Some(def)` to `.expect(...)` so a `SUPPORTED_TILE_MATRIX_SETS` entry without a matching `get_tile_matrix_set` definition fails loudly (in tests) instead of silently shortening/emptying `crs`.

3. **Features `conf/collections` without `bbox`/`datetime` filtering.** Keeping the declaration: it mirrors the precedent #292 set for EDR/Maps/Tiles, and `bbox`/`datetime` filtering on `/collections` is being added across all four surfaces in the follow-up **Part 4 (Searchable Collections)** PR — which also declares `ogcapi-common-4/1.0/conf/searchable-collections`. That PR resolves the filtering concern for Features and the others together.

(`itemType: "feature"` on vector Tiles collections is unchanged — `feature` is a registered value and those collections are genuinely feature data, served with `/items` via the Features API.)

mrauhala:
Addressed in 97eddc0: removed `itemType: "feature"` from vector Tiles collections.

Agreed — the Tiles router has no `/collections/{id}/items` route (only `/tiles/…`), so advertising `itemType` is an over-claim by the same reasoning that removed it from EDR. A collection that is also a Features collection still carries `itemType` on its `/features` representation, where the `/items` resource actually exists. Tiles tests green; clippy clean.

mrauhala:
Round 3 addressed in 79f7b9a — all three real issues fixed:

1. **`storageCrs ∈ crs[]` (§7.13.3)** — the Tiles `crs` array was only the TMS CRSs (`[EPSG:3857, CRS84]`), so a projected-native raster (`EPSG:3067`/`3035`) would have `storageCrs` outside `crs[]`. Now the native `storageCrs` URI is added to `crs[]` when present. Also ordered **CRS84-first** to match Maps/Features (the ordering nit).

2. **No panic in the request path** — reverted the `crs[]` builder from `.expect()` back to `if let Some` (it runs per-request and there's no `CatchPanicLayer`). The divergence concern from round 1 is now handled where it belongs: a new CI test `tilematrixset::tests::every_supported_tms_resolves` asserts every `SUPPORTED_TILE_MATRIX_SETS` id resolves, so `crs[]` can't silently shorten without a red build.

3. **Features OpenAPI** — `api_definition()` now emits `GET /collections/{id}` per collection, so `common-2 conf/json` is no longer over-claimed (Maps/Tiles already had it).

Both crates green; clippy `-D warnings` clean; new exhaustiveness test passes.

PR #300, opened 2026-06-02
feat(api): OGC API - Common - Part 4 Searchable Collections (bbox/datetime/q/limit)

mrauhala:
Implements the **Searchable Collections** requirements class of [OGC API – Common – Part 4](https://docs.ogc.org/DRAFTS/25-046.html) ("Discovery within many collections") on **EDR, Maps, Tiles, and Features**.

This server routinely exposes *many* collections — the ODIM PVOL model-B auto-split fans one source into N per-site collections — so `/collections` needs server-side discovery rather than always returning the full list.

## What

`GET /collections` now accepts:

| Param | Behaviour |
|---|---|
| `bbox` (+ `bbox-crs`) | Keep collections whose CRS84 footprint intersects. 4- or 6-number bbox, anti-meridian aware. CRS84 only (else 400). |
| `datetime` | Keep collections whose temporal extent intersects this RFC 3339 instant or interval (open ends via `..`). |
| `q` | Free-text, comma-separated OR terms, whole-word & Unicode-case-insensitive, over title + description. |
| `limit` + `offset` | Pagination. |

The response gains `numberMatched`, `numberReturned`, and `self`/`next`/`prev` links. Default/max `limit` is 1000, so realistic catalogs still come back in one page — **no behaviour change** for small deployments, just the extra count fields.

All four `/conformance` lists now declare `ogcapi-common-4/1.0/conf/searchable-collections`, and each `/collections` OpenAPI operation documents the parameters.

## Design

Filtering + pagination + link-query rebuilding lives once in the new **`ds_core::collection_search`** module (chrono + serde only — no `serde_json::Value`, per the ds-core architecture rule). Each crate's `collections()` handler builds lightweight `CollectionMatch` descriptors, calls `search()`, and assembles the JSON; a shared `SearchQueryParams` extractor centralises parse + link rebuilding. Invalid `bbox`/`bbox-crs`/`limit`/`datetime` → HTTP 400.

Resolves EDR spec-compliance **FINDING 17** (collections list missing `numberMatched`/`numberReturned`) — that test is flipped to assert the fix.

**Out of scope (deliberate):** CQL2 and Part 3 → the Sortable, Filterable, and Hierarchical classes are not implemented.

## Tests

- `ds_core::collection_search`: 11 unit tests (bbox incl. anti-meridian, datetime overlap, `q` whole-word + Unicode, pagination cursors, query-string round-trip, parse/validation).
- api-maps: 14 integration tests (counts, `q`, `bbox`, `datetime`, 400s, next/prev links, self-link query preservation).
- api-edr / api-tiles / api-features: handler-wiring smoke tests (conformance class, count fields, 400).
- Full workspace incl. `server` green; `clippy -D warnings` clean.

## Merge order

Part 4 builds on Common Part 2 "collections". EDR/Maps/Tiles already declare it (#292); **Features declares it in #298** (the companion #296 PR) — land #298 first so the Features Part 4 declaration isn't dangling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,6 +95,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "tower",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -580,7 +580,15 @@ pub async fn collections(
         .collections
         .values()
         .filter_map(|config| {
-            let engine = state.engines.get(&config.id)?;
+            let Some(engine) = state.engines.get(&config.id) else {
+                // A registered collection with no engine (e.g. a partial
+                // reload) would otherwise vanish from /collections silently.
+                tracing::warn!(
+                    collection = %config.id,
+                    "collection has no registered EDR engine; omitting from /collections"
+                );
+                return None;
+            };
             let value = build_collection_metadata(engine.as_ref(), config, base);
             Some((
                 config.id.clone(),
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -611,7 +611,7 @@ pub async fn collections(
 
     let link = |rel: &str, offset: usize, title: Option<&str>| {
         let mut o = json!({
-            "href": format!("{base}/edr/collections{}", sp.query_string(offset)),
+            "href": format!("{base}/edr/collections{}", sp.query_string(params.limit, offset)),
             "rel": rel,
             "type": "application/json"
         });
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -607,7 +607,6 @@ pub async fn collections(
         .map(|r| CollectionMatch {
             title: &r.1,
             description: &r.2,
-            keywords: &[],
             bbox: r.3,
             time: r.4,
         })
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -171,6 +171,26 @@ pub async fn landing_page(State(state): State<AppState>) -> impl IntoResponse {
     }))
 }
 
+/// OpenAPI `parameters` array for the OGC API – Common – Part 4 searchable
+/// `/collections` query parameters. Documented per the CLAUDE.md rule and
+/// Part 4 §5.5.
+fn searchable_collections_parameters() -> serde_json::Value {
+    json!([
+        {"name": "bbox", "in": "query", "required": false, "schema": {"type": "string"},
+         "description": "Filter to collections intersecting this CRS84 bbox: 4 (or 6) comma-separated numbers west,south,east,north."},
+        {"name": "bbox-crs", "in": "query", "required": false, "schema": {"type": "string"},
+         "description": "CRS of the bbox values. Only CRS84 is supported."},
+        {"name": "datetime", "in": "query", "required": false, "schema": {"type": "string"},
+         "description": "Filter to collections whose temporal extent intersects this RFC 3339 instant or interval (start/end, ../end, start/..)."},
+        {"name": "q", "in": "query", "required": false, "schema": {"type": "string"},
+         "description": "Free-text search (comma-separated terms, OR) over collection title and description."},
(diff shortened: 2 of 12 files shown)


mrauhala:
Rebased onto `main` (now includes #298) — the `/conformance` conflict in api-features is resolved (Common Part 2 + Part 4 classes both retained).

Review round addressed in 91544bf:

1. **Unknown extent ≡ unbounded** (Part 4 §7.14.2 / §7.14.3) — a collection with no spatial/temporal extent now matches any `bbox`/`datetime` filter rather than being dropped. Fixes cold-start GRIB/ODIM (`bbox: None`) and Features (`time: None`) silently vanishing from filtered responses.
2. **`prev` guard** — `has_prev = offset > 0 && offset <= numberMatched`, so an out-of-range `?offset=` no longer emits a `prev` link that chains through empty pages.
3. **Clamped `limit` in links** — `query_string` now carries the *resolved* limit, so `limit=9999` (clamped to 1000) shows `limit=1000` in self/next/prev hrefs; links still omit `limit` when the client supplied none.

New ds-core tests: `unknown_extent_is_unbounded`, `prev_suppressed_when_offset_out_of_range`, `query_string_reflects_clamped_limit`; updated the bbox filter test for the new semantics. Full workspace + clippy `-D warnings` green.

mrauhala:
Both documentation-correctness issues fixed in ab7eb70 (comment/test-only, no behaviour change):

1. **api-features** — the `collections()` comment that said a `datetime` filter *excludes* no-temporal-extent collections now correctly states they *match* (per the §7.14.3 unbounded fix), pointing at `collection_search::matches`.
2. **api-edr FINDING 5** — flipped from "collections endpoint ignores bbox" to RESOLVED, and strengthened the test to assert real filtering (intersecting bbox → `numberMatched` 1, disjoint → 0), mirroring the FINDING 17 treatment.

api-edr + api-features green; clippy `-D warnings` clean.

(Note: the earlier stale inline re-cites on `collection_search.rs` §7.14.2/.3 and the clamped-limit item were already resolved in 91544bf.)

mrauhala:
All four issues fixed in b42c284:

1. **bbox NaN/Inf → 400** — `parse_bbox` now rejects non-finite values (they were slipping through to a silent empty result).
2. **bbox south > north → 400** — latitude inversion rejected; longitude inversion still allowed for anti-meridian bboxes.
3. **datetime start > end → 400** — inverted intervals rejected in `parse_search_params`.
4. **Silent collection drop** — EDR/Maps/Features `collections()` now emit `tracing::warn!` when a registered collection has no engine (added `tracing` to api-features). Tiles already iterates without dropping.

New ds-core tests `parse_rejects_nonfinite_and_inverted_bbox` and `parse_rejects_inverted_datetime`; full workspace + clippy `-D warnings` green.

mrauhala:
All three fixed in cc33c13:

1. **bbox coordinate range** — latitudes outside [-90, 90] and longitudes outside [-180, 180] now return HTTP 400 (anti-meridian `west > east` still allowed since both bounds stay in range).
2. **Phrase `q` matching** — a whitespace phrase is now matched within a single field (title / description / one keyword), so it can't form a phantom match across a field boundary.
3. **`has_prev` boundary** — switched to strict `offset < number_matched`, matching the page-population guard. Confirmed a `next` link can never produce `offset == number_matched` (`has_next` is false exactly then), so normal paging is unaffected; only an explicit overshoot now gets an empty page with no `prev`.

Three new ds-core tests added; full workspace + clippy `-D warnings` green.

mrauhala:
Both items addressed:

**Finding 1 (dead `keywords`)** — removed in b2ce04a. `CollectionMatch.keywords` and the keyword branches in `q_matches` are gone; `q` now matches title + description (what the OpenAPI already advertises). I went with removal over adding a `CollectionConfig.keywords` field because wiring the field touches ~30 `CollectionConfig` literals across six crates (incl. api-wms / server, which this PR doesn't otherwise touch). Keyword search returns as a complete feature in **#303**. Tests reworked to exercise the same paths via title/description.

**Finding 2 (`/features` `?datetime` no-op)** — agreed it's a pre-existing `FeatureEngine` trait limitation, out of scope here. Tracked in **#303** (add `FeatureEngine::temporal_extent()` and wire it for PostGIS/CSV/GeoJSON so `datetime` filters Features and vector tiles).

Workspace + clippy `-D warnings` green.

PR #304, opened 2026-06-02
feat(api): HTML representation + ?f=json|html content negotiation (#296)

mrauhala:
Closes the remaining (HTML) part of #296. The structural items (itemType, Tiles `crs`, Features Common declarations) shipped in #298; this adds the HTML representation + content negotiation across **EDR, Maps, Tiles, Features**, and declares `ogcapi-common-2/1.0/conf/html`.

## What

- New **`ds_core::html`** module (pure — chrono/serde/std only, no `serde_json`): `escape`, `Wanted`, `negotiate(f, accept)`, `FormatParams`, `LinkView`/`CollectionCard`, and builders `landing_html` / `conformance_html` / `collections_html` / `collection_html`. Branded, fully-escaped, minimal-CSS pages built from primitives the handlers already have.
- The four metadata endpoints (`/`, `/conformance`, `/collections`, `/collections/{id}`) on each surface now resolve `?f=` + `Accept` and branch JSON vs HTML.
- `/collections` HTML honours the Part 4 filters (`?f=html&q=…&bbox=…`) and renders self/next/prev nav.
- `f` added to the shared `SearchQueryParams` (so `collections()` negotiates without a conflicting second extractor) and documented in each `api_definition()`.

## Negotiation rules

`?f=html` → HTML · `?f=json` → JSON · unknown `?f=` → **400** · no `f` + `Accept: text/html` → HTML · otherwise → **JSON** (default unchanged; q-value weighting is a documented v1 simplification).

## Default behaviour is unchanged

No `?f=` / no `Accept: text/html` ⇒ JSON, byte-for-byte as before. All existing JSON integration tests pass unmodified.

## Tests

- `ds_core::html` unit tests (escape all entities, negotiate precedence/default/400, builder escaping).
- Per-surface `content_negotiation` suites: `f=html` → `text/html` on all four endpoints, `f=json` → `application/json`, `Accept: text/html` → HTML, no header → JSON, unknown `?f=` → 400.
- Flipped the EDR/Maps/Tiles conformance tests that asserted `conf/html` ABSENT → now assert PRESENT; flipped EDR FINDING 6 (no `f` param) → RESOLVED.
- Full workspace (incl. `server`) + `clippy -D warnings` green.

### DIFF
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -208,10 +208,18 @@ pub async fn landing_page(
                 .into_response()
         }
         Wanted::Html => {
-            let views: Vec<LinkView> = links
+            let mut views: Vec<LinkView> = links
                 .iter()
                 .map(|(h, r, _, ti)| LinkView::new(h.clone(), *r, Some(ti)))
                 .collect();
+            // rel="alternate" to the JSON representation (parity with the
+            // collection-detail HTML page), so the HTML landing page links to
+            // its machine-readable twin.
+            views.push(LinkView::new(
+                format!("{base}/edr/?f=json"),
+                "alternate",
+                Some("This document as JSON"),
+            ));
             Html(ds_core::html::landing_html(title, description, &views)).into_response()
         }
     }))
@@ -595,11 +603,14 @@ pub async fn api_docs(State(state): State<AppState>) -> impl IntoResponse {
 }
 
 pub async fn conformance(
+    State(state): State<AppState>,
     Query(fp): Query<ds_core::html::FormatParams>,
     headers: HeaderMap,
 ) -> Result<Response, HandlerError> {
-    use ds_core::html::Wanted;
+    use ds_core::html::{LinkView, Wanted};
     let wanted = negotiate(fp.f.as_deref(), &headers)?;
+    let state = state.load_full();
+    let base = &state.base_url;
     let classes = [
         "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/core",
         "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/landing-page",
@@ -622,7 +633,17 @@ pub async fn conformance(
     ];
     Ok(with_vary(match wanted {
         Wanted::Json => Json(json!({ "conformsTo": classes })).into_response(),
-        Wanted::Html => Html(ds_core::html::conformance_html(&classes)).into_response(),
+        Wanted::Html => {
+            let nav = [
+                LinkView::new(format!("{base}/edr/"), "up", Some("Landing page")),
+                LinkView::new(
+                    format!("{base}/edr/conformance?f=json"),
+                    "alternate",
+                    Some("This document as JSON"),
+                ),
+            ];
+            Html(ds_core::html::conformance_html(&classes, &nav)).into_response()
+        }
     }))
 }
 
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -3,8 +3,8 @@ use std::sync::Arc;
 
 use arc_swap::ArcSwap;
 use axum::extract::{Path, Query, State};
-use axum::http::{header, StatusCode};
-use axum::response::{IntoResponse, Response};
+use axum::http::{header, HeaderMap, StatusCode};
+use axum::response::{Html, IntoResponse, Response};
 use axum::Json;
 use serde_json::json;
 
@@ -74,6 +74,20 @@ fn server_error() -> HandlerError {
     )
 }
 
+/// A 400 from a plain message (used for `?f=` content negotiation errors).
+fn bad_request_msg(msg: &str) -> HandlerError {
+    (
+        StatusCode::BAD_REQUEST,
+        Json(json!({ "code": "BadRequest", "description": msg })),
+    )
+}
+
+/// Resolve the requested representation from `?f=` + the `Accept` header.
+fn negotiate(f: Option<&str>, headers: &HeaderMap) -> Result<ds_core::html::Wanted, HandlerError> {
+    let accept = headers.get(header::ACCEPT).and_then(|v| v.to_str().ok());
+    ds_core::html::negotiate(f, accept).map_err(|e| bad_request_msg(&e.to_string()))
+}
+
 /// Shared state for the EDR API: a registry of collection engines + metadata.
 #[derive(Clone)]
 pub struct EdrState {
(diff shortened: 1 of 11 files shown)


mrauhala:
Both bugs fixed in 92281b3:

1. **Pagination dropped `?f=html`** — `page_query_string`/`query_string` now forward the `f` param, so an HTML `/collections` page's `next`/`prev` links stay HTML. Fixed once in `ds_core::collection_search` (all four surfaces benefit). New test `query_string_preserves_format_across_pagination`.
2. **Missing `Vary: Accept`** — added a `with_vary` helper wrapping every negotiated response on the four metadata endpoints across EDR/Maps/Tiles/Features, so shared caches won't cross-serve JSON/HTML. New per-surface tests `negotiated_responses_set_vary_accept` + `accept_header_selects_html`.

**Spec gap 3 (landing self-link type)** — respectfully declining: a `?f=html` request is served by the HTML branch (untyped `<a>` links), so the JSON branch's `type: application/json` self-link is only ever emitted on actual JSON responses, where it's correct. Adding `alternate`-rel cross-links between representations is a reasonable future enhancement, but the current self-link isn't wrong and changing it would alter the (intentionally unchanged) JSON shape.

Full workspace (incl. `server`) + clippy `-D warnings` green; JSON output byte-for-byte unchanged.

mrauhala:
Both items fixed in 9e80120:

1. **Accept q-value** — `negotiate` now honours `text/html;q=0` as an explicit opt-out (no longer served HTML). It matches an explicit `text/html` range with non-zero q; `*/*` still stays on JSON (curl-style clients), and `?f=` still wins. New `accept_allows_html` parser + `negotiate_accept_honours_q_values` test (covers `q=0`, `q=0.9`, `*/*`).
2. **OpenAPI completeness** — `f` is now documented on the landing, `/conformance`, and `/collections/{id}` operations across all four crates (was only on `/collections`), via a shared `format_parameter()` helper that also de-duplicates the inline `f` in `searchable_collections_parameters()`.

Workspace (incl. `server`) + clippy `-D warnings` green; JSON output byte-for-byte unchanged.

mrauhala:
Round 3 addressed in 5a6ea74:

2. **Misleading comment** — reworded the `page_query_string` note (`f` is appended *after* `offset`; the real reason for the inline push vs the `push` closure is to avoid overlapping the closure's mutable borrow of `parts`).
3. **Vary test coverage** — added the collection-detail path to each crate's `negotiated_responses_set_vary_accept` loop (EDR `/collections/weather`, Maps/Tiles `/collections/radar`, Features `/collections/cities`), so `with_vary` on the detail handler is now verified.

1. **Landing HTML self-link drops `?f=html`** — respectfully declining: a browser following an in-page link sends `Accept: text/html` and so gets HTML back via Accept negotiation, and a bare URL is correctly format-neutral for bookmarking/sharing. The paginated `/collections` nav (the case where losing the format actually matters) already preserves `f`. Making *every* HTML-page link carry `?f=html` is a reasonable future enhancement, but the current behaviour isn't a defect.

ds-core + the four API crates green; comment/test-only, no behaviour change.

mrauhala:
Round 4 addressed in 849d1d1:

1. **`text/*` Accept range** — `accept_allows_html` now matches `text/*` (RFC 9110 §12.5.1), not just `text/html`; `Accept: text/*` (non-zero q) negotiates HTML, `*/*` still stays JSON, `q=0` still opts out. Test extended.
2. **`page()` escaping invariant** — documented that `body` must be pre-escaped HTML (only `title` is escaped inside); all builders comply, so the doc just makes the contract explicit for future builders.

ds-core green; clippy clean.

(Note: the other two inline comments this round — the `page_query_string` "above" comment and the `/collections/{id}` Vary-test coverage — were already fixed in 5a6ea74; they appear to be re-cites of the prior round against a pre-fix snapshot.)

mrauhala:
Round 5 addressed in 93f69a3:

1. **`Vary` clobber** — `with_vary` now uses `append` (not `insert`) across all four crates, so it never overwrites a `Vary` an upstream layer set (e.g. compression's `Vary: Accept-Encoding`); HTTP combines multiple `Vary` headers.
2. **Most-specific Accept precedence** — `accept_allows_html` now takes the effective `text/html` q-value from an exact `text/html` range when present, else the `text/*` wildcard (RFC 9110 §12.5.1). So `Accept: text/html;q=0, text/*` correctly yields JSON (the explicit opt-out wins over the wildcard). `*/*` still ignored; `?f=` still wins.

Tests extended for the precedence cases; workspace + clippy `-D warnings` green; JSON output unchanged.

mrauhala:
Round 6 addressed in `e2ff6aa`:

- **`html.rs:93` malformed q-value** — fixed. Unparseable q (`text/html;q=abc`) now parses to `0.0` (not acceptable) instead of `1.0`, so a garbled `Accept` falls back to the safe JSON default. Most-specific precedence preserved: a garbled exact `text/html` still overrides a healthy `text/*`. Regression tests added.
- **`html.rs:156` `class` not escaped** — fixed. `render_links` now `escape()`s the class attribute. Was latent only (both callers pass literals), but the helper no longer trusts any interpolated value.
- **`collection_search.rs:401` borrow-workaround** — fixed. `f` now routes through the `push` closure before the direct `offset` push; the inline duplication and its comment are gone. Query-string test updated for the (semantically irrelevant) param order.
- **`api-edr` collection-id percent-encoding (2 comments)** — **deferred, by design.** IDs are operator config (not HTTP input) and are interpolated into URLs across the whole codebase (every JSON `self`/`items` href, not just these two HTML lines). Per-call encoding here would be whack-a-mole. The correct one-place fix is id-charset validation at config load, tracked in #305.

`cargo fmt` + `clippy -D warnings` clean; ds-core (155) and api-edr (220) tests green.

mrauhala:
Round 7 addressed in `8587c1d`:

- **`html.rs:188` + `handlers.rs:625` — conformance HTML dead-end** — fixed. `conformance_html` now takes `nav: &[LinkView]` and renders it. Each crate's `conformance` handler gained `State` (for the proxy-aware `base_url`) and passes `rel="up"` → landing page and `rel="alternate"` → `…/conformance?f=json`. Applied to all four APIs (edr/maps/tiles/features), not just EDR.
- **`handlers.rs:215` — landing HTML missing `rel="alternate"` to JSON** — fixed. Each landing HTML branch now appends `rel="alternate"` → `…/?f=json`, matching the collection-detail page. All four APIs.
- **`handlers.rs:95` — with_vary/negotiate/bad_request_msg duplicated across 4 crates** — **deferred, by design → #306.** The substantive logic already lives once in `ds_core::html`; the remaining ~14 lines/crate is axum- and error-type-aware glue that can't move into ds-core without breaking the framework-free rule. De-duplicating it means a new `api-common` crate — a separate architectural change, not something to fold into a content-negotiation PR.

`cargo fmt` + `clippy -D warnings` clean; ds-core (155) + all four API crates' tests green.

mrauhala:
Round 8 addressed in `912d93f`:

- **`/collections?f=html` missing `rel="alternate"` to JSON** — fixed. Added `SearchQueryParams::query_string_with_format` (forces the `f` selector, preserves bbox/datetime/q/limit/offset), and each crate's collections HTML branch now appends `rel="alternate"` → `…/collections?…&f=json`. So the alternate link keeps the current filter + page rather than dropping back to an unfiltered default. Applied to all four APIs.

That brings every HTML metadata page (landing, conformance, collections list, collection detail) to parity: each links to its JSON twin. `cargo fmt` + `clippy -D warnings` clean; ds-core (156) + all four API crates green.

mrauhala:
Round 9 reviewed — **merging**. None of the six comments are blocking; all are tracked:

- **q-value not clamped to [0,1]** — benign (`q>1` → HTML, which the client wants; negatives already → JSON). → #307
- **`escape()` doesn't strip `javascript:` scheme** — not exploitable: every href is `format!("{base}/…")`, always `http(s)`-prefixed, so no href can *start* with `javascript:` unless `base_url` itself does. Covered by the id/URL-safety work in #305.
- **Unknown `?f=` → 400** — deliberate and OGC-aligned (only affects invalid `f` on metadata endpoints). Behaviour decision tracked in #307.
- **Pagination links not sticky to `?f=html` when negotiated via Accept** — cosmetic; browsers re-send `Accept: text/html`, so navigation stays HTML. → #307
- **JSON `/collections` missing `rel="alternate" type=text/html`** — HATEOAS symmetry nice-to-have. → #307
- **`with_vary`/`negotiate`/… duplicated ×4** — re-cite of round 7; the shared logic already lives in `ds_core::html`, the remaining axum glue needs an `api-common` crate. → #306

CI green (Check/Lint/Test + Docker Build pass; claude-review pass). Squash-merging.

Issue #287, opened 2026-06-02
engine-odim COMP: template-based timestamp discovery for non-listable HTTP sources (DWD opendata)


mrauhala:
## Problem

DWD opendata publishes ODIM_H5 radar composites over a plain **Apache autoindex** (HTML directory listing), e.g.:

- `https://opendata.dwd.de/weather/radar/composite/hx/`
- files: `composite_hx_YYYYMMDD_HHMM-hd5` (ODIM HDF5; the `.hd5` is just DWD's extension), 5-min cadence, 99 KB–3.6 MB
- plus a fixed `composite_hx_LATEST-hd5` symlink to the newest file

Even with HTTP transport added (#286), the COMP engine's remote scan **discovers files by LISTing the prefix**, and object_store's `HttpStore` lists via **WebDAV `PROPFIND`** — which a plain Apache/nginx autoindex does **not** support. So a LIST-based scan against DWD will fail or return nothing.

## Scope

Add a **discovery mode that doesn't rely on directory listing** for HTTP sources: construct candidate filenames from `filename_template` + recent timestamps and fetch them directly (`HEAD`/`GET`), instead of `list()`.

Two complementary mechanisms:

1. **LATEST shortcut (MVP).** When configured, `GET` a fixed latest-file name (`composite_hx_LATEST-hd5`) every poll. Trivial, no timestamp math — gives "always current" with one request. Good first deliverable.
2. **Timestamp probe (history window).** Walk back from now over the cadence (e.g. 5 min) for `time_window` / `max_files` steps, build each `composite_hx_%Y%m%d_%H%M-hd5` via `filename_template`, and `HEAD` to check existence before `GET`. Bounds the number of probes per cycle (don't loop unboundedly — see the GRIB sequential-network-call lesson in CLAUDE.md; batch/bounded).

Reuse the existing `filename_template` / `timestamp_format` parsing the local source already uses, so the same template drives both filename construction (HTTP) and matching (local/S3).

## Config sketch

```toml
[collections.odim]
# HTTP source (data_path is the http(s):// directory) — depends on #286
filename_template = "composite_hx_%Y%m%d_%H%M-hd5"
poll_interval_secs = 300
time_window = "-PT2H"       # how far back the timestamp probe walks
# discovery = "template"     # vs "list" (default); or auto-detect on list failure
# latest_filename = "composite_hx_LATEST-hd5"   # MVP always-latest mode
```

## Open questions

[...]

PR #310, opened 2026-06-03
feat(engine-odim): HTTP(S) directory sources for the COMP engine (#286)

### LINKED ISSUE #286: engine-odim COMP: support HTTP(S) directory sources (not just S3 + local)
## Problem

The ODIM **composite** engine (`OdimEngine`) supports only two source types — its `build_source` (`crates/engine-odim/src/engine.rs`) matches:

- `endpoint` **+** `bucket` → **S3** (`build_s3_store_from_parts`)
- both absent → **local** directory (`data_path` as a filesystem `PathBuf`)
- anything else → error

There is **no arm for a plain `http(s)://` source**, so you can't point it at an HTTP directory of ODIM composites (e.g. DWD opendata `https://opendata.dwd.de/weather/radar/composite/hx/`). Putting the URL in `data_path` makes it a local path; putting it in `endpoint`+`bucket` builds an *S3* client, and a plain HTTP server doesn't speak the S3 LIST API.

## The capability already exists — the COMP engine just doesn't use it

`ds_storage::build_store()` (`crates/storage/src/lib.rs:183`) already dispatches `http://`/`https://` URLs to `build_http_store` (object_store `HttpStore`). The **GeoTIFF engine already uses HTTP this way** — it calls `ds_storage::build_store(data_path)` directly (`crates/engine-geotiff/src/lib.rs:285`), so an `https://` `data_path` transparently becomes an HTTP store. The COMP engine predates that and hand-rolls S3/local selection instead.

## Scope

Add an HTTP(S) source arm to `OdimEngine::build_source`, mirroring the GeoTIFF precedent:

- When the configured source is an `http(s)://` URL (cleanest: an `http(s)://` `data_path`, consistent with GeoTIFF; alternatively a dedicated field), build a remote `DataStore` via `ds_storage::build_store` / `build_http_store` and produce a `Source::Remote { store, .. }`.
- Reuse the existing `enumerate_remote` → `build_catalog` scan path for the new store.
- `source_label` + logs should name the HTTP base URL.

## Caveat (split into a follow-up)

object_store's `HttpStore` lists via **WebDAV `PROPFIND`**. This works for HTTP stores that support it, but **not** for a plain Apache/nginx autoindex (HTML), which is what DWD opendata serves — there the LIST-based scan will fail. Listable HTTP stores work with this issue alone; non-listable autoindex servers need template-based discovery, tracked separately (see the linked follow-up).

## Acceptance

- A collection configured with an `http(s)://` ODIM-composite source (against a *listable* HTTP store) scans, polls, and serves WMS/Maps/Tiles/EDR like the S3 source.
- Test via the `LocalFileSystem`-as-remote trick already used for the S3 path (`pvol_engine_remote_scan_discovers_*` / COMP `scan_remote`), or a mock HTTP store.

## Prior art

- GeoTIFF: `build_store(data_path)` (HTTP, S3, STAC, local) — `engine-geotiff/src/lib.rs:285`.
- GRIB: HTTP discovery via index sidecar files.

Enables (with the discovery follow-up) DWD opendata radar composites over HTTP.

mrauhala:
## Summary

Adds an **HTTP(S) source arm** to the ODIM composite engine (`OdimEngine`), so a collection can read ODIM_H5 composites from a plain `http(s)://` directory — not just S3 (`endpoint` + `bucket`) or a local directory.

Closes #286.

## Problem

`OdimEngine::build_source` matched only S3 (`endpoint` + `bucket`) or local (`data_path` as a filesystem path). There was no arm for a plain `http(s)://` source: a URL in `data_path` became a local path, and `endpoint` + `bucket` builds an *S3* client that a plain HTTP server doesn't speak.

The capability already existed — `ds_storage::build_store()` dispatches `http(s)://` to an object_store `HttpStore`, and the **GeoTIFF engine already uses it this way** (`engine-geotiff/src/lib.rs:285`). The COMP engine predated that and hand-rolled S3/local selection.

## What changed

- **HTTP arm in `build_source`** (mirrors the GeoTIFF precedent): an `http(s)://` `data_path` (with `endpoint`/`bucket` absent) calls `ds_storage::build_store(data_path)` and produces a `Source::Remote`, reusing the existing `scan_remote` → seed → render path. `build_store` dispatches the URL, so an `amazonaws.com`/`cloudferro.com` URL still resolves to S3.
- **Prefix handling**: the URL path becomes the list prefix; any `prefix_pattern` is appended under it for date-partitioned listable stores (new `combine_http_prefix` helper). `time_window` now applies to HTTP sources.
- **Diagnostics**: `Source::Remote`'s `endpoint`/`bucket` fields are refactored into a `RemoteOrigin` enum (`S3` | `Http`) so `source_label` + logs name the HTTP base URL.
- **`time_window` warning** now fires only for a resolved `Source::Local` — an `http(s)://` `data_path` is correctly exempt instead of being wrongly warned.
- **Refactor**: post-source seed/assemble logic extracted into `OdimEngine::assemble`, shared by `new` and a new `new_remote_for_test` constructor (mirrors `PolarVolumeEngine::new_remote_for_test`).
- Catalog scan error/warn wording: "S3" → "remote" (now covers HTTP too).

## Tests

- Unit: `combine_http_prefix` (base/pattern join, slash normalization, empty cases); `build_source` routing an `http(s)://` `data_path` to a remote source with an `Http` origin and correct label; `prefix_pattern` appended; plain `data_path` still local.
- Integration: `comp_engine_remote_scan_discovers_and_renders_dmi_fixture` drives the full remote `scan_remote` → seed → `get_raster_tile` path over a `LocalFileSystem`-backed `DataStore` (the committed `testdata/odim-dmi-fixture.h5`) — the same offline trick the PVOL remote test and catalog tests already use.

`cargo fmt`, `cargo clippy -p engine-odim -p server`, and `cargo test -p engine-odim` all pass.

## Acceptance (#286)

- ✅ A collection configured with an `http(s)://` ODIM-composite source against a *listable* HTTP store scans, polls, and serves WMS/Maps/Tiles/EDR like the S3 source.
- ✅ Tested via the `LocalFileSystem`-as-remote trick.

## Caveat → follow-up #287

object_store's `HttpStore` lists via WebDAV `PROPFIND`, which a plain Apache/nginx autoindex (e.g. DWD opendata) does **not** support — so a LIST-based scan against such a server won't discover files. Listable HTTP stores work with this PR alone; non-listable autoindex servers need template-based discovery, tracked in #287.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -383,18 +383,22 @@ pub struct OdimConfig {
     pub max_files: Option<usize>,
 
     /// S3-compatible endpoint URL (e.g. `https://s3.waw3-1.cloudferro.com`).
-    /// When `endpoint` + `bucket` are set the engine reads from S3;
-    /// otherwise it scans the collection's local `data_path`.
+    /// When `endpoint` + `bucket` are set the engine reads from S3.
+    /// Otherwise the source is the collection's `data_path`: an
+    /// `http(s)://` URL selects an HTTP(S) object store (COMP only — the
+    /// directory must be WebDAV/`PROPFIND`-listable; a plain Apache/nginx
+    /// autoindex is not), and any other value is a local directory.
     pub endpoint: Option<String>,
     /// S3 bucket name. Required when `endpoint` is set.
     pub bucket: Option<String>,
     /// Object key prefix, optionally carrying strftime templates
     /// (e.g. `%Y/%m/%d/OPERA/COMP/`). Expanded per UTC date on every
-    /// poll so the scan stays current across day boundaries.
+    /// poll so the scan stays current across day boundaries. For an
+    /// `http(s)://` `data_path` (COMP) it is appended under the URL path.
     pub prefix_pattern: Option<String>,
     /// ISO 8601 duration bounding which timesteps to keep, relative to
     /// now (e.g. `-PT12H` for the last 12 hours). Drives both prefix
-    /// date expansion and timestamp filtering for S3 sources. When
+    /// date expansion and timestamp filtering for S3/HTTP sources. When
     /// unset, the scan falls back to a fixed recent-days window.
     pub time_window: Option<String>,
 }
--- a/crates/engine-odim/src/catalog.rs
+++ b/crates/engine-odim/src/catalog.rs
@@ -300,14 +300,14 @@ pub fn scan_remote(
 
     if entries.is_empty() && !errors.is_empty() {
         return Err(DataServerError::Engine(format!(
-            "all {} ODIM S3 prefix scan(s) failed: {}",
+            "all {} ODIM remote prefix scan(s) failed: {}",
             errors.len(),
             errors.join("; ")
         )));
     }
     if !errors.is_empty() {
         warn!(
-            "[catalog] {} ODIM S3 prefix scan(s) failed (kept {} entries from the rest): {}",
+            "[catalog] {} ODIM remote prefix scan(s) failed (kept {} entries from the rest): {}",
             errors.len(),
             entries.len(),
             errors.join("; ")
--- a/crates/engine-odim/src/engine.rs
+++ b/crates/engine-odim/src/engine.rs
@@ -57,20 +57,42 @@ const DEFAULT_SCAN_DAYS: u32 = 2;
 enum Source {
     /// A local filesystem directory, scanned with `read_dir`.
     Local { data_dir: PathBuf },
-    /// An S3/HTTP object store. `prefix_pattern` may carry strftime
-    /// codes (e.g. `%Y/%m/%d/OPERA/COMP/`); it is expanded per UTC
-    /// date on every scan so the listing stays current across day
-    /// boundaries. `time_window`, when set, bounds both the dates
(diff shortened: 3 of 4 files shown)


mrauhala:
Re: the case-sensitive scheme nit on `engine.rs:717` — leaving it as-is in this PR, deliberately.

The reviewer's own note is the reason: `ds_storage::build_store` does the identical case-sensitive check (`lib.rs:210`), so the two sites must change together. If I only lowered-case the check here, an `HTTP://…` `data_path` would route into the new HTTP arm and then fall through `build_store`'s dispatch to `build_local_store`, which fails on `canonicalize` — a *worse* error than today's "not a directory". And touching `build_store`'s dispatch changes behavior for every engine that uses it (GeoTIFF, GRIB, …), which is broader than this focused COMP-HTTP-arm change.

Given uppercase URL schemes essentially never appear in real TOML configs (low real-world impact) and the proper fix is a coordinated, cross-engine `build_store` change, I'd rather not bundle it here. Happy to do it as a small standalone `ci`/`refactor` PR (`eq_ignore_ascii_case` at both sites) if you want the hardening.

PR #312, opened 2026-06-03
fix(engine-odim): fast PVOL startup — bootstrap scan + concurrent fetch

mrauhala:
## Problem

A whole-network `odim-volume` S3 source — e.g. FMI's `fmi-opendata-radar-volume-hdf5` with `prefix_pattern = "%Y/%m/%d/"`, `time_window = "-PT2H"` — made **server startup take minutes**. `PolarVolumeEngine::new` ran a fully synchronous, **sequential** scan that downloaded + HDF5-parsed *every* in-window volume (~12 radars × ~24 five-minute slots ≈ 240 multi-MB files) before the HTTP listener could bind.

Unlike the COMP engine (which catalogs from filenames and downloads a single seed), PVOL must open each file — the site `nod` and sweep geometry live *inside* the HDF5.

### Range-reading just the header? Investigated, rejected.
`hdf5-reader` supports a custom random-access `Storage` (`from_storage`), so in principle we could range-read only the metadata. But an instrumented `Storage` over a real 22 MB fixture showed a structure parse issues **~78,000 tiny reads (≤384 B) scattered across 100% of the file** — there is no front-loaded header block to fetch cheaply. A single full GET is the efficient per-file transfer; the real problem was fetching *every* file *sequentially* at startup.

## Fix

**1. Bootstrap scan at construction (`ScanDepth::Bootstrap`).** `new` now fetches only the **newest volume per site**: list the window, group keys by `stream_key` (basename with the timestamp digit-run masked, e.g. `#_fivih_PVOL.h5`), keep the latest per group. This discovers every active radar and seeds its metadata — which `derive_site_meta` already derives from the latest volume per site — while downloading the theoretical minimum (one file per radar). The background `poll_loop` (already non-blocking, on the poll runtime) backfills the full `time_window` history in `ScanDepth::Full` on its first tick.

*Why "newest per site" and not "latest N slots":* producer uploads are staggered, so the freshest slots are the **least** complete. A slot-based rule under-discovers radars — observed live (7/12 sites). Newest-per-site can't miss an active radar whose newest upload lags. Site registration in `admin.rs` stays synchronous, so all per-site collections appear at boot with no admin reload.

**2. Concurrent fetch.** New `DataStore::get_many` downloads a bounded batch concurrently (`buffer_unordered`) with a **per-object** 30 s timeout — *not* a batch-wide cap. (A batch-wide cap fails the whole chunk once combined transfer exceeds 30 s on a constrained link, losing every site — caught in live testing.) `build_catalog` collects non-cached remote misses and fetches them in `FETCH_CONCURRENCY`-sized chunks, bounding peak memory to ~one chunk of raw volumes (each chunk is parsed to compact `PolarVolume`s and dropped). Speeds up both the bootstrap and the full background poll.

`PendingFile` now carries a `FetchSpec` (Local path / Remote key+store) instead of an opaque fetch closure, so `build_catalog` can batch remotes.

## Verified live against FMI S3

All **12 sites discovered, 12 volumes** (one per radar), every per-site collection registered at boot, no batch timeout. Startup **~22 s on a ~7 MB/s home link** (download-bandwidth-bound; ~seconds on a datacenter link). Was minutes.

```
PVOL bootstrap catalog: 12 site(s), 12 volume(s) (newest volume per site; full history fills on the first poll)
… wired per-site radar collection 'fi-radar-pvol-fivih' (Vihti)   [×12]
Server ready, accepting requests
```

## Tests

- `DataStore::get_many`: input-order results, per-item error isolation (missing object), `max_bytes` oversize guard.
- PVOL `enumerate_remote` bootstrap reduction: asserts exactly the newest volume per site is kept — including a site whose newest upload *lags* the freshest slot (the staggered-upload case).

`cargo fmt --all --check`, `cargo clippy -p ds-storage -p engine-odim -p server --all-targets`, `cargo test -p ds-storage -p engine-odim` all pass.

## Note / possible follow-up

The first **Full** background poll still downloads every in-window volume (now concurrently). Off the startup path and bounded by `time_window`, but heavy transfer on a slow link. If that ever matters, the lever is reducing `time_window` or making *older-timestep structure* parse lazy — a larger refactor, separate from this fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -796,6 +796,7 @@ dependencies = [
  "ds-core",
  "futures",
  "object_store",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/crates/engine-odim/src/volume_engine.rs
+++ b/crates/engine-odim/src/volume_engine.rs
@@ -733,17 +733,17 @@ fn enumerate_local(
     Ok(pending)
 }
 
-/// Extract an acquisition timestamp from an object key by finding the
-/// first run of ≥ 12 consecutive ASCII digits in the basename and
-/// parsing its leading 12 as `%Y%m%d%H%M` (UTC).
+/// Byte range of the acquisition timestamp in `basename`: the first run
+/// of ≥ 12 consecutive ASCII digits, returned as `(start, end)`.
 ///
 /// Handles both source layouts — FMI `202605150000_fivih_PVOL.h5`
 /// (timestamp leads) and DMI `dkste_202512150405.vol.h5` (timestamp
-/// follows the station code). Returns `None` when no such digit run
-/// exists; the caller then keeps the file and relies on the
-/// post-parse `time_window` filter.
-fn parse_key_timestamp(key: &str) -> Option<DateTime<Utc>> {
-    let basename = key.rsplit('/').next().unwrap_or(key);
+/// follows the station code). The single source of truth for "where the
+/// timestamp is," shared by [`parse_key_timestamp`] (which parses it) and
+/// [`stream_key`] (which masks it) so the two can't silently diverge —
+/// the bootstrap newest-per-stream reduction relies on the masked run
+/// being exactly the parsed timestamp.
+fn timestamp_run(basename: &str) -> Option<(usize, usize)> {
     let bytes = basename.as_bytes();
     let mut i = 0;
     while i < bytes.len() {
@@ -753,12 +753,7 @@ fn parse_key_timestamp(key: &str) -> Option<DateTime<Utc>> {
                 i += 1;
             }
             if i - start >= 12 {
-                return chrono::NaiveDateTime::parse_from_str(
-                    &basename[start..start + 12],
-                    "%Y%m%d%H%M",
-                )
-                .ok()
-                .map(|t| t.and_utc());
+                return Some((start, i));
             }
         } else {
             i += 1;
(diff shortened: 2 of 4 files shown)


mrauhala:
Fixed in 95104a4 — good catch. Rather than group by store, I removed the per-file store from `FetchSpec::Remote` entirely: a scan has exactly one `Source`, so `build_catalog` now takes that single store and uses it for every remote key. The misleading "future mixed-source" comment is gone, and the trap with it (a `None` store alongside remote keys — unreachable in practice — now logs a warning instead of fetching through the wrong endpoint).

mrauhala:
Round 2 addressed in c0eb962:

- **Duplicated digit-scan loop** — extracted a single `timestamp_run(basename) -> Option<(start, end)>`; both `parse_key_timestamp` (parses it) and `stream_key` (masks it) now call it, so they can't diverge. Good catch — that coupling is exactly what the bootstrap invariant depends on.
- **Misleading memory comment** — reworded: `get_many` returns a whole chunk's bytes together (peak ~one chunk resident), parsed to compact `PolarVolume`s and freed before the next chunk — not per-file streaming within a chunk.

Two I'm intentionally leaving:

- **`store: None` with remote specs → empty, not `Err`** — `build_catalog` returns a `HashMap` (not a `Result`), and the branch is unreachable: a `Remote` spec is only produced by `enumerate_remote`, which is only called from the `Source::Remote` arm that always passes `Some(store)`. The `warn!` makes a future misuse visible without widening the signature for a path that can't happen today.
- **`PendingFile.id` == `FetchSpec::Remote { key }`** — equal today, but they're conceptually distinct identities (`id` is the parse-cache key; `key` is the object path), and `id` is shared with the `Local` variant. Collapsing them would couple the cache key to the storage key. Low value; leaving as-is.

(The first two re-cited inline comments about the per-file store / "mixed-source" claim were the round-1 finding — already fixed in 95104a4.)

mrauhala:
Round 3 — both are accurate observations, not bugs; leaving as-is:

- **Outer `Err` warn only catches the bridge failure** — correct: per-object failures come back as per-slot `Err`s in the `Ok(r)` vec and are logged individually in the loop below. The outer arm only fires if `block_on_untimed` can't get/build a runtime — effectively unreachable in the server (always inside a runtime). "PVOL batch fetch failed" is fair for that case; not worth another CI round to reword.
- **HEAD+GET when `max_bytes` is set** — deliberate: it mirrors the COMP engine's `fetch_bytes` head-before-get and preserves the original PVOL closure's behavior (close the list→get TOCTOU window + bound peak memory before pulling bytes). You're right it's largely redundant for immutable radar volumes, and the HEADs are concurrent, so on a download-bound transfer the extra RTT is negligible. Happy to pass `None` to drop the HEAD if you'd prefer to halve request count — just say the word.

PR is green and mergeable.

PR #313, opened 2026-06-03
feat(engine-odim): template-based HTTP discovery for non-listable sources (#287)

### LINKED ISSUE #287: engine-odim COMP: template-based timestamp discovery for non-listable HTTP sources (DWD opendata)

mrauhala:
## Summary

Enables ingesting **DWD opendata** radar composites (and any non-listable autoindex server) directly over HTTP. DWD serves `https://opendata.dwd.de/weather/radar/composite/hx/` as a plain Apache autoindex (`composite_hx_YYYYMMDD_HHMM-hd5`, 5-min cadence). object_store's `HttpStore` lists via WebDAV `PROPFIND`, which an autoindex doesn't support — so the #286 list-based scan returns nothing.

This adds a **listing-free discovery mode**: build candidate filenames from the strftime `filename_template` for timestamps walked back from now over the configured cadence, and probe them with concurrent `HEAD`s.

Closes #287. (Builds on #286/#310, merged.)

## Changes

- **`DataStore::head_many`** (ds-storage) — concurrent bounded HEADs, mirroring `get_many` (`block_on_untimed` + `buffer_unordered`, per-object timeout). `NotFound`/404 → `Ok(None)` so a caller keeps the keys that exist. A plain autoindex answers a direct `HEAD` but not `PROPFIND`, so this is the right discovery primitive.
- **`Source::TemplateHttp` + `discover_template_at(now, …)`** (engine-odim) — align `now` to the cadence grid, walk back `min(window/cadence, max_files, HARD_MAX_PROBES=288)` slots, render each candidate via `time.format(template)` (the strftime inverse of the catalog matcher), HEAD-probe concurrently (16-wide), and turn survivors into `CatalogEntry`s (timestamp known from generation; bytes fetched lazily on render). `now` is a parameter for deterministic testing.
- **Config** — `[odim].discovery = "list"` (default) `| "template"` and `cadence_secs`. Template mode requires `filename_template` (the `filename_pattern` form isn't invertible to a filename) + `cadence_secs > 0`, applies to HTTP only, and an unknown `discovery` value is an explicit error. All validated in `build_source`.
- **Fix COMP `poll_once`** — it wrapped `scan_source` in `spawn_blocking`, which makes `ds-storage`'s `block_in_place` *panic* (swallowed as a `JoinError`), so **no remote source ever refreshed via poll** (S3, HTTP-list, or template) — a pre-existing latent bug. Now runs the scan directly on the poll runtime, matching the GRIB/GeoTIFF/QueryData/PVOL pattern. Required for a live DWD feed to actually refresh.

## Scope

**Probe-only.** The optional fixed-`LATEST`-symlink shortcut (`composite_hx_LATEST-hd5`) is deferred to a follow-up — the timestamp probe already reaches `now`, so it covers the always-current case for DWD.

## Validated live against DWD opendata

Ran the server against `https://opendata.dwd.de/weather/radar/composite/hx/`:

```
ODIM HTTP template source: url=…/composite/hx/ prefix='weather/radar/composite/hx' cadence=300s
Server ready, accepting requests
```

`GET /edr/collections/dwd-hx` → temporal extent of **13 discovered slots** (17:05→18:05 over `-PT1H`); `GET …/position?coords=POINT(10 51)` → a valid CoverageJSON Coverage with 13 `reflectivity` time values. The engine discovered the rolling window by `HEAD` probe (no listing), seed-parsed DWD `hx` as ODIM COMP, and served EDR end-to-end.

## Config example

```toml
[[collections]]
id = "dwd-hx"
engine_type = "odim"
data_path = "https://opendata.dwd.de/weather/radar/composite/hx/"

[collections.odim]
parameter = "reflectivity"
unit = "dBZ"
filename_template = "composite_hx_%Y%m%d_%H%M-hd5"
discovery = "template"
cadence_secs = 300
time_window = "-PT1H"
poll_interval_secs = 300
```

## Tests

- `head_many`: presence in input order, missing object → `Ok(None)`.
- Template `build_source` validation: cadence + filename_template required, HTTP-only, unknown mode rejected.
- Deterministic offline probe over a `LocalFileSystem` `DataStore` with an **injected clock** — present slots found, a gap and a `LATEST` decoy excluded, never listing.

`cargo fmt --all --check`, `cargo clippy -p ds-core -p ds-storage -p engine-odim -p server --all-targets`, `cargo test -p engine-odim -p ds-storage` all pass.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -401,6 +401,19 @@ pub struct OdimConfig {
     /// date expansion and timestamp filtering for S3/HTTP sources. When
     /// unset, the scan falls back to a fixed recent-days window.
     pub time_window: Option<String>,
+    /// Discovery mode for an `http(s)://` `data_path` source (COMP only;
+    /// ignored for local and S3 sources):
+    /// - `"list"` (default) — enumerate the directory with WebDAV
+    ///   `PROPFIND`. Works for listable HTTP stores (#286).
+    /// - `"template"` — don't list; instead build candidate filenames
+    ///   from `filename_template` + `cadence_secs` walking back from now,
+    ///   and probe them with `HEAD`. For non-listable autoindex servers
+    ///   such as DWD opendata (#287). Requires `filename_template`.
+    pub discovery: Option<String>,
+    /// Probe cadence in seconds for `discovery = "template"` — the spacing
+    /// of candidate timestamps (e.g. `300` for a 5-minute radar feed).
+    /// Required in template mode; ignored otherwise.
+    pub cadence_secs: Option<u64>,
 }
 
 fn default_grib_poll_interval() -> u64 {
--- a/crates/engine-odim/Cargo.toml
+++ b/crates/engine-odim/Cargo.toml
@@ -19,3 +19,4 @@ thiserror = "2"
 
 [dev-dependencies]
 tempfile = "3"
+chrono = { version = "0.4", features = ["serde"] }
--- a/crates/engine-odim/src/engine.rs
+++ b/crates/engine-odim/src/engine.rs
@@ -70,7 +70,32 @@ enum Source {
         prefix_pattern: String,
         time_window: Option<TimeWindow>,
     },
+    /// A non-listable HTTP(S) directory discovered by **template probe**
+    /// (#287): instead of `list`, candidate filenames are built from
+    /// `template` (the strftime `filename_template`) for timestamps walked
+    /// back from now over `cadence`, and each is `HEAD`-probed. For
+    /// autoindex servers (DWD opendata) where `list` (WebDAV `PROPFIND`)
+    /// returns nothing. `time_window` bounds how far back to walk;
+    /// `base_url` is diagnostics only.
+    TemplateHttp {
+        store: ds_storage::DataStore,
+        base_url: String,
+        base_prefix: String,
+        template: String,
+        cadence: Duration,
+        time_window: Option<TimeWindow>,
+    },
 }
+
+/// Hard cap on candidate timestamps probed per template scan — a backstop
+/// against an unbounded walk when neither `time_window` nor `max_files`
+/// bounds it. 288 = 24 h at a 5-minute cadence. The probes are concurrent
+/// `HEAD`s (cheap), so this is generous; `time_window`/`max_files` clamp it
+/// far lower in practice (DWD `-PT2H` @ 300 s → 25 probes).
+const HARD_MAX_PROBES: usize = 288;
+
+/// Max concurrent `HEAD` probes in a template scan.
+const PROBE_CONCURRENCY: usize = 16;
 
 /// Diagnostic descriptor for a [`Source::Remote`] store — names the
 /// backend in logs and error messages. The `store` does the real work;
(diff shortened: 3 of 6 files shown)


mrauhala:
Thanks, but I believe this one is inverted — `DateTime::<Utc>::from_timestamp(secs, nsecs) -> Option<DateTime<Utc>>` is the *modern* API (added in chrono 0.4.31) and is **not** deprecated. It's `Utc.timestamp(...)` (the panicking `TimeZone` method) that was deprecated in 0.4.23; `Utc.timestamp_opt(...).single()` was its interim non-panicking replacement, and `DateTime::from_timestamp` superseded both.

Verified: workspace is on chrono 0.4.44, and `cargo clippy -p engine-odim` (CI runs `-D warnings`) emits no deprecation warning for this call — which it would if it were deprecated. Leaving it as `DateTime::<Utc>::from_timestamp`.

PR #314, opened 2026-06-03
feat(engine-odim): DWD COMP collection config + incremental template probe

mrauhala:
## Summary

Two related changes:

1. **New collection config** `collections.d/radar-de-composite-dbzh-http-h5.toml` — the DWD Germany HX radar composite streamed directly from DWD opendata over HTTP via #287 template discovery. It's the HTTP twin of the existing local-fixture `radar-de-composite-dbzh-local-h5`.
2. **Incremental template probe** — answering "why re-probe slots we already have?": the probe now carries forward already-known slots and HEADs only new ones.

## Incremental probe

`discover_template_at` previously re-`HEAD`-ed **every** candidate slot in the `time_window` on every poll — ~25 for `-PT2H`, up to 288 for a 24h window. That's redundant: radar composites are immutable once published, so re-probing a slot already in the catalog is pure overhead.

Now it takes the current catalog (`known`), carries forward already-known in-window slots untouched, and `HEAD`-probes only the rest:
- The freshly-arrived newest slot.
- Any still-missing slot (a gap or late upload), which keeps being retried until it arrives or ages out of the window.

So the **first scan probes the whole window**, but each subsequent poll probes **~one slot**. A wider `time_window` now costs more only at boot, not every poll.

Plumbing: `scan_source` gains `known: &[CatalogEntry]` (list-based S3/HTTP sources ignore it — one `list` already returns the whole set); `assemble` passes `&[]` (cold start), `poll_once` snapshots the catalog before scanning and passes it.

## Config

```toml
id = "radar-de-composite-dbzh-http-h5"
engine_type = "odim"
data_path = "https://opendata.dwd.de/weather/radar/composite/hx/"

[odim]
filename_template = "composite_hx_%Y%m%d_%H%M-hd5"
parameter = "reflectivity"
unit = "dBZ"
discovery = "template"
cadence_secs = 300
time_window = "-PT2H"
poll_interval_secs = 300

[wms]
colormap = "radar_dbz"
```

(DWD's URLs carry no extension, so the template ends in `-hd5` — unlike the local twin which appends `.h5`.)

## Validation

- The collection loads against **live DWD**; `GET /edr/collections/radar-de-composite-dbzh-http-h5` reports the 25 discovered slots (`-PT2H` window).
- **Test**: a known slot whose file is *absent from disk* still appears in the probe result — proving it was carried forward, not re-probed (a re-probe would 404 and drop it). Plus the existing cold-probe test (present slots found, gap + `LATEST` decoy excluded).

`cargo fmt`, `cargo clippy -p engine-odim --all-targets`, `cargo test -p engine-odim`, and `cargo build -p server` all pass.

### DIFF
--- a/collections.d/radar-de-composite-dbzh-http-h5.toml
+++ b/collections.d/radar-de-composite-dbzh-http-h5.toml
@@ -1,0 +1,48 @@
+# DWD Germany radar composite HX (ODIM_H5 v2.3) — HTTP source, template discovery.
+#
+# Streams `composite_hx_YYYYMMDD_HHMM-hd5` composites directly from DWD open
+# data over plain HTTP — no local mirror, no S3. DWD serves the directory as
+# an Apache autoindex (HTML), which `object_store`'s HttpStore can't list (it
+# lists via WebDAV PROPFIND, which autoindex doesn't support). So this uses
+# `discovery = "template"` (#287): candidate filenames are built from
+# `filename_template` + `cadence_secs` walked back from now over `time_window`
+# and probed directly with HEAD — no listing required.
+#
+#   https://opendata.dwd.de/weather/radar/composite/hx/
+#
+# 4400×4800 grid at 250m/px over Germany + neighbours, polar stereographic
+# (lat_0=90, lat_ts=60, lon_0=10 — same family as FMI/OPERA composites). DBZH
+# quantity, u16 pixels (gain=0.00293, offset=-64 → -64..+128 dBZ; nodata=65535,
+# undetect=0). A new composite lands every 5 minutes; DWD keeps a rolling ~48h
+# archive.
+#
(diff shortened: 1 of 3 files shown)


mrauhala:
Agreed, and it's the deliberate, documented assumption (the doc comment notes "radar files are immutable once published"). Worth adding: even if a producer *did* republish at the same timestamp key, the catalog entry only stores the `key` — bytes are fetched lazily on render via `fetch_bytes` (HEAD+GET), so a render would still pull the *current* object at that key. The only window where a same-key change is masked is the engine's warm single-entry composite cache, until a different timestep evicts it. For 5-minute ODIM nowcasts (distinct timestamp per scan, never corrected-in-place) this doesn't arise in practice, so I'm keeping the carry-forward as-is.

Issue #285, opened 2026-06-01
Expose radar sites as an OGC API - Features collection (base-id collection, one Feature per site)


mrauhala:
## Motivation

Model B (#281 / #282) split each `odim-volume` source into per-site **EDR/WMS/Maps/Tiles** collections (`{base}-{nod}`) and **freed up the base id** — there is no longer a network-level collection there. That base id is the natural home for a **site inventory**: an OGC API - Features collection, named after the network's base id from config, where **each radar site is a point Feature** carrying the site's metadata as properties.

This gives clients a single, queryable "where are the radars and what do they measure" endpoint — useful for map overlays (plot the network), discovery (which site to drill into), and tooling — without needing to crawl N per-site collections.

## Goal

For an `odim-volume` source with `id = "radar-fi-volume-local-h5"` and `apis` including `"features"`, register **one Features collection** with `id = "radar-fi-volume-local-h5"` (the base id) whose items are the radar sites:

- `GET /features/collections/radar-fi-volume-local-h5/items` → FeatureCollection, one Point Feature per site
- `GET /features/collections/radar-fi-volume-local-h5/items/{nod}` → that site's Feature
- bbox / limit / offset filtering (and `datetime` → sites with data in the window)

### Feature shape

- **`id`** = ODIM NOD code (e.g. `fivih`)
- **`geometry`** = `Point(lon, lat)` — the antenna position (WGS84)
- **`properties`** (via `ds_core::feature::PropertyValue`, **not** `serde_json::Value` — ds-core rule):
  - `nod`, `name` (ODIM `/what` PLC place name), `wmo` (if present)
  - `latitude`, `longitude`, `antenna_height_m` (MSL)
  - `quantities` — the bare moments the site measures (DBZH, VRADH, ZDR, …)
  - `elevation_angles` — the sweep angles (deg)
  - `coverage_radius_m`
  - `latest_volume_time`, `volume_count`
  - `collection` — the per-site collection id (`{base}-{nod}`) so a client can jump straight to its EDR/WMS layers

## Design

The base id is free, and a network site list is inherently **network-level** (one collection, N features), so implement `FeatureEngine` on the **owning `PolarVolumeEngine`** (which already holds the `by_site_meta` snapshot) rather than on a per-site view. The engine doesn't implement `EdrEngine`/`MapEngine` (the views do) — adding `FeatureEngine` to it is the clean split: views = per-site data, engine = network inventory.

[...]

PR #315, opened 2026-06-03
feat(engine-odim): human-readable PVOL labels + site-prefixed WMS layer titles

mrauhala:
## Problem

PVOL (`odim-volume`) site collections exposed bare ODIM quantity codes (`DBZH`, `VRADH`, `ZDR`, `RHOHV`, …) as **both** the parameter id and the display label, and WMS child layers were titled with just the bare quantity. In a WMS client that renders a **flat** layer list (ignoring the parent-layer tree), every radar site's child layer showed an identical `DBZH` title — 12 layers called `DBZH` with no way to tell which site each belongs to.

## Changes

**1. Human-readable labels from the ODIM spec** — new `crates/engine-odim/src/quantities.rs` dictionary maps each bare quantity code → (human name, canonical unit). The bare code stays the parameter *id* everywhere (URL short-name, WMS `<Name>`, CoverageJSON key); only the label changes (acronym + name):

- `DBZH` → `"DBZH — Reflectivity (horizontal)"`, unit `dBZ`
- `VRADH` → `"VRADH — Radial velocity (horizontal)"`, unit `m/s`
- `RHOHV` → `"RHOHV — Correlation coefficient"` (dimensionless)
- unknown codes fall back to the bare string (previous behaviour)

Wired into the two label sources in `volume_engine.rs`: `quantity_description()` (EDR/CoverageJSON — now also fills the previously-blank unit) and `derive_site_meta()` (WMS/Maps/Tiles parameter titles).

**2. WMS flat-client disambiguation** — adds `RasterInfo.layer_subtitle: Option<String>` (ds-core). The PVOL site view sets it to the site place name (ODIM `/what` PLC, falling back to NOD); every other engine passes `None`. `api-wms/capabilities.rs` prepends it to each child layer's `<Title>`:

```
Name:  radar-…-fivih/DBZH          (unchanged, requestable)
Title: Vihti — DBZH — Reflectivity (horizontal)
Title: Korpo — DBZH — Reflectivity (horizontal)
```

The parent layer Title (already site-qualified) is untouched. Single-parameter and non-PVOL multi-parameter layers are unaffected (`layer_subtitle = None`).

## Tests

- New `quantities` unit tests (known codes, dimensionless, unknown fallback).
- New `api-wms` capabilities test asserting site-prefixed child `<Title>` while the child `<Name>` stays `{id}/{quantity}`.
- Existing PVOL test `site_view_advertises_bare_quantities_and_single_location` still passes (ids unchanged).
- Full workspace test suite green; `cargo fmt` + `clippy --workspace` clean.

## Not in scope

Radar sites as an OGC API – Features collection remains open issue #285 — untouched here.


mrauhala:
Thanks for the review. Receipts on the two non-blocking findings — decided **not** to change either, with reasoning:

**1. `RasterInfo` `#[non_exhaustive]` / `#[derive(Default)]` — won't apply (would break the build).**
`#[non_exhaustive]` forbids struct-literal construction *outside the defining crate*. In this workspace every engine lives in its own crate (`engine-geotiff`, `engine-grib`, `engine-odim`, `engine-querydata`, …) and constructs `RasterInfo { … }` directly across the `ds-core` crate boundary — that cross-crate literal construction is the core engine pattern (each engine returns a fully-populated `RasterInfo` from `raster_info()`). Marking it `#[non_exhaustive]` would make all of those fail to compile. `#[derive(Default)]` alone doesn't help either, since the callers still spell out every field (none use `..Default::default()`). So the all-callers-patch-on-new-field cost is an accepted, intentional tradeoff here: there is no external `MapEngine` implementor — engines are first-party and in-tree by design.

**2. `quantity_label` `replace('_', " ")` fallback — keeping it intentionally.**
Correct that ODIM codes don't contain underscores, so it's effectively unreachable for real quantities. It's a deliberate carry-over of the exact previous fallback (`label: quantity.replace('_', " ")`) so the unknown-code path is behaviourally identical to pre-PR. Harmless and zero-cost; not worth a behavioural change to remove.

XML-injection, id/name separation, transient-startup, and unit-table points all noted — agreed, no action needed.

Issue #284, opened 2026-06-01
Network composite collection over multiple radar sites (on-the-fly mosaic from PVOLs, e.g. DBZH)


mrauhala:
## Motivation

Model B (#281 / #282) made each radar **site** its own collection (parameter = bare quantity) and deliberately **dropped** any network-level aggregate. That is right for per-site detail, but users also want the *big picture*: a single seamless layer that **composites multiple radar sites** of one network into one product — e.g. a Finland-wide **DBZH** reflectivity mosaic — renderable as one WMS/Maps/Tiles layer and queryable via EDR.

Today the only network view is the **ODIM COMP** engine, which serves a *pre-computed* 2-D composite **shipped by a provider** (DMI/OPERA `.h5`, or a GeoTIFF). That does not help a network where we only ingest **per-site polar volumes** (the FMI `odim-volume` source): there is no provider composite to point at. This issue is about compositing **on-the-fly from the PVOLs we already ingest**.

This is the *complement* to model B, not a replacement — per-site collections stay; a composite collection is an additional, optional view over the same source.

## Goal

One collection (id e.g. `radar-fi-composite`, parameter = bare quantity such as `DBZH`) that, for any requested bbox/CRS/time/elevation:
1. finds every site whose coverage intersects the request,
2. samples each site's chosen quantity at each output pixel (reusing the existing per-cell polar sampler),
3. reduces overlapping samples with a **per-quantity compositing rule**,
4. returns one raster tile (Maps/WMS/Tiles) — and, ideally, EDR position/area answered from the covering site(s).

## Compositing rules (quantity-dependent — important)

A single rule does **not** fit all moments:
- **Reflectivity (`DBZH`, `TH`)** → **max** of overlapping samples ("max reflectivity composite", the canonical product). Optionally max within a height band.
- **Radial velocity (`VRADH`), dual-pol (`ZDR`, `RHOHV`, `KDP`)** → **nearest-radar** (averaging radial velocities across radars is physically wrong — different viewing angles/aliasing; dual-pol moments aren't additive either).
- Default: `max` for reflectivity-like quantities, `nearest-radar` otherwise; make it configurable per quantity.

## Design options

[...]

PR #316, opened 2026-06-03
feat(engine-odim): radar sites as an OGC API - Features collection (#285)

### LINKED ISSUE #285: Expose radar sites as an OGC API - Features collection (base-id collection, one Feature per site)

mrauhala:
Closes #285.

## Problem

Model B (#281/#282) split each `odim-volume` source into per-site EDR/WMS/Maps/Tiles collections (`{base}-{nod}`) and **freed the network-level (base) id**. That base id is the natural home for a **site inventory**: an OGC API – Features collection where each radar site is a Point Feature. Clients get one queryable "where are the radars and what do they measure" endpoint instead of crawling N per-site collections.

## What this does

For an `odim-volume` source whose `apis` includes `"features"`:

- `GET /features/collections/{base}/items` → FeatureCollection, one Point per site
- `GET /features/collections/{base}/items/{nod}` → that site's Feature
- `bbox` / `limit` / `offset` / `datetime` filtering (`datetime` keeps sites with a volume in the window)

### Feature shape
- `id` = ODIM NOD (e.g. `fivih`), `geometry` = antenna `Point(lon,lat)`
- `properties`: `nod`, `name` (PLC), `wmo`, `latitude`/`longitude`, `antenna_height_m`, `quantities` (array), `elevation_angles` (array), `coverage_radius_m`, `latest_volume_time`, `volume_count`, `collection` (the `{base}-{nod}` id so a client can jump to the per-site EDR/WMS layers)

```json
"properties": {
  "nod": "fivih", "name": "Vihti",
  "quantities": ["DBZH", "RHOHV", "VRADH", "ZDR"],
  "elevation_angles": [0.3, 0.7, 1.5, 3.0],
  "collection": "radar-fi-volume-local-h5-fivih"
}
```

## Design

- **`FeatureEngine` is implemented on the owning `PolarVolumeEngine`** (holds the shared `by_site_meta` snapshot), not the per-site views. Engine = network inventory, views = per-site data — the clean Model-B split. Mirrors the PostGIS station-as-Feature analog.
- **`PropertyValue::List`** (new ds-core variant) gives `quantities`/`elevation_angles` clean JSON arrays; api-features emits a JSON array, the ds-mvt tag encoder flattens `List` → comma-joined scalar tag (MVT tags are scalar). The variant touched exactly the two exhaustive matches (verified by `cargo check --workspace --tests`).
- **`SiteMeta.{wmo,height_m}`** added from the latest volume's `RadarSite`.
- **admin.rs** registers the engine under the base id when `apis` has `features` (defensive base-id collision guard); hot-reload re-runs it so new sites appear as new features.
- **`data_version()`** is a self-contained content hash over the site set, so a poll refresh invalidates any future vector-tile ETag without threading a generation counter through the catalog swap.

## Tests
- ds-core/api-features/ds-mvt: `PropertyValue::List` → JSON array / comma-joined MVT tag.
- engine-odim unit tests: feature projection, bbox/datetime filters, pagination, `get_feature` hit + `FeatureNotFound` miss, `feature_count`, `spatial_extent`, `data_version`.
- engine-odim integration test: full `FeatureEngine` surface end-to-end on the real FMI Vihti fixture (skips if the uncommitted fixture is absent).
- Full workspace test suite green; `cargo fmt` + `clippy --workspace` clean.

## Out of scope (follow-ups)
- Vector tiles **of** the site points (registering the base id in `tiles_feature_engines`) — `data_version` already supports it.
- Per-quantity sub-metadata (units/gain/offset) in properties.
- Network raster composite on the same base id (#284).


mrauhala:
Addressed all three findings in 2bc398c:

1. **Integer overflow in `get_features`** — fixed: `end = offset.saturating_add(limit).min(number_matched)`.
2. **Empty/all-null `List` → spurious empty MVT tag** — fixed: `add_tag` now drops the tag when the flattened string is empty (matching the `Null` path); added `encode_point_drops_all_null_list_tag` test.
3. **Non-deterministic `data_version` hash** — fixed: `quantities` is now sorted before hashing, so the hash is order-independent regardless of how `derive_site_meta` builds the vec (it does sort today at construction, but `data_version` no longer relies on that distant invariant).

Full workspace tests + `clippy --workspace` green.

mrauhala:
Round-2 findings addressed in 2c9b099:

1. **Blocker — `"features"` missing from `odim-volume`'s `supported_apis`** — fixed: added `"features"` so a collection listing it isn't rejected before the base-id registration. (Good catch — this made the whole feature unreachable end-to-end via config.)
2. **`limit == 0` meant "return all"** — fixed: dropped the special-case to mirror `CsvEngine` (limit taken verbatim, `0` ⇒ empty page). The API layer clamps `limit` to `[1,1000]` so `0` only arises internally; added `limit_zero_returns_empty_page_not_all` test.
3. **Double health entry when `sites.is_empty()` + features** — fixed: the `Ready` push is now gated on `!sites.is_empty()`, so the empty-source `Degraded` entry stands alone. The empty inventory is still registered (valid empty FeatureCollection, fills on poll).
4. **Quantities hashed without a separator** — fixed: a `b"|"` delimiter after each quantity (matching `ds-mvt`'s `properties_hash`), so `["AB","CD"]` and `["ABCD"]` no longer collide.

Full workspace tests + `clippy --workspace` green.

mrauhala:
Round-3 findings addressed in 7fa4ea3:

1. **`spatial_extent` advertised the coverage bbox, not the antenna points (OGC conformance)** — fixed: the extent now bounds the antenna *points* only, matching what `get_features` filters `bbox` on. Tightened the test to assert the exact antenna-point extent (`[21.0, 60.0, 24.5, 60.3]`) so a regression back to the coverage bbox fails — note the synthetic fixtures *do* carry a ~100 km coverage bbox, so the test now genuinely exercises the distinction.
2. **`plc` not in the content hash** — folded `plc` + `wmo` into `data_version` (both feed the `name`/`wmo` properties), so a re-published volume that corrects either invalidates the hash even when the nominal time is unchanged. Ahead of the vector-tile follow-up.

Full workspace tests + `clippy --workspace` green.

mrauhala:
Round-4 findings addressed in 703c055 — rather than patch each accessor, I moved all per-request inventory work to **catalog-build time** (the background poll runtime), per the CLAUDE.md O(1)-hot-path rule. `Catalog` now carries precomputed `sorted_nods`, `features_extent`, and `feature_version` (computed once in `derive_catalog`, the single constructor).

1. **`spatial_extent()` O(N) per request** → now an O(1) snapshot read of `features_extent`.
2. **`get_features` / `data_version` sort+alloc per call** → `get_features` iterates the pre-sorted NOD slice (no per-request sort); `data_version` is an O(1) read of the precomputed hash.
3. **Missing `nod`/`plc` separator + absent `height_m`** → `feature_version_of` is now comprehensive and collision-safe: every property-bearing field is folded, with a `0x1f` delimiter after each variable-length string and fixed-width LE bytes for numerics (`nod`/`plc`/`wmo`/`lon`/`lat`/`height_m`/`coverage_radius`/time/count/sorted-quantities/elevation-angles). Added `feature_version_reflects_metadata_corrections` (PLC and height changes flip the hash).
4. **Collision guard** → broadened to all registries (`edr`/`map`/`maps`/`tiles`), mirroring the per-site guard, so the same id can't mean different things across services.

Full workspace tests + `clippy --workspace` green.

mrauhala:
Round-5 findings handled in 5166cc1:

1. **`tiles_feature_collections` missing from the collision guard** — fixed: added it (a tiles-only feature collection lands only in that registry). Good catch.
2. **Misleading error message** — fixed: the health error now names all registries the guard checks (EDR/Map/Maps/Tiles/Features), not just "Features".
3. **Unbounded recursion for nested `PropertyValue::List`** — **declining the runtime depth guard**, with a doc note instead. Rationale: no engine constructs nested lists (the radar engine builds `List<String>`/`List<Float>`), and the geojson loader doesn't emit `List` at all — there is **no path from untrusted input to a `PropertyValue`**, so depth is bounded by construction. Round-1 of this review reached the same conclusion ("depth is bounded by engine code, not user input"). Adding a depth-param threaded through the serializer would harden an unreachable case; I documented the flat-scalar invariant on the variant instead so future engine authors know the contract. Happy to add the guard if you'd prefer defence-in-depth here.

Full workspace tests + `clippy --workspace` green.

mrauhala:
Round-6 findings handled in 641aa8f:

1. **`Failed` health entry on base-id collision** — fixed, and it was actually a latent duplicate-health bug: the guard fires *precisely* when `collection.id` is already registered (so it already has a health entry from whoever registered it first). Pushing a second `Failed` entry for that id created a contradictory duplicate in `/health`. Dropped the push — log only, with wording clarifying that only the Features inventory is skipped and the per-site collections are unaffected.
2. **Empty `datetime` response outside the catalog window** — this is issue #285's chosen semantics (filter to sites with data in the window). Returning 400 (option a) would be non-conformant — OGC API Features wants an empty `200` for "no data in window". Took option (c): documented the behavior in `get_features` so operators expect it.

Full workspace tests + `clippy --workspace` green.

---

For context: the review has now run six rounds. Rounds 2 and 4 caught real, important issues (the `supported_apis` blocker; the hot-path violation); the rest have been correctness polish and documentation. The remaining notes are converging on design choices that are already settled (datetime semantics per #285) or non-conformant alternatives. I believe the substance is exhausted and this is ready to merge — happy to address anything further you flag specifically.

mrauhala:
Round-7 handled in 6fa6dc2:

1. **Inconsistent delimiter (`|` vs `\x1f`)** — fixed: quantities now use `\x1f` like the other variable-length fields in `feature_version_of`.

2. **`site_to_feature` allocates per site per call** — **declining**, with rationale. The cited PostGIS precompute is the *exception*, not the rule: CSV (`engine.rs:328`) and GeoJSON (`loader.rs:107`) both build `Arc::new(properties)` fresh per `get_features` call, exactly like PVOL. PostGIS precomputes only because its properties require expensive per-row `pg_type` coercion against a DB refresh cycle — it caches in a *presentation-layer* `MetadataCache`, not in a raw data structure. PVOL's projection from in-memory `SiteMeta` is trivial by comparison. Precomputing here would couple the volume `Catalog` (pure radar-data grouping) to the OGC `collection_id` (a presentation concern, needed for the `collection` property) to save tens of µs per request — bounded by site count and dwarfed by JSON-serialising the same features. This is distinct from round-4's precompute, which targeted O(1) *capability accessors* (`spatial_extent`/`data_version`) per the CLAUDE.md hot-path rule; `get_features` is the data-serving method and is inherently O(page). Happy to add the precompute if you'd like parity with PostGIS, but I don't think the coupling earns its keep.

Full workspace tests + `clippy --workspace` green.

Issue #317, opened 2026-06-04
Incremental startup: serve each collection as soon as it's ready (not all-or-nothing)


mrauhala:
## Motivation

Today the server is **all-or-nothing** at startup: `server/src/main.rs` binds the listen socket, then runs `admin::load_collections` (`server/src/admin.rs`) to build the **entire** registry synchronously, and only then calls `axum::serve`. No collection is queryable until *every* collection has finished loading.

That penalizes a deployment as slow as its slowest engine. A config with one cheap CSV collection and one PVOL radar network (full S3/HDF5 scan) makes the CSV collection wait behind the radar scan — and during the whole load the socket accepts TCP but no HTTP is served (the readiness gap tracked in #247).

We want startup to **publish each collection the moment it's ready**, so fast collections serve immediately while slow ones keep loading in the background.

## Current behavior (code refs)

- `main.rs`: `TcpListener::bind` → `admin::load_collections(...)` (blocks on the full load) → wrap the finished `LoadResult` states in `ArcSwap` → `axum::serve`.
- `load_collections` iterates `for collection in collections { … }` building all engines into one `LoadResult`, returned only when the loop completes.
- Per-API state (`edr_state`, `features_state`, `wms_state`, `maps_state`, `tiles_state`) is published by swapping the whole `ArcSwap` at once.
- `main.rs` refuses to start with zero successful collections ("Refusing to start an empty server").
- Poll loops already run on the dedicated background runtime (`poll_runtime()`), so the loading machinery for this can follow the same isolation rule.

## Proposed approach (sketch — open to alternatives)

1. Start `axum::serve` **early** against initially-empty `ArcSwap` registries so HTTP is live immediately (this also narrows the #247 gap — collections return 404/`Loading` until ready instead of hanging).
2. Load collections on the background runtime; as each engine finishes, **insert it into the live registries** (per-collection publish) rather than building the whole `LoadResult` first. Registries are keyed by collection id, so incremental insertion is natural — needs an append/swap helper instead of a wholesale swap.
3. Extend `CollectionHealth` with a `Loading`/`Pending` status so `GET /health` distinguishes "still loading" from "failed"; `/collections` should omit (or mark) not-yet-ready collections.
4. Revisit the "refuse to start empty" guard (`main.rs`) — with incremental serving the server legitimately starts empty and fills in; the guard should instead reflect "all configured collections failed" once loading settles.
[...]

PR #319, opened 2026-06-04
feat(server): watch collections_dir and auto-reload on changes (#318)

### LINKED ISSUE #318: Watch collections_dir and auto-reload on add/remove/change (no manual POST /admin/collections/reload)
## Motivation

Per-collection configs live in `collections_dir` (`collections.d/*.toml`). Today changes are only picked up when an operator manually calls `POST /admin/collections/reload` — the hot-reload re-reads config and atomically swaps engines, but **nothing triggers it automatically**. Operators dropping in, editing, or removing a collection file have to remember the reload call (or restart).

We want the server to **watch `collections_dir`** and reload automatically when files are added, changed, or removed — so adding `collections.d/radar-new.toml` registers a new collection, deleting (or renaming to `.toml.disabled`) removes it, and editing one updates it, all without a manual API call.

## Current behavior (code refs)

- `collections_dir` is loaded once at startup and again on each `POST /admin/collections/reload`, via `admin::load_collections` (`server/src/admin.rs`), which atomically swaps the `ArcSwap` registries.
- Loading rules (CLAUDE.md): non-recursive, only `.toml` files, `id` required in each file, a single invalid file rejects the *entire* config (no partial loads), duplicate ids rejected.
- The reload guard (`reload_handler`) keeps the previous working registry if the new load has zero `Ready` collections — so a bad edit doesn't take the server down.

## Proposed approach (sketch)

1. Add a filesystem watcher (the `notify` crate — cross-platform inotify/FSEvents/ReadDirectoryChangesW) over the configured `collections_dir`, **non-recursive** to match the loader.
2. **Debounce** events (editors emit multiple write/rename events; batch rapid changes into one reload) before triggering.
3. On a settled change, run a reload through the **same atomic-swap path** as `POST /admin/collections/reload` so behavior is identical (added/removed/changed files reconciled, duplicate-id and invalid-file rules enforced, previous registry retained if the new load is empty/all-failed).
4. Gate behind a config flag, default **off** for safety, e.g.:
   ```toml
   [server]
   collections_dir = "collections.d"
   watch_collections_dir = true          # default false
   # watch_debounce_ms = 500             # optional
   ```
5. Run the watcher + reload on a background runtime, never the request-serving workers (CLAUDE.md runtime-isolation rule); reuse `poll_runtime()` or a dedicated thread.

## Considerations / risks

- **Atomicity of edits**: a half-written file (editor saving) must not be loaded — debounce + the existing "invalid file rejects the load, keep old registry" guard handle this; verify rename-into-place vs in-place-write both behave.
- **Single bad file** currently rejects the *whole* config — for an auto-watcher this means one typo blocks all updates until fixed; the reload guard keeps the old registry serving, but we should log loudly which file failed. (A future enhancement could pair this with the per-collection incremental publish from #317 so one bad file doesn't block unrelated good ones.)
- **Security**: only the configured dir, non-recursive, `.toml` only — no traversal, no following symlinks outside the dir.
- **Editor noise / atomic-save patterns** differ across platforms (vim writes a temp + rename; some editors truncate-in-place) — the debounce window must absorb these.
- **Removal semantics**: deleting a file removes its collection; ensure poll loops / engines for the removed collection are shut down cleanly (mirrors what reload already does on engine swap).

## Acceptance criteria

- [ ] With `watch_collections_dir = true`, adding a valid `collections.d/*.toml` registers the collection within the debounce window — no manual reload.
- [ ] Removing / `.disabling` a file removes its collection and tears down its engine/poll loop.
- [ ] Editing a file updates the collection.
- [ ] A malformed file does **not** take down the running registry (old registry retained; failure logged with the offending filename).
- [ ] Watcher runs off the request runtime; no request-worker stalls.
- [ ] Feature is opt-in (default off).

## Relationship

- Shares the per-collection add/remove machinery with **#317** (incremental startup serving) — both want to reconcile individual collections against the live registry without a full rebuild; #317's incremental publish would let one bad watched file fail in isolation.

mrauhala:
Closes #318.

## Problem

Per-collection configs in `collections_dir` (`collections.d/*.toml`) were only picked up via a manual `POST /admin/collections/reload`. Dropping in, editing, or removing a collection file required remembering the reload call (or a restart).

## What this does

With `[server] watch_collections_dir = true` (opt-in, default **off**) the server watches the directory and reloads automatically when `.toml` files are added, edited, or removed:

```toml
[server]
collections_dir = "collections.d"
watch_collections_dir = true     # default false
# watch_debounce_ms = 500        # coalesce window, default 500
```

Adding `collections.d/radar-new.toml` registers it, deleting (or renaming to `.toml.disabled`) removes it, editing updates it — no API call.

## Design

- **Shared reload core.** Extracted the reload logic out of `reload_handler` into `admin::do_reload` → `Result<ReloadOutcome, ReloadError>`. The HTTP handler (auth + `try_lock` → 409) and the watcher (`lock().await`) now drive the **exact same** atomic-swap path, 0-ready guard, vanished-site detection, and poll-loop lifecycle. Pure refactor — HTTP behavior unchanged.
- **Watcher.** A `notify` (v8) watcher over the canonicalized `collections_dir` (non-recursive), bridged to a tokio unbounded channel; a debounce loop coalesces an editor's write/rename/delete burst into a single reload.
- **Isolation & safety.** Watcher + `do_reload` run on the background `poll_runtime`, never the request workers (CLAUDE.md hot-path rule). A malformed/empty new config keeps the live registry intact (the existing reload guard) and logs which change was ignored. Startup is best-effort: a watcher init failure logs and the server runs without auto-reload.

## Tests
- Config parsing: defaults (`false` / `500`) + explicit values.
- The `.toml` event filter (`.toml`, `.toml.disabled`, editor temps vs unrelated files).
- **End-to-end**: build a live `AdminState` over a temp `collections.d`, start the watcher, add a second collection file → assert it appears in the live registry; remove it → assert it disappears. (In-crate test, since the server is a bin without a lib target.)
- Full workspace tests + `clippy --workspace` green.

## Notes / out of scope
- Watches only `collections_dir`, not the top-level `config.toml` (a collections.d change re-reads the whole config anyway).
- Pairs naturally with #317 (incremental startup) — that issue's per-collection publish would let one bad watched file fail in isolation instead of the current "keep the whole old registry" behavior.


mrauhala:
All three addressed in 8d69691:

1. **`do_reload` blocks a request worker** — fixed: `reload_handler` now runs the reload on `poll_runtime` (`spawn` an async task + await the `JoinHandle`), so an admin reload no longer parks a request-serving worker — consistent with the watcher and the CLAUDE.md hot-path rule. Kept it `spawn` (not `spawn_blocking`) on purpose, since engine construction calls `ds-storage`'s `block_in_place`, which panics on a spawn_blocking pool thread but is valid on a multi-thread worker.
2. **`watch_debounce_ms = 0` silently disables coalescing** — fixed: `warn!` at watcher start when the debounce is zero.
3. **Watcher vs `ADMIN_TOKEN`** — did both of your concrete asks: documented the asymmetry in CLAUDE.md and added a startup WARN when the watcher is enabled and a token is set. I did **not** add a token check to the reload itself, because the watcher's trust boundary is *filesystem write access to `collections_dir`* — a different control plane than the HTTP token, and an FS event carries no credential to check. Anyone who can write a collection file already controls what the server serves (same as `config.toml` at boot). The doc + WARN make that explicit so operators avoid shared/NFS mounts for `collections_dir`.

Full server tests + `clippy --workspace` green.

PR #321, opened 2026-06-04
perf(engine-geotiff): cache RasterInfo, rebuild on catalog swap (#211)

### LINKED ISSUE #211: perf(engine-geotiff): raster_info() allocates fresh Vec<DateTime> on every Maps/Tiles request

mrauhala:
Closes #211. Part of epic #201.

## Problem

`GeoTiffEngine::raster_info()` is called per Maps/Tiles request (before the render semaphore) and re-derived everything on each call:
- a CRS scan over catalog entries — **with a lazy metadata-load side effect** (for STAC, an HTTP fetch on the request worker);
- a fresh `Vec<DateTime<Utc>>` of all timestamps;
- a grid-size scan.

CLAUDE.md cites this as the canonical violation of the O(1)-from-snapshot hot-path rule.

## Fix

Cache a `RasterInfo` snapshot in an `ArcSwap` on the engine, rebuilt **once per catalog swap** via `refresh_raster_info()` — called from `new()` (after the initial scan), `poll_once()` (after the poll swap), and `store_stac_catalog()` (STAC poll). `raster_info()` becomes an O(1) `ArcSwap` read + a cheap clone. The expensive derivation — including any STAC metadata fetch — now runs at startup or on the background `poll_runtime`, never a request worker.

Key correctness point: `times` always derive from the catalog entry **keys** (fixed at scan time), so the cached snapshot is correct after every swap regardless of lazy per-entry metadata loads (which only affect CRS/grid for STAC, and self-heal on the next poll). For local radar COGs — the epic's subject — metadata is parsed during the scan, so the cache is fully correct from construction.

## Notes
- Added `#[derive(Debug, Clone)]` to `ds_core::map_engine::RasterInfo` (additive).
- Kept the trait return as `RasterInfo` (by value) rather than `Arc<RasterInfo>`: the latter would touch all five engines + caller move-outs in `preview.rs` — disproportionate for this LOW-priority item. The remaining per-call cost is one small struct clone; the recompute and the STAC fetch are gone from the request path.

## Tests
- New: `raster_info` is populated at construction (real projected CRS, non-empty times, grid_size — not the cold-start `CRS:84` placeholder) and stable across calls.
- Full workspace tests + `clippy --workspace` green; existing geotiff render/STAC tests unchanged.

### DIFF
--- a/crates/core/src/map_engine.rs
+++ b/crates/core/src/map_engine.rs
@@ -109,6 +109,7 @@ impl RasterTile {
 }
 
 /// Metadata about a map-capable raster collection.
+#[derive(Debug, Clone)]
 pub struct RasterInfo {
     /// Native CRS identifier (e.g., "EPSG:3067").
     pub native_crs: String,
--- a/crates/engine-geotiff/src/lib.rs
+++ b/crates/engine-geotiff/src/lib.rs
@@ -204,7 +204,9 @@ impl GeoTiffEngine {
     /// metadata. Local entries are `Loaded` at scan (CRS correct from
     /// construction); STAC entries are stubs until first render, so `native_crs`
     /// is the `CRS:84` placeholder until `do_load_metadata` loads one and calls
-    /// [`refresh_raster_info`](Self::refresh_raster_info).
+    /// [`refresh_raster_info`](Self::refresh_raster_info). The STAC cold-start
+    /// `CRS:84` window (GetCapabilities / `/collections` before the first
+    /// render) is tracked as a known limitation in #322.
     fn build_raster_info(&self) -> ds_core::map_engine::RasterInfo {
         let catalog = self.catalog.load();
         let crs_name = catalog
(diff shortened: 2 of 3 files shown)


mrauhala:
Both findings addressed in 779970a — fixed together by making the snapshot derivation pure:

1. **Request-path `block_in_place` via `store_stac_catalog` → `refresh_raster_info` → `build_raster_info` → `ensure_metadata`** — real risk (`store_stac_catalog` is called from `get_raster_tile`'s lazy STAC init, line 1387). Fixed by **dropping the cold-start `ensure_metadata` fetch from `build_raster_info`**: it's now a pure, I/O-free read of already-loaded metadata, so `refresh_raster_info` can run anywhere — including the render path — without touching `DataStore`. Local entries are `Loaded` at scan (verified: `FileEntry::new` → `FileState::Loaded`), so CRS stays correct from construction; STAC entries are stubs (`CRS:84` placeholder) until first render.
2. **`do_load_metadata` missed call site** — now calls `refresh_raster_info()` after its store, so a STAC stub gaining metadata refreshes the snapshot **immediately** (not stale until next poll). Safe because the rebuild is I/O-free.

Net: the only metadata fetch is the *pre-existing* `ensure_metadata` in `get_raster_tile` (separate STAC concern, #221's domain) — this PR no longer adds any. Local-radar path (the epic's subject) is fully correct from construction.

Full workspace tests + `clippy --workspace` green; existing geotiff render/STAC tests unchanged.

mrauhala:
Round 2 addressed in a8f9f79:

1. **STAC cold-start `CRS:84` regression (lib.rs)** — tracked, with your reasoning baked in. Filed **#322** (a known-limitation follow-up) and referenced it from `build_raster_info`'s doc comment. I went with option (a) rather than (b): an eager probe in `new()` can't help because a STAC catalog has *no items* at construction (only the extent) — items are discovered later, so the earliest a probe could run is the first poll. #322 proposes exactly that (load one item's metadata on the **poll runtime**, off the request path). The old eager load only got real CRS pre-render by fetching on the request worker — the #221 anti-pattern this PR removes — so the trade is correct.
2. **Test gap (render_projection.rs)** — added `raster_info_stub_reports_crs84_then_real_crs_after_load` (in-crate, since it needs the private catalog). It builds the engine over the committed fixture, swaps in a synthetic **stub** catalog → asserts `CRS:84` / `grid None` / `times` from keys / **no fetch**, then swaps in a real **Loaded** entry + `refresh_raster_info()` → asserts the real CRS/grid returns. No mock STAC server needed (reuses the fixture's real loaded entry). This guards both the cold-start behaviour and the `do_load_metadata` refresh path.

Full geotiff tests + `clippy` green.

---

Also: apologies for claiming "0 new findings" after round 1 — my check compared a local offset timestamp against GitHub's UTC `Z` timestamps and silently matched nothing. Fixed how I check (reading the latest review summary + all inline comments instead).

mrauhala:
Thanks — acknowledging the round-3 design note (no change). The `catalog`/`raster_info` two-`ArcSwap` window is intentional: `times` are eventually-consistent and can transiently lag the catalog by one entry between the `catalog.store()` and the immediately-following `refresh_raster_info()`. For capability metadata (GetCapabilities/`/collections`) that's benign — the render path reads the catalog directly and is always correct — and collapsing both fields into one `ArcSwap` (or a lock) isn't warranted here. Recorded for future readers.

PR #324, opened 2026-06-05
feat(config): per-collection keywords + license across all APIs

mrauhala:
## Summary

Adds two optional per-collection discovery fields to `CollectionConfig`, surfaced from the shared config through **every** API. No engine-trait changes were needed — `CollectionConfig` already reaches each API's collection builder.

- **`keywords: Vec<String>`** — emitted as the OGC API – Common – Part 2 `"keywords"` array in EDR/Features/Maps/Tiles collection JSON, a `<KeywordList>` in WMS `GetCapabilities`, chips on the HTML collection pages, and matched by `/collections?q=` (whole-word + phrase, alongside title/description).
- **`[collections.license]`** (`title` required, `url` optional) — a `rel="license"` link in the JSON APIs, an `<Attribution>` element in WMS, and a link on the HTML pages. When `url` is omitted and `title` is a plausible SPDX id, the URL is synthesized as `https://spdx.org/licenses/<id>.html`.

```toml
[[collections]]
id = "weather"
keywords = ["weather", "observations", "Finland"]

[collections.license]
title = "CC-BY 4.0"
url = "https://creativecommons.org/licenses/by/4.0/"   # optional
```

## Behaviour details

- WMS elements follow the WMS 1.3.0 schema order: `<KeywordList>` after `<Abstract>`, `<Attribution>` after the `<Dimension>` elements.
- A license with no resolvable URL (free-text title, no `url`) still shows its name in WMS/HTML but produces **no** JSON `rel="license"` link (a link object requires an `href`).
- Validated at config load: empty keyword entries, empty license `title`, and non-`http(s)` license `url` are rejected.
- Resolves the keyword-search "follow-up" previously noted in `collection_search.rs`.

## Tests

- `ds-core`: config parse + `validate()` rejections, SPDX URL synthesis, `q=` keyword search, HTML card render.
- `api-maps`: collection JSON carries `keywords` + `rel="license"` link (explicit URL and SPDX-synthesized).
- `api-wms`: `<KeywordList>` + `<Attribution>` present and in correct schema order.
- All 27 `CollectionConfig { … }` struct literals patched (no `#[non_exhaustive]`, per project convention).

`cargo test --workspace` passes; `cargo clippy --workspace --tests` clean; `cargo fmt` applied.

## Docs

CLAUDE.md config example + a new "Collection Keywords & License" subsection.


mrauhala:
Addressed both review findings in 0cebae6:

1. **Duplicated `license_link()`** — removed from all four API crates (`api-edr/features/maps/tiles`). Each call site now inlines a two-line `json!()` over the existing `LicenseConfig::card_link()` in `ds-core`, so the `rel="license"` link shape lives in one place.
2. **Whitespace-padded keywords** — keywords are now trimmed at deserialization (`de_trimmed_keywords`); " radar " → "radar", and an all-whitespace entry trims to "" which `validate()` still rejects. Added a parse test asserting the trim.

Full workspace tests + clippy clean.

mrauhala:
Addressed round-2 review in f9ea26d:

1. **Missing EDR/Features/Tiles JSON assertions** — added a `metadata_extras` test module to `api-edr`, `api-features`, and `api-tiles` (mirroring `api-maps`): each asserts `keywords` and the SPDX-synthesized `rel="license"` link appear in the collection JSON, and that both are absent when unset.
2. **Deprecated SPDX `+` suffix** — documented on `LicenseConfig::resolved_url` that `GPL-2.0+` synthesizes a URL spdx.org serves only via redirect; recommend an explicit `url` or the `-or-later` id for a canonical href.

The round-1 inline comments on `license_link` duplication and keyword padding were already resolved in 0cebae6 (the round-2 summary confirms 'Correctness — Clean'). Full workspace tests + clippy clean.

mrauhala:
Tightened the two round-3 nits in the latest commit (both were non-blocking):
1. **WMS order test anchor** — now anchors on the layer's own `<Abstract>PVOL radar site</Abstract>` instead of a bare `<Abstract>` (which also matched the service-level Abstract), so the ordering check is meaningful.
2. **Redundant trim in validate()** — `de_trimmed_keywords` already trims at load, so the check is now `k.is_empty()` with a clarifying comment.

The remaining inline comments (license_link dup, keyword padding, `+` SPDX, Tiles assertions) are stale re-cites from rounds 1–2, already resolved in 0cebae6 / f9ea26d.

mrauhala:
Tightened the two round-3 nits in the latest commit (both were non-blocking):
1. **WMS order test anchor** — now anchors on the layer's own `<Abstract>PVOL radar site</Abstract>` instead of a bare `<Abstract>` (which also matched the service-level Abstract), so the ordering check is meaningful.
2. **Redundant trim in validate()** — `de_trimmed_keywords` already trims at load, so the check is now `k.is_empty()` with a clarifying comment.

The remaining inline comments (license_link dup, keyword padding, `+` SPDX, Tiles assertions) are stale re-cites from rounds 1–2, already resolved in 0cebae6 / f9ea26d.

mrauhala:
Fixed the round-4 correctness gap in the latest commit — thanks, this was a real bug.

A free-text license (no resolvable URL) was shown in WMS `<Attribution>` but dropped from HTML/JSON, contradicting the documented behavior, because the HTML card reused the href-gated `card_link()`.

- Added `LicenseConfig::card_label() -> (String, Option<String>)` (title always, href optional) for display; kept `card_link()` (href-required) for the JSON `rel="license"` link.
- `CollectionCard.license` is now `Option<(String, Option<String>)>`; `render_license` emits an `<a>` when an href exists, else the plain name. WMS already showed the name.
- Tests: html renderer name-only case, `card_label()` unit test, and an end-to-end Maps `?f=html` test asserting a free-text license shows its name in HTML while producing no JSON license link.

Now consistent across WMS + HTML (name shown) and JSON (link only with href), matching the docs.

mrauhala:
Addressed round-5 (both non-blocking):
1. **Untrimmed `license.url`** — now trimmed in `resolved_url()` and before the scheme check in `validate()`, mirroring keyword trimming; added a unit test (`  https://x  ` → `https://x`).
2. **No HTTP `?q=keyword` coverage** — added a `GET /collections?q=<keyword>` round-trip test to **api-edr, api-features, api-maps, api-tiles** (using a keyword absent from title/description so it proves the config→tuple→`CollectionMatch` wiring; Features is covered separately since it drops the time slot → different tuple index).

The permissive-SPDX note (`"1.0"` synthesizing a non-existent URL → 404) is the accepted, documented trade-off — no change. Full workspace tests + clippy clean.

mrauhala:
Addressed round-6 (all non-blocking polish):
1. **License on HTML list page** — `collections_html` now calls `render_license` (it only called `render_keywords` before), so the license shows on both list and detail pages, matching the docs. Added a list-page assertion.
2. **Redundant `xmlns:xlink`** — audited: the root `<WMS_Capabilities>` already declares it, so I removed the per-element declaration from all three `<OnlineResource>` emitters (service, legend, attribution) — consistent across the file, valid since the namespace is in scope from the root.
3. **Redundant `title()` getter** — removed; callers use the public `title` field directly.

Full workspace tests + clippy clean.

mrauhala:
Round-7 addressed:
1. **xmlns:xlink** — this was a hypothetical ("*if* the root lacks xmlns:xlink…"). The root `<WMS_Capabilities>` **does** declare it (`capabilities.rs:31`) — which is exactly why round 6 had me remove the redundant per-element ones — so the output is valid and I'm not reverting. To address the real concern (the test wouldn't catch a future regression), I added an assertion that the root carries `xmlns:xlink`.
2. **`license.title` not trimmed** — now trimmed at load via `de_trimmed_string` (like keywords); `validate()` uses `is_empty()`; added a padded-title parse assertion.
3. **Single-param schema-order gap** — added `capabilities_single_param_layer_emits_keywords_and_attribution` covering the `write_layer` branch (the prior test only covered the multi-param parent layer).

Full workspace tests + clippy clean.

mrauhala:
Round-8 addressed:
1. **Hardcoded `type` on the license link** — dropped `"type": "text/html"` from all four JSON builders. An explicit `url` may point to a non-HTML resource, and OGC Common §6.5.2 wants the real media type; `type` is optional, so omitting is correct.
2. **`license.url` trimmed only lazily** — now trimmed at load via `de_trimmed_opt_string` (mirrors `title`/keywords); `resolved_url()`/`validate()` no longer re-trim, and the test exercises the deserialization path.

The reviewer confirmed WMS order, escaping, search, validation, and the xmlns:xlink root invariant are all correct. Full workspace tests + clippy clean.

mrauhala:
Round-9:
1. **Duplicate keywords** — `de_trimmed_keywords` now de-duplicates (exact match, first occurrence wins, order preserved) in addition to trimming, so `["radar", "radar"]` yields a single entry. Added a parse assertion.
2. **`type` on the license link** — note this reverses round 8, where the same review flagged the hardcoded `"type": "text/html"` as a spec bug (an operator-supplied `url` may point to a non-HTML resource; OGC Common §6.5.2 wants the *real* media type). That reasoning still holds, so I'm keeping `type` **omitted** (valid — the field is optional) and have made the omission **explicit** with a code comment at all four sites, which is what this round asked for ("it should be explicit either way").

All five surfaces remain consistent; full workspace tests + clippy clean.

Issue #328, opened 2026-06-05
Testdata consolidation: rename collections + commit minimal one-per-engine fixtures


mrauhala:
## Goal

Consolidate `testdata/` (currently **3.7 GB**, mostly untracked dev data) into a
small, committed, consistently-named fixture set — one runnable local collection
per engine plus documented remote examples — and rename all collections to a
structured scheme.

## Naming scheme

`<engine>-<type>-<producer>-<projection>-<surface/product>-<source>`

| Slot | Values |
|---|---|
| engine | csv · geojson · geotiff · grib · odim |
| type | radar · model · observation · boundary |
| producer | ecmwf · noaa · fmi · dmi · dwd · opera · metno |
| projection | wgs84 · lcc · laea · stere · tm35fin · polar |
| surface/product | surface · dbzh · acrr · volume · municipalities · regions |
| source | local · s3 · http · stac |

**Collection id == `testdata/` directory name.**

## Committed local collections (small data + real test coverage)

| Collection id | Data |
|---|---|
| `csv-observation-fmi-wgs84-surface-local` | weather.csv (shrink) |
| `geojson-boundary-fi-wgs84-municipalities-local` | .geojson (shrink) |
| `geojson-boundary-fi-wgs84-regions-local` | .geojson (shrink) |
| `geotiff-radar-metno-wgs84-dbzh-local` | radar/ |
| `geotiff-radar-metno-lcc-dbzh-local` | radar-native/ |
| `geotiff-radar-fmi-tm35fin-dbzh-local` | radar-tm35fin/ (promoted) |
| `odim-radar-dmi-stere-dbzh-local` | radar-dmi/ (+ fold in odim-dmi-fixture) |
| `odim-radar-dwd-stere-dbzh-local` | radar-dwd/ |
| `odim-radar-opera-laea-dbzh-local` | radar-opera/ |
| `odim-radar-fmi-polar-volume-local` | radar-fmi-pvol/ — **PVOL shrink TBD** |
| `querydata-model-ecmwf-wgs84-surface-local` | sqd 16 KB ✅ |
| `querydata-model-metno-lcc-surface-local` | sqd 277 KB ✅ (new, meps) |

## Remote example collections (no committed data)

`grib-model-ecmwf-wgs84-surface-s3` · `grib-model-noaa-wgs84-surface-s3` ·
`geotiff-radar-metno-lcc-dbzh-stac` · `geotiff-radar-opera-laea-acrr-s3` ·
[...]

PR #329, opened 2026-06-06
test(querydata): commit qdcrop-shrunk .sqd fixtures so the parse tests run

mrauhala:
## Summary

The QueryData parse/engine tests referenced multi-GB `.sqd` files that were never committed, so they **silently skipped in CI** — no real coverage for the QueryData decode path. This commits small `qdcrop` subsets so the tests actually run.

| Fixture | Parse path | Was | Now |
|---|---|---|---|
| `ecmwf-kenya` | WGS84 lat/lon | 486 MB (uncommitted) | **16 KB** (3 params, 16×21, 4 times) |
| `meps` | Lambert Conformal Conic | 2.99 GB (uncommitted) | **277 KB** (the only LCC parse case) |

`qdcrop` (FMI `smartmet-qdtools`) crops a small geographic box, decimates the grid, and keeps a few params/timesteps. It preserves the source projection (no `-P`), so the cropped `meps` stays LCC. Provenance + exact recipes are in `testdata/QUERYDATA_FIXTURES.md`.

## Test changes

- Updated the exact-value assertions (`params.len`, `grid.nx/ny`, `times.len`, spatial extent, grid corners) to the cropped shapes.
- Re-cropped `ecmwf-kenya` with params in the original order (msl, 2t, …) so the param-name and `value(idx, …)` assertions stay unchanged.
- The position-query / raster-tile / projected-tile tests are unchanged — the query point (36.8, −1.3) stays inside the cropped Kenya box and `2t`/`msl` are retained.

`cargo test -p engine-querydata` — all tests now **run** (previously skipped) and pass; `clippy` clean.

## Context

Part of the larger testdata consolidation (#328). This lands the QueryData slice on its own; the dir/collection renames stay in that effort. Validated against the real `QueryData::open` before committing.


mrauhala:
Fixed both:
1. **Inverted extent** — `get_spatial_extent` now normalizes to `[west, south, east, north]` via min/max, so a north-to-south / cropped grid can't emit south>north to WMS/EDR/Maps/Tiles. Good catch — this was a pre-existing bug the original SW-cornered fixture masked; the cropped fixture (NW corner) exposes it, and `engine_spatial_extent` now asserts south<north to guard it. `raster_info` was already correct (uses `geo_transform().bbox()`).
2. **LCC test strength** — `parse_meps_lcc` now asserts the documented shape (`params.len()==2`, `times.len()==3`) instead of just non-empty.

Tests pass, clippy clean.

mrauhala:
Round-2 addressed:
1. **Silent-skip guards → asserts** — converted all 11 `if !fixture { return }` guards (9 engine.rs, 2 parse.rs) + the kenya/meps test guards to `assert!(fixture present)`. Now a future rename/.gitignore hard-fails instead of silently skipping — closing the exact gap this PR set out to fix.
2. **LCC extent end-to-end** — added `engine_spatial_extent_lcc`: builds the engine on the meps fixture and asserts `get_spatial_extent` returns south(60.02) < north(64.96), proving the min/max normalization works for projected grids (FMI's north-to-south storage), not just WGS84.
3. **Removed** the leftover `println!` debug from `parse_meps_lcc`.

14 tests pass, clippy clean.

mrauhala:
Round-3 addressed:
1. **Colon in MEPS filename** — renamed to `20260405T180000Z_meps_northeurope_surface.sqd` (no `:`, so Windows `git checkout` works); the test reads the dir so nothing else changes. Added a portability note to the recipe in `QUERYDATA_FIXTURES.md`.
2. **LCC longitude assertions** — `engine_spatial_extent_lcc` now pins west≈9.04 / east≈19.13 (not just ordering), catching a degrees-vs-metres parse regression.
3. **Large-grid bbox under-reporting** — confirmed pre-existing (two-corner min/max; projected edges bow outward). Tracked as #330 (edge-sampling fix, like engine-geotiff's bbox_to_pixels), not in scope here.

14 tests pass, clippy clean.

mrauhala:
Round-4 (all non-blocking polish):
1. **Corner labels** — `grid_lonlat_corners` asserts now say `corner0`/`cornerN` (index 0 is this fixture's NW corner, so `BL` was misleading).
2. **README clarity** — added a note that `qdcrop -G` snaps to the nearest original grid cells, so the committed bounds (e.g. east 41.5°/south −5.25°) sit just inside the `-G 34,-5,42,5` box — expected, not a wrong file.
3. **`.gitattributes`** — marked `.sqd/.h5/.tif/.grib2`… as `binary` so git never applies text/EOL normalization (pairs with the colon-free filename fix for cross-platform safety).

The large-grid bbox edge-sampling note remains tracked as #330. 14 tests pass, clippy clean.

mrauhala:
Round-5 (no blockers): added failure messages to the two `parse_meps_lcc` count asserts so a CI failure shows fixture context, not just `left == right failed`.

The antimeridian-crossing `get_spatial_extent` case (raw min/max on longitudes spanning ±180° → west>east) is pre-existing and now tracked on #330 alongside the projected-edge-bowing fix.

14 tests pass, clippy clean.

PR #331, opened 2026-06-06
feat(engine-grib): local directory data source (data_path) (#327)

### LINKED ISSUE #327: GRIB engine: support a local directory data source (data_path)
## Summary

The GRIB engine only supports remote data sources. `GribConfig` exposes
`endpoint` + `bucket` + `prefix_pattern` (S3/HTTP via index sidecars) but **no
local directory option** — unlike every other file engine (`geotiff`, `odim`,
`querydata`, `csv`, `geojson`), which all accept a local `data_path`.

## Why it matters

This blocks two things in the testdata consolidation (see the reorg issue):

1. A **committed local GRIB collection** — there's no way to serve GRIB from a
   committed fixture dir, so GRIB is the only engine without a runnable local
   collection.
2. **Real collection-level coverage** — the GRIB decode/index tests currently
   read `testdata/ecmwf/sample-message.grib2` directly as bare unit tests; they
   can't exercise the full engine→collection path the way the other engines do.

## Proposal

Add a local data source to the GRIB engine:

- Extend `GribConfig` with a `data_path` (directory of `.grib2` + index
  sidecars), mutually exclusive with `endpoint`/`bucket` (mirror the
  geotiff/odim local-vs-remote split).
- Resolve it through `ds-storage`'s local object store (same `DataStore`
  abstraction the remote path uses), so byte-range reads work uniformly.
- Keep index discovery identical (ecmwf-json / wgrib2).

## Acceptance criteria

- [ ] A `[grib]` collection with `data_path = "testdata/…"` loads and serves
      EDR/WMS/Maps/Tiles from local `.grib2` + index files.
- [ ] The `ecmwf/` fixture can be promoted to a committed local collection
      `grib-model-ecmwf-wgs84-surface-local`.
- [ ] Local poll loop runs on the background runtime (per the engine perf rules).

## Notes

Follow-up from the testdata consolidation / collection-rename work; that effort
ships GRIB as S3-only and keeps `ecmwf/` + `gfs/` as test-only fixtures until
this lands.

mrauhala:
## Summary

Closes #327. GRIB was the only file engine without a **local** data source — `GribConfig` exposed only S3 `endpoint`+`bucket`+`prefix_pattern`. This adds a local `data_path`, so a committed directory of `.grib2` + index sidecars can back a runnable collection (and real engine-level test coverage), mirroring the geotiff/odim local-vs-remote split.

## Changes

- **`GribConfig`**: new `data_path: Option<String>`. `prefix_pattern` is now `Option<String>` (required only for the remote source; optional for local — matches geotiff/odim). No struct-literal churn (configs are TOML).
- **Engine** (`ScanMode` enum):
  - `Remote { prefix_pattern }` — expands recent dates × run hours (the existing "now"-relative NWP S3 layout).
  - `Local { prefix }` — lists a single literal prefix; no date/run-hour templating, since local fixtures are static. The local `DataStore` is built via the existing `ds_storage::build_store` (roots a `LocalFileSystem` at the path); `data_path` also accepts an `s3://`/`http(s)://` fixed-prefix URL.
  - The list → filter-index → parse → byte-range pipeline is **shared** between modes.
- **Config validation**: exactly one of `data_path` (local) or `endpoint`+`bucket` (S3); remote additionally requires `prefix_pattern`.
- **Fixture** `testdata/grib-local/`: the existing ECMWF sample message (`q` @ 150 hPa — the first message at offset 0, length 572555) + a 1-entry `ecmwf-json` index sharing its basename, so the index→data derivation (`X.index` ↔ `X.grib2`) resolves.

## Acceptance criteria (#327)

- ✅ A `[grib]` collection with `data_path` loads and serves — the integration test constructs the engine and renders a tile end-to-end (list → index → local byte-range read + decode → resample).
- ✅ The `ecmwf/` fixture is promotable to a local collection (the committed `grib-local/` pair proves it).
- ✅ Poll loop unchanged — still spawned on the background runtime by the server.

## Tests

`grib_engine_serves_local_directory` (engine-grib) + 5 config-validation cases (local ok, remote ok, data_path+S3 rejected, no-source rejected, remote-missing-prefix rejected). `cargo test -p ds-core -p engine-grib` passes; `cargo check --workspace --tests` + clippy clean.

The dir/collection rename to `grib-model-ecmwf-wgs84-surface-local` stays in the testdata reorg (#328).


mrauhala:
Round-1 addressed:
1. **Partial-remote config** — `validate()` now requires local XOR remote, and a remote source requires **both** `endpoint` and `bucket` (plus `prefix_pattern`); `endpoint` without `bucket` (or vice versa) is rejected at load. Added `grib_rejects_partial_remote`.
2. **`ScanMode::Local` → `ScanMode::FixedPrefix`** — renamed since `data_path` accepts a fixed-prefix remote URL too; the variant is about the prefix behavior (one literal prefix, no date templating), not the backend.

175 core + grib tests pass, clippy clean.

mrauhala:
Round-2: fixed the `prefix_pattern` doc — it said "ignored" for a local `data_path` but it's actually used as a literal sub-prefix (no date templating, default "" = directory root). Wording now matches the `data_path` doc and the implementation. Everything else was verified correct (validation, FixedPrefix path, settled-prefix behavior, no new SSRF surface).

Issue #125, opened 2026-05-11
Zarr: remaining STAC/kerchunk integration and unmet follow-up phases


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** Zarr local/remote, rendering, area and instances exist. Keep only unfinished STAC/kerchunk or other explicitly unmet phase requirements.

**Next step / remaining scope:** Retain only demonstrably unmet integration phases, including STAC/kerchunk where required. Local/remote Zarr serving is already implemented.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Motivation

Zarr is the cloud-native format for multidimensional scientific arrays. NOAA Open Data, NASA Earthdata, CMIP6, Copernicus Marine, Pangeo, Copernicus EOPF (Sentinel) — all publishing Zarr today. A Zarr engine opens an entire ecosystem (climate, ocean, satellite, air quality) that GeoTIFF + GRIB don't reach.

The trait surface maps cleanly: chunked array with time/y/x (and optionally z) dimensions, much like GRIB without the message-discovery dance. The heavy lifting (byte-range S3 reads, tile cache, reprojection) is already solved in `engine-geotiff` and `engine-grib`.

## Scope

- New `engine-zarr` crate implementing `Engine` + `MapEngine`
- Zarr v2 + v3 support (v3 is the current spec; v2 is still common in practice)
- Storage backends: local, S3, HTTP (reuse `ds-storage`)
- Compression: blosc, zstd, gzip, lz4 (most common in practice)
- CF-conventions metadata for variable units, dimensions, CRS
- Lazy chunk loading via byte-range reads with LRU cache
- Multi-variable collections (one variable per parameter, similar to GRIB)
- Bad-chunking detection: warn at startup if chunk shape implies pathological access patterns (e.g. `time=1, lat=full, lon=full` for time-series workloads)

## Config sketch

```toml
[[collections]]
id = "era5-t2m"
engine_type = "zarr"
apis = ["edr", "wms", "maps", "tiles"]

[collections.zarr]
endpoint = "https://s3.eu-central-1.amazonaws.com"
bucket = "era5-pds"
path = "zarr/2026/01/data/air_temperature_at_2_metres.zarr"
zarr_version = 3
[...]

PR #332, opened 2026-06-06
feat(engine-zarr): Zarr V2/V3 engine — Phase 1 local EDR (#125)

mrauhala:
## Summary

Implements **Phase 1** of #125 — a new `engine-zarr` crate that reads cloud-native multidimensional arrays (Zarr V2/V3) with CF-conventions metadata and serves them over **EDR position queries**.

The Zarr format + codec pipeline (blosc/zstd/gzip/crc32c/sharding/transpose) is delegated to the [`zarrs`](https://github.com/zarrs/zarrs) crate; this engine adds the CF semantics, the OGC domain mapping, and the poll-and-swap lifecycle shared by the other engines.

## What's in Phase 1

- **Local store** (`data_path`), geographic WGS84 lat-lon grid.
- **Multi-variable EDR position queries** with bilinear interpolation (nearest fallback for nodata neighbours); monotonic ascending **or** descending coordinate axes.
- **CF time-axis decoding** (`<unit> since <reference>`, seconds/minutes/hours/days; standard/gregorian/proleptic calendars; non-standard calendars approximated with a WARN).
- **CF packing**: `scale_factor` / `add_offset`, and `_FillValue` / `missing_value` → nodata (compared against the raw pre-scale value).
- **Bad-chunking WARN** at startup for the `time=1, lat=full, lon=full` pathology called out in the issue.
- **Poll-and-swap lifecycle** (`ArcSwap` catalog) on the dedicated background poll runtime, per the engine concurrency rules.

## Deferred (tracked in #125)

- Phase 2: S3/HTTP byte-range backend + chunk LRU cache (config fields are present and validated; the engine rejects remote sources with a clear error for now).
- Phase 3: Map/Tiles/WMS rendering (`MapEngine`). The `engine_type → supported_apis` allowlist is `["edr"]`, so a zarr collection requesting `wms`/`maps`/`tiles` is rejected at load.
- Phase 4: per-item-CRS STAC mode. Phase 5: kerchunk.

## Notable details

- `zarrs` pulls C-backed codecs (blosc/zstd) built via `cmake`+`cc` — the same toolchain `libaec-sys` already needs for GRIB. Verified it builds in CI's environment.
- `zarrs` uses rayon **internally for codec decode** (CPU-bound, not I/O), which is safe with the Phase-1 `FilesystemStore`; the "no `ds-storage` from rayon" hazard is noted for the Phase-2 adapter.
- `retrieve_array_subset::<Vec<T>>` requires the exact dtype, so the read path branches on `data_type()` and widens every supported int/float to `f64`.

## Config

```toml
[[collections]]
id = "era5-t2m"
apis = ["edr"]
engine_type = "zarr"

[collections.zarr]
zarr_version = 3
# parameters = ["t2m"]   # optional variable filter
```

## Tests

- **Network-free committed fixture** `testdata/zarr-era5-t2m` (Zarr V3, gzip, two variables including an int16 packed one with a `_FillValue` block). Regenerate with `cargo run -p engine-zarr --example gen_fixture`. The field is linear in lat/lon, so bilinear interpolation is exact and assertions are tight.
- 7 unit tests (CF time, axis classification, coordinate location, bad-chunking) + 10 integration tests (parameter listing, extents, bilinear values, CF packing, fill→nodata, parameter/time filters, off-grid).
- 8 new `ZarrConfig` validation tests in ds-core.
- Verified end-to-end against a running server: `/edr/collections/era5-t2m` metadata + `/position` CoverageJSON return correct values and fill→`null`.
- `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --check` clean.


mrauhala:
Thanks for the review — dispositions (pushed in 366cbdd):

**Fixed**
1. **Full time-axis read regardless of datetime filter** (`catalog.rs`) — now reads only the contiguous `[min..=max]` span of the requested time indices and shifts to local offsets. A single-step query no longer decodes the whole time axis.
2. **Dead `f.is_nan() && raw.is_nan()` branch** — removed; `raw` is already guaranteed non-NaN there. Switched to `fill_values.contains(&raw)`.
4. **`PathBuf::join` discards base on absolute `path`** — `path` is now validated at config load: absolute or `..`-containing paths are rejected (with a test).

**Deferred (intentionally, for this Phase-1 PR)**
3. **`get_available_times` clones the Vec** — the `EdrEngine` trait returns an owned `Vec<DateTime<Utc>>`, so the clone at the boundary is unavoidable regardless of an internal `Arc`; this matches `engine-grib`/`engine-querydata`. The genuinely hot accessor, `get_temporal_extent`, is already O(1).
5. **Linear scan → binary search in `locate`** — negligible at typical axis sizes (called twice per query, not per pixel); left as a follow-up to keep the diff tight and the code obviously correct.

mrauhala:
Round 2 addressed in c65b084 — all four fixed:

1. **`locate` O(n) → O(log n)** — switched to `partition_point`. Note: the suggested snippet returns `None` for a query exactly at the last coordinate (`pos == n`), so I kept both endpoints inclusive via a clamp and added `locate_inclusive_endpoints` tests.
2. **float16/bfloat16 (and other unsupported dtypes)** — variables whose dtype the read path can't widen to `f64` are now skipped at `catalog::build` with a `WARN` (the recommended Phase-1 approach), so the rest of the store still serves and there's no confusing read-time error. Added `dtype_supported`.
3. **1-D array without dimension names** — added a `debug!` log when such an array is treated as a coordinate variable.
4. **duplicate `get_temporal_extent`** — the zarr arm now computes it once and reuses it for both the log line and `has_data`.

fmt + clippy `-D warnings` + tests all green locally.

mrauhala:
Round 3 — three **new** findings fixed in 0bbac1b:

- **Zarr-native `fill_value` not honored** (catalog.rs) — the array's own fill value is now widened from its raw native-endian bytes per dtype (`fill_value_as_f64`) and added to the fill set alongside the CF attributes, so an integer array with a Zarr fill but no CF `_FillValue` no longer emits scaled garbage for unwritten chunks.
- **Axes assumed monotonic** (catalog.rs) — lat/lon are now validated with `cf::is_monotonic` and the build fails with a clear error otherwise (with a unit test).
- **NaN time value → silent epoch** (cf.rs) — `decode_times` now returns an error on any non-finite time value (with a unit test).

The other 4 inline comments on this commit are **re-cites of already-resolved items**:
- *`get_available_times` clone* — deferred intentionally; the `EdrEngine` trait returns an owned `Vec`, so the boundary clone is unavoidable and matches grib/querydata. The hot accessor `get_temporal_extent` is O(1).
- *`PathBuf::join` absolute path* — already fixed: rejected at **config load** in `config.rs` (`zarr_rejects_absolute_or_traversal_path` test), so it never reaches the join.
- *float16/bfloat16* — already fixed in c65b084: unsupported dtypes are skipped at `catalog::build` with a WARN.
- *1-D array without dim names* — already fixed in c65b084: `debug!` log added.

fmt + clippy `-D warnings` + tests (10 unit + 10 integration) green locally.

mrauhala:
Round 4 — one **new** finding fixed in 5a6db55:

- **±infinity not caught as nodata** — `convert` checked `is_nan()` only; a non-finite sample scaled to `±inf` (no JSON representation, breaks CoverageJSON). Now rejects any non-finite value. Extracted `convert_sample` as a free function with a unit test (scale/offset, fill sentinel, NaN, ±inf).

Every other inline comment on this commit is a **re-cite of an already-resolved finding** — the bot re-anchors them each round. For the record, all are addressed:
- native `fill_value`, axis monotonicity, NaN time → fixed in 0bbac1b
- float16/bfloat16 skip, 1-D dim-names debug log → fixed in c65b084
- absolute `path` → validated at config load (config.rs, with test)
- `get_available_times` clone → intentionally deferred (trait returns an owned `Vec`; matches grib/querydata; the hot accessor `get_temporal_extent` is O(1))

green locally: fmt, clippy `-D warnings`, 11 unit + 10 integration tests.

mrauhala:
Re: the single-element axis edge case (cf.rs) — addressed in 3670610.

I took the "snap to the only cell, documented + tested" option rather than a tolerance. Reason: a one-point lat/lon axis isn't necessarily a stray point — it's how a **collapsed dimension** is stored (a zonal- or meridional-mean field with a single lat or lon). For that case, returning the single value for *any* position along the collapsed axis is the correct behaviour, and an arbitrary half-degree tolerance would wrongly drop it. So I clarified the doc-comment to state the snap-to-single-cell intent and added `locate_single_element_snaps_any_target` covering near, far, and non-finite targets.

fmt + clippy `-D warnings` + tests (12 unit + 10 integration) green locally.

mrauhala:
Good call — extended `off_grid_position_is_nodata` (fa01392) to also assert all-nodata when only longitude or only latitude is out of range, not just both. Confirms the sample path requires *both* axes to locate. Green locally.

mrauhala:
On `get_parameter_descriptions` (and the related `get_available_times`) per-call allocation — **deferring intentionally**, same rationale as before:

- These are EDR **collection-metadata** accessors (called per `/collections` request), not the CLAUDE.md #211 **per-tile render** hot path that rule targets (`raster_info()` cloning timestamps on every tile).
- They match `engine-grib` and `engine-querydata`, which build the same `HashMap`/`Vec` per call.
- The `EdrEngine` trait returns **owned** values (`HashMap<String, ParameterDescription>` / `Vec<DateTime>`), so the allocation at the boundary is unavoidable without changing the trait for all engines — out of scope for this Phase-1 PR.

This covers the class of "per-call allocation in EDR metadata accessors" findings. All substantive/correctness findings from the review rounds are resolved; CI is green. I'll stop iterating here unless a new correctness issue surfaces.

PR #333, opened 2026-06-06
feat(engine-zarr): S3/HTTP backend + chunk cache — Phase 2 (#125)

mrauhala:
## Summary

**Phase 2** of #125 — remote **S3/HTTP** support for the Zarr engine, reusing `ds-storage` for every backend, plus a byte-range chunk **LRU cache**. Builds on the merged Phase 1 (#332).

## Design — and the one load-bearing decision

`zarrs` parallelises multi-chunk retrieval with **rayon by default**, and would call the storage layer from rayon workers. `ds-storage`'s `block_in_place` bridge is valid on a request/poll worker but **panics on a rayon thread** (the exact hazard CLAUDE.md warns about). So the engine pins all retrieval to the calling thread with `CodecOptions::with_concurrent_target(1)` — storage reads never land on rayon. This is documented at every relevant site as load-bearing, not tuning.

With that settled, the rest reuses ds-storage cleanly:

- **`ds-storage`** gains `get_opt`/`get_range_opt` (a missing key → `Ok(None)`, since Zarr treats an absent chunk as "use the fill value") and `list_dir` (one-level **delimiter** listing for cheap Zarr group/child discovery — no recursive walk of every chunk key on S3).
- **`engine-zarr/src/store.rs` `DsStore`** implements zarrs `ReadableStorageTraits` + `ListableStorageTraits` over `ds_storage::DataStore`, with a `quick_cache` LRU of full chunk-object bytes (byte-range requests are served by slicing the cached buffer) and a root-prefix mapping for bucket-relative stores.
- **The catalog's store type is now `DsStore` for both local and remote**, so local/S3/HTTP all flow through one ds-storage path (`build_store` / `build_s3_store_from_parts`). Phase 1's `FilesystemStore` special-case is gone.

## Config

```toml
# Local (unchanged) — data_path may also be an s3:// or http(s):// URL
[collections.zarr]

# Remote S3 (new)
[collections.zarr]
endpoint = "https://s3.us-west-2.amazonaws.com"
bucket   = "mur-sst"
path     = "zarr-v1"
parameters = ["analysed_sst"]
```

## Tests

- The **existing integration suite now exercises the entire Phase-2 storage stack** (DsStore adapter, cache, delimiter listing, byte-range slicing, `concurrent_target=1`) **network-free** — the committed fixture flows through `ds-storage`'s local backend now, not zarrs `FilesystemStore`. Verified the running server returns identical EDR values through the adapter.
- Added `DsStore` unit tests for the remote **root-prefix** mapping (`object_path`/`strip_root`), the part the local (`root=""`) path doesn't cover.
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

## Not validated in CI

A **live remote** fetch (network) isn't run in CI. The remote path differs from the (tested) local path only in the object_store backend, which `ds-storage` already provides and the GRIB engine already uses for S3. A manual smoke test against a public geographic Zarr (e.g. MUR-SST) is recommended before relying on it in production. HRRR-Zarr remains out of scope (Lambert projected grid → Phase 4).

## Deferred

Sharded huge objects are read whole (fine for the common non-sharded case; a follow-up could do true server-side range reads). Phase 3 (Map/Tiles/WMS), Phase 4 (per-item-CRS STAC), Phase 5 (kerchunk) remain.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,7 +167,7 @@ radar elevation angle. WMS exposes it as the `ELEVATION` dimension; Maps/Tiles a
 | ODIM COMP | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
 | ODIM PVOL | `EdrEngine` + `MapEngine` (per-site views) + `FeatureEngine` (network engine) | EDR (position, locations, area, trajectory), WMS, Maps, Tiles, Features (site inventory) |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
-| Zarr | `EdrEngine` | EDR (position only) — Phase 1 |
+| Zarr | `EdrEngine` | EDR (position only), local + S3/HTTP |
 | PostGIS | `EdrEngine` + `FeatureEngine` | EDR (position, locations, area), Features |
 
 ## ODIM PVOL Engine Notes (per-site collections)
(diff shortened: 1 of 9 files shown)


mrauhala:
Addressed in 27f4527:

- **Dead `get_range_opt`** — removed; `DsStore` reads whole objects via `get_opt` (cache-friendly for non-sharded chunks), so the range variant was unused. Will re-add if/when true server-side range reads land for sharded data.
- **Recursive `list_prefix`** — documented. It's recursive by contract ("every key under the prefix"); the engine's array/child discovery goes through the one-level `list_dir`, so `list_prefix` isn't on the open/read hot path. Added a note warning against calling it at the root of a large remote store.
- **"config-validated" comment** — this one's a misread: the remote-`path` requirement *is* validated, at `config.rs` `ServerConfig::validate` ("remote zarr (endpoint+bucket) requires 'path'"). It was added in #332 so it's not in this PR's diff. I reworded the comment to point at the exact check.

fmt + clippy `-D warnings` + tests green locally.

mrauhala:
Same situation as the earlier "config-validated" note — the validation **does** exist, it's just in the already-merged Phase 1 (#332), not in this PR's diff:

`crates/core/src/config.rs` `ServerConfig::validate`:
\`\`\`rust
if let Some(p) = &zarr.path {
    if std::path::Path::new(p).is_absolute() || p.split(['/', '\\']).any(|c| c == "..") {
        return Err(... "zarr 'path' must be a relative path without '..' components");
    }
}
\`\`\`
covered by the `zarr_rejects_absolute_or_traversal_path` test. So the comment's claim holds. No code change.

This and the previous note are re-cites of merged code the diff can't see; the genuine findings (dead `get_range_opt`, `list_prefix` recursion doc) were fixed in 27f4527. CI is green.

PR #334, opened 2026-06-06
feat(engine-zarr): WMS/Maps/Tiles rendering — Phase 3 (#125)

mrauhala:
## Summary

**Phase 3** of #125 — the Zarr engine now implements `MapEngine`, so a Zarr collection can be served as **WMS 1.3.0**, **OGC API - Maps**, and **OGC API - Tiles** (one layer per variable), alongside EDR. Builds on Phase 1 (#332) and Phase 2 (#333).

## How rendering works

- **`Catalog::read_window`** reads a 2-D spatial *slab* of the chosen variable at the chosen time, covering the request bbox (+1 cell margin), into memory — not the whole field, so it scales to large remote grids. It returns a `Window` that samples via bilinear interpolation over its own windowed coordinate axes using `cf::locate`, so **ascending/descending and irregular axes** all work. The window read inherits Phase 2's single-threaded retrieval (`concurrent_target=1`).
- **`get_raster_tile`** samples per output pixel for `Wgs84`/`WebMercator` (cheap `project_node`), and via a coarse **`ProjectionGrid`** for `Projected` output — never projecting per output pixel (#203).
- **`raster_info()`** is served from a cached `ArcSwap<RasterInfo>` rebuilt on each catalog swap, so the per-tile capability lookup is O(1) from a snapshot (#211).

## Wiring

- The `engine_type → supported_apis` allowlist now lists `"zarr" => ["edr", "wms", "maps", "tiles"]`.
- The zarr arm in `admin.rs` registers the engine in the map/maps/tiles registries with per-variable layer styles (`register_parameter_layer_styles`), exactly like the GRIB engine. WMS/Maps/Tiles need a `[wms]` colormap (or a `style_bundle`).

## Config

```toml
[[collections]]
id = "era5-t2m"
engine_type = "zarr"

[collections.zarr]

[collections.wms]
colormap = "temperature"
```

## Tests

- Integration tests: `raster_info` shape, a WGS84 tile asserting an **exact** linear-field pixel value, the **projected** coarse-grid path (no NaN leak, data present), and an off-grid **transparent** tile.
- Verified **end-to-end**: `GET /maps/collections/era5-t2m/map?...` returns a valid 200 `image/png`; all four APIs (EDR/WMS/Maps/Tiles) wire up on load.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` clean.

## Notes / deferred

- Map rendering reads a bbox-sized window per tile (good for large grids); the chunk LRU from Phase 2 makes repeated/adjacent tiles cheap.
- Phase 4 (per-item-CRS STAC — what HRRR's projected grid needs) and Phase 5 (kerchunk) remain.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,7 +167,7 @@ radar elevation angle. WMS exposes it as the `ELEVATION` dimension; Maps/Tiles a
 | ODIM COMP | `EdrEngine` + `MapEngine` | EDR (position, area), WMS, Maps, Tiles |
 | ODIM PVOL | `EdrEngine` + `MapEngine` (per-site views) + `FeatureEngine` (network engine) | EDR (position, locations, area, trajectory), WMS, Maps, Tiles, Features (site inventory) |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
-| Zarr | `EdrEngine` | EDR (position only), local + S3/HTTP |
+| Zarr | `EdrEngine` + `MapEngine` | EDR (position), WMS, Maps, Tiles; local + S3/HTTP |
 | PostGIS | `EdrEngine` + `FeatureEngine` | EDR (position, locations, area), Features |
 
 ## ODIM PVOL Engine Notes (per-site collections)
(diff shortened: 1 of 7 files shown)


mrauhala:
Both findings fixed in 5f2230e:

- **meta/catalog swap race** — moved `RasterInfo` into the `Catalog`, so a single `ArcSwap<Catalog>` swaps data + capabilities atomically; there's no longer any window where `raster_info()` and the served data disagree. (Dropped the separate `meta` field and its rebuild; still O(1) per #211 — built once in `catalog::build`.)
- **transparent tile between cell centres** — `axis_window` now selects every cell whose half-cell footprint overlaps the bbox (plus one bracketing cell each side), so a zoomed-in tile sitting between grid centres interpolates from the surrounding cells instead of rendering empty. Added `raster_tile_between_cell_centres_still_renders`.

fmt + clippy `-D warnings` + tests (15 unit + 15 integration) green locally.

mrauhala:
Thanks — agreed the pre-built `RasterInfo` is the right call. On the consistency note: I went with storing it *inside* `Catalog` (rather than a separate `ArcSwap<RasterInfo>` like GeoTIFF) specifically so the data and capabilities swap in one atomic `ArcSwap<Catalog>` store — that's what removes the metadata/data race flagged in the previous round. So this is an intentional, slightly stronger variant of the GeoTIFF pattern; leaving as-is. CI is green.

Issue #335, opened 2026-06-07
Icechunk support: read transactional/versioned Zarr repositories (engine-zarr)


mrauhala:
## Motivation

[Icechunk](https://icechunk.io) (Earthmover, Apache-2.0) is a **transactional, git-like-versioned storage engine for Zarr** on cloud object storage. It is rapidly becoming the *distribution* format for analysis-ready cloud datasets — e.g. the ERA5-Surface store on Arraylake is a ~60 TB Icechunk v3 repo, and more providers are publishing Icechunk rather than flat Zarr.

Our `engine-zarr` (#125) reads **plain** Zarr (flat `zarr.json` + chunk keys via the `zarrs` `ReadableStorageTraits` layer). It **cannot** read an Icechunk repository — Icechunk presents Zarr arrays/groups through its own snapshot + manifest layer, not flat object keys. Supporting it unlocks a growing class of high-quality datasets.

This is a **read-only** follow-on to the Zarr engine (#125), extending its storage layer.

## The Rust path exists

- [`icechunk`](https://crates.io/crates/icechunk) v2.x (MSRV 1.91 — fine for us).
- [`zarrs_icechunk`](https://crates.io/crates/zarrs_icechunk) provides `AsyncIcechunkStore`, a `zarrs` store over an Icechunk repository. So once we have a store, the existing `catalog::build` (which is storage-agnostic over the zarrs traits) should largely Just Work.

## Scope

- New optional backend in `engine-zarr` (read an Icechunk repo as a source), **feature-flagged off by default** (the transactional engine is a heavy dependency tree) — same posture as the kerchunk phase of #125.
- Open a repository, **checkout a read-only session at a ref** (branch HEAD, default `main`; or a tag; or an explicit snapshot id), build the store, hand it to the catalog.
- Surface the resolved snapshot id in logs / health for provenance.

## Config sketch

```toml
[[collections]]
id = "era5-surface-icechunk"
engine_type = "zarr"
apis = ["edr", "wms", "maps", "tiles"]

[zarr.icechunk]
# storage of the repo (S3/local), reusing the same fields as the plain backend
endpoint = "https://s3.us-east-1.amazonaws.com"
bucket   = "some-public-bucket"
path     = "era5-surface.icechunk"
# version selection (mutually exclusive; default = branch "main" HEAD)
branch   = "main"
# tag = "v2024.1"
# snapshot = "<snapshot-id>"
parameters = ["t2m"]
```

## Considerations / open questions

[...]

PR #336, opened 2026-06-07
feat(engine-zarr): Icechunk support (read-only, feature-gated) (#335)

mrauhala:
## Summary

Implements **#335** — reads transactional/versioned **Icechunk** repositories as a Zarr source, so the engine can serve datasets like the dynamical.org **ECMWF AIFS-single / NOAA GFS / DWD ICON-EU** forecasts (all published as anonymous Icechunk on public S3). Read-only, **off by default** behind the `icechunk` cargo feature (heavy tree: `icechunk` + `zarrs_icechunk` + `zarrs/async`).

Builds on the merged Zarr engine (#332/#333/#334).

## Design

- **`EngineStore`** (`store.rs`): a backend-agnostic `Readable`+`Listable` wrapper holding two upcast trait-object handles to the same backend. This keeps `Catalog`/`ZarrEngine` **non-generic** across the plain (`ds-storage`) and Icechunk backends. (A single `Arc<dyn ReadableListableStorageTraits>` won't work — zarrs's `child_arrays` requires `ReadableStorageTraits` **and** `ListableStorageTraits` as *separate* bounds, which the combined supertrait object doesn't satisfy; full generics would propagate into the server wiring.)
- **`icechunk.rs`**: open the repo read-only at a version (`branch` HEAD default `main`, or `tag`, or `snapshot`) → `zarrs_icechunk::AsyncIcechunkStore` (async) → `AsyncToSyncStorageAdapter` → sync read path. The adapter's `block_on` mirrors `ds-storage` (`block_in_place` in a runtime, temp runtime in tests). **Safe because retrieval stays `concurrent_target(1)`** — the same rayon×`block_in_place` invariant as the rest of the engine. Icechunk owns its own object storage, so this path deliberately does **not** use `ds-storage`.
- **Versioning ↔ reload**: opens at the configured version on load; new snapshots on a branch surface on `POST /admin/collections/reload` (v1 — poll is a no-op for a fixed session). Pinned `tag`/`snapshot` are immutable.

## Config

```toml
[zarr]
endpoint = "https://s3.us-west-2.amazonaws.com"
bucket   = "dynamical-ecmwf-aifs-single"
path     = "ecmwf-aifs-single-forecast/v0.1.0.icechunk"

[zarr.icechunk]      # presence selects the Icechunk backend
branch = "main"      # or tag = "...", or snapshot = "..."
region = "us-west-2"
```

A disabled example is committed at `collections.d/zarr-icechunk-aifs-s3.toml.disabled`.

## Tests

- **Network-free e2e** (`cargo test -p engine-zarr --features icechunk`): generates a *local* Icechunk repo (async zarrs write + commit) and reads it back through the engine — `get_parameters`, temporal extent, and an exact-value EDR position query all verified.
- Config-validation tests (multiple version selectors rejected).
- **CI guards the feature** from bitrot (clippy `-D warnings` + the e2e test under `--features icechunk`), since the default jobs don't compile it.
- Plain build/tests unchanged and green; `cargo clippy --workspace -- -D warnings` + `fmt --check` clean.

## Not covered

- A **live** fetch from a real public Icechunk repo isn't run in CI (network-free). The remote path differs from the tested local path only in the icechunk Storage backend (S3 vs local); a manual smoke test against e.g. `dynamical-ecmwf-aifs-single` is recommended (and may need S3 region/path-style tuning — `region` is configurable).
- Writing/committing (read-only engine); authenticated Arraylake catalogs; VirtualiZarr.


mrauhala:
Addressed the review (e875127):

**Fixed**
- **`forecast_time` regression** (cf.rs) — restored to `AxisRole::Time`; the forecast refactor had dropped it from the name fallback so it pinned to index 0.
- **`data_path = "s3://…"` → local-FS storage** (icechunk.rs) — now rejected with a clear error directing to `endpoint`+`bucket` for S3 repos.
- **Hardcoded `force_path_style`** — now `IcechunkConfig.force_path_style` (default `true`; set `false` for virtual-host).
- **Redundant anonymous** — dropped `with_anonymous`; `S3Credentials::Anonymous` is the single no-signing source. **Re-verified live** against AIFS/GFS/ICON-EU (all still load + read; Helsinki t2m ~13 °C).
- **Empty version selectors** (`branch = ""`) — rejected at config load, with a test.

**Considered, no change**
- *"Missing Send+Sync on `EngineStore`"* — false positive: the zarrs storage traits have `MaybeSend + MaybeSync` supertraits, so `dyn ReadableStorageTraits`/`dyn ListableStorageTraits` are `Send + Sync`. `EngineStore` is therefore `Send + Sync`, which is *required* for the server to compile (it uses `Arc<ZarrEngine> as Arc<dyn EdrEngine>`, a `Send + Sync` bound) — and it does.
- *Authenticated/private S3 repos* — intentional v1 non-goal (#335); access is anonymous (public datasets). Documented in code + the PR. A follow-up can add credentials config.

CI was green before these (Check/Lint/Test, Docker, claude-review); fmt + clippy `-D warnings` (incl. `--features icechunk`) + tests green locally.

mrauhala:
Round 2 — one new finding fixed (58823d5):

- **Reference-time calendar not checked** (catalog.rs) — the forecast reference-time decode now warns on a non-standard CF calendar, matching the plain time path.

Considered, kept as-is:
- *Temp-runtime uses multi-thread, current-thread would be lighter* (icechunk.rs) — the fallback only runs when there's no ambient runtime (tests); a **multi-thread** runtime matches `ds-storage`'s fallback and avoids subtle single-thread issues driving the aws-sdk/object_store stack. Negligible cost (rare path); leaving it.

Everything else on this commit is a **re-cite of an already-resolved item** — `forecast_time` (restored), empty selectors (rejected), `s3://` data_path (guarded), `with_anonymous` (removed), `Send+Sync` (satisfied via the traits' MaybeSend/MaybeSync supertraits), private-S3 auth (documented v1 non-goal). See the previous receipt.

CI green; fmt + clippy `-D warnings` + tests pass locally.

mrauhala:
Round 3 — one new finding fixed (e9cb07d):

- **NaN in the lead axis → silent run-time** (catalog.rs) — `NaN as i64` is 0, so a non-finite lead silently yielded valid = run time. Now rejected at build, mirroring `cf::decode_times`. Good catch.

Caveats, no change:
- *region needed for AWS signing* — access is anonymous (no signing) and `region` is configurable (all the example configs set it).
- *`path` None for icechunk remote* — the shared zarr validation already requires `path` when `endpoint`+`bucket` are set (applies to icechunk too).

The review loop has converged: across three rounds every substantive finding is fixed (forecast_time, s3:// guard, force_path_style, with_anonymous, empty selectors, reference-calendar warn, NaN-lead guard); remaining comments are re-cites of resolved items. CI green; fmt + clippy + tests (incl. `--features icechunk`) pass.

mrauhala:
Round 3 receipt (cont.) — addressed the one genuinely new finding:

- **integration.rs: `read_window` (map render) ref_axis pinning untested for forecasts** → `a270392` extends `forecast_uses_latest_run_with_lead_as_time` to also render a tile via `get_raster_tile` (the WMS/Maps/Tiles path) and assert every pixel comes from the latest run (init=1, value ≥ 1000). The forecast latest-run pinning is now regression-covered on **both** the position path (`sample_series`) and the render path (`read_window`).

The other two comments on `e9cb07d` are re-cites of already-resolved items (reference-axis calendar warn — `58823d5`; temp multi-thread runtime in the no-runtime fallback — documented, only hit in the test/no-ambient-runtime path).

Review loop converged. Merging once this run is green per the "review loops before merge" gate.

Issue #337, opened 2026-06-07
Model-run support: explicit reference-time selection in Maps and Tiles


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** EDR instances, WMS DIM_REFERENCE_TIME and Zarr forecast axes shipped. Maps/Tiles explicit run selection remains.

**Next step / remaining scope:** Add explicit run selection to Maps/Tiles and preserve run-aware cache identity; EDR instances, WMS reference time and Zarr axes already work.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Motivation

Forecast datasets have **two** time axes: the **model run** (a.k.a. *forecast reference time* / *init time* / *analysis time*) and the **valid time** (run + *lead/forecast* offset). Today the server only ever exposes a single time axis:

- **GRIB and the other forecast engines** keep multiple runs internally but expose only the **latest run**, with the temporal extent = its **valid times**. There is no way for a client to pick an earlier run.
- **The Zarr engine** (#125) is currently *inconsistent*: for a store with `init_time` + `lead_time` (e.g. dynamical.org's ECMWF AIFS, #335), `classify_axis` maps `init_time`'s `forecast_reference_time` standard-name to the time axis and pins `lead_time` to index 0 — so its temporal extent shows **run times** and it serves only **lead 0** (≈ the analysis), not a forecast. This is the bug that surfaced this issue.

We want first-class **model-run** selection, defaulting to the latest run.

## Goal / API shape

- **OGC API - EDR — instances.** Expose each run as an *instance*:
  - `GET /collections/{id}/instances` → one instance per run (`id` = run reference time, each with its own temporal extent = that run's valid times).
  - `GET /collections/{id}/instances/{instanceId}/{query}` → query a specific run.
  - `GET /collections/{id}/{query}` (no instance) → the **latest** run (today's behaviour, unchanged).
- **OGC WMS 1.3.0 — `reference_time` dimension.** Add a custom `<Dimension name="reference_time">` (the run) alongside the standard `TIME` (valid time). `GetMap` accepts `DIM_REFERENCE_TIME=<run>` (+ `TIME=<valid>`); default = latest run. (The de-facto ncWMS/THREDDS convention.)
[...]

PR #338, opened 2026-06-07
feat(edr): model-run support — EDR instances + shared run machinery (#337)

mrauhala:
Implements **Phase 1** of #337: first-class forecast **model-run** selection, defaulting to the latest run, with the shared machinery in **ds-core** so GRIB and QueryData (and later Zarr) never duplicate it (per the explicit no-duplication ask).

## The shared core — `ds_core::instances`
- `RunInfo { reference_time, valid_times }` + instance-id codec (`format_instance_id`/`parse_instance_id`, canonical compact `%Y%m%dT%H%MZ`, parse also accepts RFC 3339). The **API layer** owns the string form; engines only ever see `Option<DateTime<Utc>>`.
- `select_run(&BTreeMap<DateTime<Utc>, _>, Option<DateTime<Utc>>)` — `None` ⇒ latest, `Some(rt)` ⇒ that exact run (absent ⇒ `None` → 404). One selection rule everywhere.
- `build_instances(&runs, |rt, run| valid_times)` — `Vec<RunInfo>` from a reference-time-keyed map.

## Trait shape (Option 1, as agreed)
- `EdrEngine::get_instances() -> Vec<RunInfo>` (default empty = non-forecast).
- A trailing `reference_time: Option<DateTime<Utc>>` on `query_location/area/position/trajectory` and on `MapEngine::get_raster_tile` (`None` ⇒ latest — the unchanged default; non-forecast engines ignore it).
- `RasterInfo.reference_times: Vec<DateTime<Utc>>`.

Doing the wide signature change **once** keeps the follow-up phases purely additive (no further churn). Every non-forecast engine + the WMS/Maps/Tiles call sites accept-and-ignore (`None`).

## Engines
- **GRIB** (reference impl): exposes its catalog `runs` as instances; `resolve_time`/`query_position` honour `reference_time` (pin a run, else latest/covering-datetime); `raster_info.reference_times` carries the run list.
- **QueryData**: now retains the most recent **`max_runs`** `.sqd` files as runs keyed by **origin time** (`RunSet: BTreeMap<DateTime<Utc>, _>`), reusing already-parsed files on poll; instances + selection reuse the ds-core helpers. New `max_runs` config (default 4; `1` = latest-only). No behaviour change for the un-pinned/default path.

## api-edr
- `GET /collections/{id}/instances`, `/instances/{instanceId}` (per-run metadata: extent + data-query hrefs scoped to the run), and `/instances/{instanceId}/{position,area}` (query a specific run).
- No-instance routes default to the latest run (**unchanged**). Unknown run → **404**, unparseable id → **400**.
- Collection metadata gains an `instances` data_query and the OpenAPI spec advertises the instance paths — both gated on `get_instances()` being non-empty.

## Tests
- `ds-core` instances unit tests (codec round-trip, select/latest, build_instances).
- QueryData multi-run/instances unit test (latest default, run-pinning equivalence, bogus-run error) against the real `.sqd` fixture.
- **api-edr instances integration test** with a 2-run mock: list, per-run metadata extent, run-pinned query proves which run it hit, 404/400 cases, `instances` data_query.
- Inverted the previously-documented "instances not implemented" spec-gap test.
- **Live server smoke test** (QueryData fixture): `/instances` lists the run by origin time, `/instances/{id}/position` returns a Coverage, bogus run → 404.

All **1443** workspace tests pass; `cargo fmt` + `cargo clippy --workspace --all-targets` clean.

## Deferred (follow-ups, this PR is the foundation)
- **WMS `reference_time` dimension** + **Maps/Tiles `reference_time` query param** — the engines already honour the selector in `get_raster_tile`; the map APIs currently pass `None`.
- **Zarr** forecast-reference/lead → instances (it currently pins the latest run internally).

Part of #337 (Phase 1). #337 stays open for the WMS `reference_time` dimension, Maps/Tiles param, and Zarr instances phases.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -156,6 +156,43 @@ radar elevation angle. WMS exposes it as the `ELEVATION` dimension; Maps/Tiles a
 `elevation` query parameter; EDR as the `z` query parameter. The API layer rejects a
 `z`/`elevation` against a collection with no vertical extent (HTTP 400).
 
+### Model runs / forecast reference time (EDR instances) — #337
+
+Forecast datasets have **two** time axes: the **model run** (forecast *reference
+time* / init / analysis time) and the **valid time** (run + lead). The shared
+machinery lives in **`ds_core::instances`** so every forecast engine selects runs,
+builds instance lists, and encodes instance ids **identically** (no per-engine
+duplication):
+
+- `RunInfo { reference_time, valid_times }` — one model run; `instance_id()` =
+  canonical compact stamp `%Y%m%dT%H%MZ`.
+- `format_instance_id` / `parse_instance_id` — the URL ↔ reference-time codec
+  (parse also accepts RFC 3339); the **API layer** owns the string form, engines
+  only see `Option<DateTime<Utc>>`.
+- `select_run(&BTreeMap<DateTime<Utc>, T>, Option<DateTime<Utc>>)` — `None` ⇒
+  latest, `Some(rt)` ⇒ that exact run (absent ⇒ `None` → 404). The one selection
+  rule everywhere.
+- `build_instances(&runs, |rt, run| valid_times)` — `Vec<RunInfo>` from a
+  reference-time-keyed run map.
+
+**Engine contract:** store runs in a `BTreeMap<DateTime<Utc>, _>` keyed by
+reference time; implement `EdrEngine::get_instances() -> Vec<RunInfo>` (default
+empty = non-forecast); honour the trailing `reference_time: Option<DateTime<Utc>>`
+on the query methods and `MapEngine::get_raster_tile` (`None` ⇒ latest); populate
+`RasterInfo.reference_times`. **GRIB** (catalog `runs` map) and **QueryData**
+(`max_runs` recent `.sqd` files, keyed by origin time) implement it; other engines
+accept-and-ignore. Zarr's forecast-reference/lead detection → instances is a
+follow-up (it currently pins the latest run internally; see [[project_zarr_engine]]).
+
+**API surface:** EDR exposes `GET /collections/{id}/instances`,
+`/instances/{instanceId}` (per-run metadata), and `/instances/{instanceId}/{position,area}`
+(query a specific run); the no-instance routes default to the latest run
+(unchanged). Collection metadata gains an `instances` data_query and the OpenAPI
+spec advertises the instance paths — both gated on `get_instances()` being
+non-empty. **WMS `reference_time` dimension + Maps/Tiles `reference_time` query
+parameter are a follow-up** (the engines already honour the selector in
+`get_raster_tile`; the API layer currently passes `None`).
+
 ## Engine Capabilities
 
 | Engine | Traits | APIs |
@@ -206,11 +243,17 @@ radar elevation angle. WMS exposes it as the `ELEVATION` dimension; Maps/Tiles a
 
 - FMI QueryData (.sqd) binary format. Memory-mapped file access via `memmap2`.
 - Multi-parameter: exposes all parameters from the file. `wms_parameter` config selects which to render.
-- Polls directory for latest `.sqd` file, atomically swaps via `ArcSwap`.
+- **Model runs (#337):** polls the directory and retains the most recent `max_runs`
+  `.sqd` files as model runs, keyed by each file's **origin (analysis) time**
+  (`RunSet: BTreeMap<DateTime<Utc>, _>`), atomically swapped via `ArcSwap`.
+  Already-loaded files are reused on poll (not re-parsed). Each run is an EDR
+  instance / `RasterInfo.reference_times` entry; the latest run is the default for
+  un-pinned queries. See the "Model runs" section above and [[project_querydata_engine_plan]].
 - Supports WGS84, Stereographic, and Rotated Lat-Lon grids.
 - EDR position queries use bilinear interpolation. Map rendering uses nearest-neighbor.
 - Missing value sentinel: 32700.0.
-- Config: `wms_parameter` (name/short name/ID), `poll_interval_secs` (default 30).
+- Config: `wms_parameter` (name/short name/ID), `poll_interval_secs` (default 30),
+  `max_runs` (default 4; recent runs retained — set 1 for latest-only/no history).
 
 ## Zarr Engine Notes
 
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -980,10 +980,12 @@ pub async fn instances(
     let (engine, config) = lookup_collection(&state, &id)?;
     let base = &state.base_url;
     let runs = engine.get_instances();
-    let collections: Vec<serde_json::Value> = runs
+    let instances: Vec<serde_json::Value> = runs
         .iter()
         .map(|run| build_collection_metadata(engine.as_ref(), config, base, Some(run)))
         .collect();
+    // OGC API - EDR 1.1 §8.2.3 `instancesJSON`: the array field is `instances`
+    // (each item a collection-shaped instance), not `collections`.
     Ok(Json(json!({
         "links": [{
             "href": format!("{base}/edr/collections/{}/instances", config.id),
@@ -991,7 +993,7 @@ pub async fn instances(
             "type": "application/json",
             "title": format!("{} — instances", config.title)
         }],
-        "collections": collections,
+        "instances": instances,
     })))
 }
 
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -131,6 +131,11 @@ fn lookup_collection<'a>(
 /// and confirm the engine exposes that run (`get_instances`): an unparseable id
 /// is 400, an unknown run is 404. The validated reference time is then passed to
 /// the engine query methods.
+///
+/// The existence check is **best-effort**: it reads a catalog snapshot, and a
+/// background poll could evict that run before the subsequent query reads its
+/// own snapshot. In that rare race the query just fails (the engine has no such
+/// run) — a benign 400 instead of 404 for a run that no longer exists.
 fn resolve_instance(
     engine: &Arc<dyn EdrEngine>,
     instance_id: Option<&str>,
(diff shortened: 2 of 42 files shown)


mrauhala:
Round 1 review fixes (c34bb9f):

- **`list_sqd_files` called twice per poll** (engine.rs) → `poll_once` now lists the directory once and threads the result into `build_runset` (renamed from `load_runset`). One read per cycle.
- **Silent overwrite on duplicate `origin_time`** (engine.rs) → `build_runset` now `WARN`s which file it drops when two `.sqd` files decode to the same origin time (later-sorted wins).
- **Route param name vs OpenAPI mismatch** (lib.rs) → instance route segments renamed `{instance_id}` → `{instanceId}` to match the OpenAPI `api_definition()` path parameter. (Handlers extract positionally, so this is purely the public template name.)
- **TOCTOU between validation and query** (handlers.rs) → documented as intentional best-effort: a run evicted by a background poll between `resolve_instance` and the query yields a benign **400** (not 404) for a run that no longer exists. Not worth distinguishing engine error kinds for this rare, harmless race; left as a `///` note.

clippy `--all-targets` clean; `engine-querydata` + `api-edr` tests green (243).

mrauhala:
Round 2 review fixes (5aa7094) — addressed the deeper 404/perf points:

- **400-vs-404 for a missing run** (engine-grib `resolve_time`/`query_position`, engine-querydata `select_data`) → new `DataServerError::ReferenceTimeNotFound` mapped to **404** by the api-edr position/area error mappers. A syntactically valid but absent run is now correctly Not Found, not Bad Request.
- **`get_instances()` on hot metadata paths** (`/api`, `/collections/{id}`) cloned a `Vec<RunInfo>` (every run's valid times) per request just to test emptiness (#211) → new **`EdrEngine::has_instances() -> bool`** (O(1), overridden by GRIB/QueryData as `!runs.is_empty()`); the instances-link gating uses it.
- **TOCTOU + per-query clone** in `resolve_instance` → now that the engine returns a 404-mapping error, I **dropped the `get_instances()` pre-check** entirely. An absent run is a correct 404 even under a concurrent poll/evict, with no validate-then-query race and no per-query `Vec` clone.

The first three round-2 comments (double dir-read, dup-origin warn, `resolve_instance` TOCTOU) were re-cites of code already fixed in c34bb9f / now superseded by this commit.

clippy `--workspace --all-targets` clean; 544 tests green across ds-core/grib/querydata/api-edr.

mrauhala:
Round 3 review fixes (58cfe84):

- **OGC EDR spec field name** (handlers.rs + both tests) → the `/instances` response now uses the spec-correct array field **`instances`** (OGC API - EDR 1.1 §8.2.3 `instancesJSON`), not `collections`. A conformant client reading `body["instances"]` now sees the runs. Good catch — this was a real spec-compliance bug.
- **`has_instances` default doc** (edr_engine.rs) → clarified that delegating to `get_instances()` is fine for the default because the engines that take it are the non-forecast ones (empty `Vec`, nothing cloned); non-trivial engines override with an O(1) check (GRIB/QueryData do).

The remaining round-3 comments (double dir-read, dup-origin warn, 400-vs-404, per-`/api` `get_instances` clone) are re-cites of items already fixed in `c34bb9f` / `5aa7094` — the bot is re-anchoring them onto the latest commit but the code is fixed.

clippy clean; api-edr + ds-core tests green (228 in api-edr).

mrauhala:
Round 3 (58cfe84) review fixes — addressed all three new items (cda5df9):

- **`/instances/{id}` cloned all runs to find one** → new `EdrEngine::find_instance(rt)` (default filters `get_instances`; GRIB/QueryData override with a direct catalog lookup that builds **only** the requested run's `RunInfo`). Handler uses it — no more all-runs clone on the metadata endpoint.
- **GRIB `query_position` vs `resolve_time` divergent run-selection** → extracted **`resolve_run(catalog, reference_time, datetime)`**, used by both, so the EDR and Maps paths share one selection rule and one error mapping (`ReferenceTimeNotFound` → 404 for a pinned-absent run).
- **`list_sqd_files` swallowed `read_dir` failures** → now logs a `WARN` so a permissions regression leaves a breadcrumb instead of silently serving stale data (matches the suggested fix).

clippy `--workspace --all-targets` clean; **1443** workspace tests pass.

The earlier round-3 inline comments (double dir-read, dup-origin, 400-vs-404, per-`/api` clone, `instances` field name) were GitHub re-anchoring already-resolved items onto the latest commit.

mrauhala:
Round 4 (cda5df9) review fix — a real bug, good catch (4777be8):

**Instance query on a non-forecast collection returned 200 instead of 404.** Instance routes are registered for every collection, and non-forecast engines (GeoTIFF/ODIM/Zarr/PostGIS) ignore `reference_time` — so `/collections/{geotiff}/instances/{anything}/position` silently served latest data. This surfaced because dropping the `resolve_instance` existence check (round 3) relied on the engine returning `ReferenceTimeNotFound`, which non-forecast engines never emit.

Fix: `resolve_instance` again takes `&engine` and rejects a `Some(instanceId)` on a collection with **no instances** → 404. The guard is the **O(1) `has_instances()`** (not a `get_instances()` clone), so the round-2/3 hot-path concern still holds. Specific-run existence on a *forecast* engine remains the engine's job (`ReferenceTimeNotFound` → 404), so there's still no validate-then-query race.

Added a non-forecast mock collection to the integration test asserting an instance path on it is 404 (and the plain position query still 200s).

clippy clean; instances tests 6/6.

mrauhala:
Round (4777be8) review fixes (df1af76):

- **Bug — instance metadata 404 consistency**: the `instance` metadata handler parsed the id (→ 400 on garbage) before checking for instances, so `/collections/{non-forecast}/instances/garbage` was **400** while the query path was **404**. Added the same O(1) `has_instances()` guard up front — a collection with no instances is now **404 for any id** (parseable or not), consistent across both paths. Test extended.
- **Minor — per-run metadata rebuild** in the instances list: documented as an accepted trade-off (bounded run count, low-QPS discovery endpoint, not a #211 hot path). Threading a precomputed-metadata variant through `build_collection_metadata` (shared with `/collections` + `/collection`) would add complexity for no meaningful gain here.

The `handlers.rs:1283` comment (non-forecast `run_area_query` → 200) was already fixed by 4777be8's shared `resolve_instance` guard (both position and area go through it).

clippy clean; instances 6/6, spec `finding_20` green.

mrauhala:
Round (df1af76) review fixes (66dee06):

- **Double allocation**: `build_collection_metadata` called `run.instance_id()` twice — bound once.
- **Staleness stamp**: QueryData `poll_once` stamped `data_updated_at` on a successful `read_dir` (before parse). If every `.sqd` in the window then failed to parse, `data_age` reported "fresh" while serving stale data. Moved the stamp to after a non-empty run set is confirmed, so freshness reflects *loadable* data, not a readable directory.
- **`has_instances` "MUST override" docstring** (re-cite): this is a perf contract with no Rust enforcement mechanism for "override a defaulted trait method." Both forecast engines (GRIB/QueryData) override `has_instances` *and* `find_instance`; the default is functionally correct (just not O(1)) for the non-forecast engines that take it, whose `get_instances` returns an empty `Vec`. Left as documented.

clippy clean; api-edr + engine-querydata tests green (244).

mrauhala:
Round (66dee06) — all three were "document the deliberate choice" (8e73e36, doc-only):

- **`/instances` 200-empty for non-forecast collections**: kept the lenient 200 (standard REST list convention, matches `/collections`; item paths `/instances/{id}` still 404). OGC EDR §8.2.3 permits stricter 404 — documented why the self-consistent 200-empty is intentional (resource is never advertised for non-forecast collections, so conformant clients never reach it).
- **`build_runset` filename-sort windowing**: documented the `YYYYMMDDHHMM.sqd` naming assumption (lexical ≈ origin-time order) and why — parsing only the window keeps poll cheap vs. parsing every file in the directory to read origin times. Reissued-out-of-order caveat noted.
- **`has_instances` default**: added the inline NOTE the suggestion asked for (the contract is already in the method doc; not compile-enforceable; both forecast engines override it).

No behavior change this round. clippy clean; 438 tests green.

This thread has converged: the substantive design (shared run machinery, 404 semantics, O(1) accessors, spec field names, run-selection unification, non-forecast 404 guard) has been stable and green for several rounds, and the remaining items are documentation of already-deliberate choices. Planning to merge once CI is green here unless a genuinely new correctness finding appears.

mrauhala:
Round (8e73e36) review fixes (d348887):

- **area `f=png` ordering**: `run_area_query` resolved the instance before the static PNG-format rejection, so `/collections/{x}/instances/{id}/area?f=png` could 404 where the non-instance `area_query` 400s. Moved the request-level PNG check ahead of `resolve_instance` (kept after `lookup_collection` so collection-not-found still wins) — now consistent with the base handler.
- **instances order test**: now asserts the exact ascending order (`build_instances` contract — latest last), not just membership.
- **`has_instances` default** (4th re-cite): kept as a documented, non-compile-enforceable contract; both forecast engines override `has_instances` and `find_instance`, and the default is correct (empty `Vec`, free) for the non-forecast engines that take it. Removing the default would force ~15+ impls/mocks to add a trivial one-liner for a hypothetical future engine — disproportionate; holding unless youd prefer it compile-enforced.

clippy clean; instances tests 6/6.

mrauhala:
Verification receipt for the latest review summary (against HEAD d348887) — all four are stale re-cites or already handled:

1. **"`ReferenceTimeNotFound` missing from `map_engine_error` → 500"** — incorrect for current code. `handlers.rs:1223-1229` maps `ReferenceTimeNotFound` (alongside Location/Collection/FeatureNotFound) to **404**; added in round 2 (5aa7094). Position queries against a missing instance return 404, verified by the `unknown_instance_is_404_bad_id_is_400` integration test.
2. **area `f=png` precedes `resolve_instance`** — fixed in d348887: the PNG check now runs before `resolve_instance` (after `lookup_collection`), matching the base handler.
3. **`has_instances`/`find_instance` defaults trap** — documented, non-compile-enforceable contract; GRIB + QueryData override both; the default is correct (empty `Vec`) for the non-forecast engines that take it. Offered to compile-enforce it if preferred.
4. **instances order not asserted** — fixed in d348887: the list test now asserts exact ascending order.

The review has converged (this run produced no valid new finding against HEAD). CI all green, `mergeable=CLEAN`, 9 rounds of review with every substantive finding resolved. Merging.

PR #339, opened 2026-06-07
perf(engine-zarr): icechunk via object_store S3 backend, drop aws-sdk (~20 MB)

mrauhala:
## What

Switch the optional `icechunk` feature from Icechunk's `aws-sdk-s3` backend to its **`object_store`** backend — the same `object_store` crate `ds-storage` already uses — so the feature stops pulling the entire AWS SDK + `aws-lc-sys` C crypto build.

Motivation: review question — *"why do we need `aws-sdk-s3` when we already do S3 via `object_store`?"* Answer: we didn't; it came in transitively because `zarrs_icechunk` pulled `icechunk` with default features (incl. the `s3`/aws-sdk backend).

## Cost win (release `-p server`)

| Build | Binary | aws-sdk/aws-lc/smithy crates |
|---|---|---|
| no icechunk | 29.6 MB | 0 |
| icechunk **before** (aws-sdk) | 57.6 MB | ~44 |
| icechunk **after** (object_store) | **37.5 MB** | **0** |

The icechunk feature's footprint drops **+28 MB → +8 MB** (~70% off), and the `aws-lc-sys` C build is gone.

## Changes

- `engine-zarr` deps: `icechunk`/`zarrs_icechunk` with `default-features = false, features = ["object-store-s3", "object-store-fs"]`.
- `build_storage`: `new_s3_storage` → `new_s3_object_store_storage`. **Anonymous access is now set via `S3Options::with_anonymous(true)`** — the object_store backend keys skip-signing off `S3Options.anonymous`, *not* the `S3Credentials` arg the aws-sdk backend used. Without it, it falls through to the AWS credential chain → EC2 IMDS and hangs off-EC2 (the live probe caught this). Public datasets only (#335).
- Migrated the `probe_aifs_schema` test to the object_store constructor.
- Fixed the `--features icechunk -- --collections=…` run command in the three disabled `collections.d` examples (missing `--` separator).
- CLAUDE.md updated.

## Two interim `[patch.crates-io]` forks (both submitted upstream)

This needs one tiny fix in each of two upstream crates; both are pinned in `Cargo.lock`. **Drop each patch + bump the version once it releases:**
- **zarrs/zarrs_icechunk#13** — forward icechunk's backend features (non-breaking; `default` still enables every backend) so dependents can opt out of `s3`/aws-sdk.
- **earth-mover/icechunk#2190** — one-line `.capture()?` so `object-store-s3` compiles without the `s3` feature (a bare `?` on a `FromUtf8Error` only convertible under `s3`).

## Verification

- `cargo tree` confirms **zero** `aws-sdk`/`aws-lc`/`aws-smithy` crates with the feature on.
- Local network-free e2e (`reads_local_icechunk_repo`) passes.
- **Live anonymous-S3 probe** of all three dynamical.org datasets (AIFS/GFS/ICON-EU) opens over the object_store backend — correct params/extents, t2m@Helsinki ~16–18 °C, latest run 2026-06-07T12:00Z.
- `cargo fmt` + `clippy --workspace --all-targets` (plain **and** `--features icechunk`) clean; 1444 workspace tests pass.

Relates to #335.


mrauhala:
Review fixes (just pushed):
- **Stale comment** (icechunk.rs): corrected — no-signing is set via `S3Options::with_anonymous(true)`, not `S3Credentials::Anonymous` (that claim was left over from the aws-sdk path).
- **Patch pinning** (Cargo.toml): both `[patch.crates-io]` entries now pin an immutable `rev` instead of `branch`, so a future `cargo update` cannot follow a moved branch tip (Cargo.lock already pinned the commits; this makes the manifest match). revs: zarrs_icechunk `4ff7690`, icechunk `41f6921`.

clippy clean; icechunk e2e green.

mrauhala:
Review round 2 fixes (just pushed) — all three were constructive maintainability points:
- **Vestigial credentials arg** → pass `None` instead of `Some(S3Credentials::Anonymous)` (the object_store backend ignores it; anonymity is `with_anonymous`). Re-verified live: AIFS/GFS/ICON-EU still open anonymously with `None`.
- **`with_anonymous(true)` unconditional** → added `TODO(#335)` that it must become conditional if private-repo auth is added (so configured creds aren't silently suppressed).
- **Patch-removal tracking** → filed #340 and referenced it in the `Cargo.toml` patch comment.

clippy clean; e2e + live probe green.

mrauhala:
Round 3:
- **Docker Build failure was a transient infra flake**, not this PR: `Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded` during "Set up Docker Buildx" (Docker Hub timed out pulling the base image, before any build). Check/Lint/Test + claude-review passed. This push re-triggers it fresh.
- **probe_aifs_schema credentials inconsistency** → aligned to `None` (matches build_storage; the object_store backend ignores the arg). The other comment (unconditional `with_anonymous`) is the item already covered by the `TODO(#335)` added last round.

mrauhala:
Final round — both are observations on the deliberately-chosen interim approach, not blockers:
- **Personal-fork patch risk**: inherent to the interim `[patch]` until upstream releases. Mitigated: rev-pinned to immutable commits, icechunk-feature-only (off by default; not in the Docker image / release binaries), tracked for removal in #340, and the forks stay alive until zarrs/zarrs_icechunk#13 + earth-mover/icechunk#2190 release. Accepted tradeoff for the ~20 MB / no-AWS-SDK win.
- **TODO wording**: the `TODO(#335)` already states the failure mode (unconditional `with_anonymous` would silently suppress configured credentials); leaving as-is.

CI green, mergeable CLEAN. Merging.

Issue #340, opened 2026-06-07
Drop icechunk/zarrs_icechunk [patch] forks once upstream releases


mrauhala:
The `icechunk` feature builds against two interim `[patch.crates-io]` git forks (root `Cargo.toml`), both pinned to an immutable `rev`, to use Icechunk's `object_store` S3 backend instead of `aws-sdk-s3` (PR #339 — drops ~20 MB / the whole AWS SDK):

- **`zarrs_icechunk`** fork — forwards icechunk's backend features. Upstream: zarrs/zarrs_icechunk#13.
- **`icechunk`** fork — one-line `.capture()?` so `object-store-s3` compiles without the `s3` feature. Upstream: earth-mover/icechunk#2190.

**When each upstream PR releases to crates.io:**
1. Remove its `[patch.crates-io]` entry from the root `Cargo.toml`.
2. Bump the dependency to the released version in `crates/engine-zarr/Cargo.toml`
   (keep `default-features = false, features = ["object-store-s3", "object-store-fs"]`).
3. `cargo update -p <crate>` and re-run `cargo test -p engine-zarr --features icechunk`.

Until then these are unaudited fork builds in the transitive tree (icechunk feature only; off by default, not in the Docker image / release binaries).

mrauhala:
**Progress (PR #343):** both *personal* forks are now gone — one upstream-pinned patch remains.

**`zarrs_icechunk` — done. ✅**
zarrs/zarrs_icechunk#13 merged 2026-06-07 and shipped in **0.5.1** (2026-06-08). The `[patch.crates-io]` entry is removed; `crates/engine-zarr/Cargo.toml` now depends on the published crate (`version = "0.5.1"`, still `default-features = false, features = ["object-store-s3", "object-store-fs"]`). Verified 0.5.1 exposes the forwarding features before relying on it.

**`icechunk` — patch still required, but no longer a personal fork.**
earth-mover/icechunk#2190 merged 2026-06-07 but is **not yet in a crates.io release** — the latest published version is `2.0.6` (cut 2026-06-04, before the merge). So the patch stays, but it's been **repointed from `mrauhala/icechunk` to the upstream `earth-mover/icechunk` merge commit** (immutable `rev = 5f56b13`). Side effect: the newer rev raised its floors, so `Cargo.lock` bumps `quick_cache 0.6.21→0.6.23` and `serde_json 1.0.149→1.0.150` (both forced by the rev).

[...]

PR #341, opened 2026-06-07
build(docker/release): ship icechunk by default (object_store backend, ~+8 MB)

mrauhala:
## What

Now that the `icechunk` feature is cheap (#339 — object_store S3 backend, no AWS SDK, +8 MB instead of +28 MB), enable it in the shipped artifacts.

- **Dockerfile:** `ARG CARGO_FEATURES="icechunk"`, threaded through both the cargo-chef `cook` and the final `build` (cargo-chef needs the same features on both so the cooked dep cache matches). **Default-on**; produce a lean image with `--build-arg CARGO_FEATURES=""` (or pass a custom feature set).
- **release.yml:** add `--features icechunk` to the release-binary build. Reuses the `object_store`/rustls stack already in the base build, so it adds **no new cross-compilation deps**.

## Verification

Local `docker build` (default) → 156 MB image. Running it with the committed `testdata/zarr-icechunk-aifs.config.toml` (anonymous AWS S3, from inside the container) loads the repo end-to-end:

```
engine_zarr::catalog: collection 'ecmwf-aifs-single': forecast — latest run 2026-06-07 12:00 UTC (of 3191 runs)...
engine_zarr: [ecmwf-aifs-single] Loaded Zarr store: 1 variable(s) [temperature_2m], 61 time step(s)
server::admin: Collection 'ecmwf-aifs-single': wired to EDR API
```

CI's "Docker Build (PR)" job re-validates the default build.

## Caveat

Until the two icechunk `[patch]` forks land upstream (tracked in #340), the default image/release build pulls them. Drop `--features icechunk` / the ARG default if a fork-free default artifact is required before then.

Relates to #335, #339, #340.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -218,7 +218,12 @@ jobs:
         # producing a release built against a drifted lockfile.
         env:
           TARGET: ${{ matrix.target }}
-        run: cargo build --release --locked --target "${TARGET}" -p server
+        # `--features icechunk`: ship Icechunk Zarr support in release binaries
+        # (object_store S3 backend, ~+8 MB, no AWS SDK — #339). Reuses the
+        # object_store/rustls stack already in the base build, so it adds no new
+        # cross-compilation deps. Remove once the icechunk [patch] forks are
+        # dropped if a leaner default is preferred (#340).
+        run: cargo build --release --locked --target "${TARGET}" -p server --features icechunk
 
       - name: Package tarball
         # All inputs flow through `env:` so the shell never sees an
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,10 @@ RUN cargo install cargo-chef --locked --version 0.1.71
 # stable across source-only changes — the next stage's cache hits.
 FROM chef AS planner
 COPY . .
+# NB: `cargo chef prepare` (0.1.71) takes no `--features` — the recipe captures
+# the full manifest incl. the *optional* `icechunk` dep, so the builder's
+# `cook --features icechunk` pre-warms those crates from it. (Newer cargo-chef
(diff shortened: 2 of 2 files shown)


mrauhala:
Review fix (just pushed): added `--locked` to both the cargo-chef `cook` and the final `cargo build` in the Dockerfile — matches the release build and pins the image to the committed `Cargo.lock` (important now that the default feature set pulls the rev-pinned `[patch]` git forks; no silent re-resolution). Verified `cargo check --locked -p server --features icechunk` succeeds (lock in sync). CI Docker Build (PR) re-validates the full image build.

mrauhala:
Re: the `cargo chef prepare --features` suggestion — verified it would **break the build**: our pinned cargo-chef **0.1.71** does not accept `--features` on `prepare`:
```
Usage: cargo chef prepare <--recipe-path <RECIPE_PATH>|--bin <BIN>>
```
(caught via a targeted `docker build --target planner`). The flag exists in newer cargo-chef, but the docs note applies to those versions.

It is also unnecessary here: `cargo chef prepare` captures the full workspace manifest **including the optional `icechunk` dependency declaration**, so the builder's `cook --features icechunk` enables and pre-warms that subtree from the recipe. Empirically the local + CI image builds produced a working icechunk binary (it opened the AIFS repo over S3 in-container). I added a comment documenting this rather than the (unsupported) flag. `cook`/`build` keep `--features` + `--locked`.

mrauhala:
The `prepare --features` comment is a re-cite of the item already verified incorrect for cargo-chef 0.1.71 (it rejects the flag — usage error shown above) and unnecessary (recipe captures the optional icechunk dep). CI green, mergeable CLEAN — merging.

PR #345, opened 2026-06-08
feat(wms): reference_time dimension for forecast model runs (#337)

mrauhala:
## What

Phase 2 of #337 (model-run support). WMS now exposes the forecast **model run** (forecast *reference time* / init / analysis time) as a custom `reference_time` dimension, alongside the standard `TIME` (valid-time) axis — the de-facto ncWMS/THREDDS convention. Default = latest run, matching the existing no-instance EDR behaviour.

## Changes

**`api-wms`**
- **GetCapabilities** advertises `<Dimension name="reference_time" units="ISO8601">` for any layer whose engine reports a non-empty `RasterInfo.reference_times` (GRIB, QueryData today), with `default=` the latest run and every run listed as RFC 3339. Placed among the existing `<Dimension>`s, before `<Attribution>` (WMS 1.3.0 schema order). **No `nearestValue`** — the run must match an advertised value exactly.
- **GetMap** parses `DIM_REFERENCE_TIME=<run>` (RFC 3339 / lenient ISO, or the compact EDR instance-id stamp `20260607T0600Z`). The handler validates it against the advertised runs and returns `InvalidDimensionValue` (HTTP 400) for:
  - an unknown run, and
  - a `DIM_REFERENCE_TIME` against a non-forecast layer.
  
  This is the correct WMS response — without it the engine's `ReferenceTimeNotFound` would render a red 200 tile. Mirrors the existing parameter/`ELEVATION` validation.
- The selected run flows through `get_raster_tile` (both the direct and Web-Mercator meta-tile paths).

**`ds-render`**
- `CacheKey.reference_time` and `TileKeyPrefix`/`TileKey.reference_time` so distinct runs don't collide in the rendered-image and meta-tile caches. ETag golden re-pinned (cache-bust event).

**`api-maps` / `api-tiles`**
- Pass `reference_time: None` for the new cache-key field. Their `reference_time` query parameter is the remaining **Phase 4** follow-up; the engines already honour the selector.

## Tests

`crates/api-wms/tests/integration.rs` — a two-run forecast mock engine that records the `reference_time` it's asked to render:
- capabilities emit the `reference_time` dimension (default = latest, both runs listed, no `nearestValue`) and omit it for non-forecast layers;
- GetMap default → engine queried with `None` (latest); explicit run (RFC 3339 **and** compact instance-id) → engine queried with that run;
- unknown run / unparseable value / run on a non-forecast layer → 400 `InvalidDimensionValue` (and the engine is not called for an invalid run).

Plus a `ds-render` test that the ETag changes with `reference_time`.

`cargo test --workspace` green; `cargo clippy --workspace --tests` clean; `cargo fmt` applied.

## Remaining for #337

- [ ] OGC API - Maps/Tiles `reference_time` query parameter (Phase 4)
- [ ] Zarr forecast-reference/lead → instances (Phase 3)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -189,9 +189,19 @@ follow-up (it currently pins the latest run internally; see [[project_zarr_engin
 (query a specific run); the no-instance routes default to the latest run
 (unchanged). Collection metadata gains an `instances` data_query and the OpenAPI
 spec advertises the instance paths — both gated on `get_instances()` being
-non-empty. **WMS `reference_time` dimension + Maps/Tiles `reference_time` query
-parameter are a follow-up** (the engines already honour the selector in
-`get_raster_tile`; the API layer currently passes `None`).
+non-empty.
+
+**WMS `reference_time` dimension (done).** Forecast layers (non-empty
+`RasterInfo.reference_times`) advertise a custom `<Dimension name="reference_time">`
+in `GetCapabilities` alongside the standard `time` dimension (the valid-time axis),
+defaulting to the latest run. `GetMap` accepts `DIM_REFERENCE_TIME=<run>`
+(RFC 3339, or the compact instance-id stamp); the handler validates it against the
+advertised runs and returns `InvalidDimensionValue` (HTTP 400) for an unknown run
+or a non-forecast layer (no `nearestValue` — the engine requires an exact match).
+The run flows through `get_raster_tile` and into the rendered + meta-tile cache
+keys (`CacheKey.reference_time`, `TileKeyPrefix.reference_time`) so distinct runs
+don't collide. **Maps/Tiles `reference_time` query parameter is still a follow-up**
+(api-maps/api-tiles pass `reference_time: None` today).
 
 ## Engine Capabilities
 
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -1066,6 +1066,9 @@ async fn render_map(
         time,
         parameter: effective_parameter.clone(),
         z: validated.z.map(ds_render::quantize_z),
+        // Maps `reference_time` query parameter is a follow-up (#337 Phase 4);
+        // the handler always queries the engine's latest run for now.
+        reference_time: None,
     };
 
     let cache_control = cache_control_value(has_explicit_time);
--- a/crates/api-tiles/src/handlers.rs
+++ b/crates/api-tiles/src/handlers.rs
@@ -1430,6 +1430,9 @@ async fn render_tile(
         time,
         parameter: effective_parameter.clone(),
         z: validated.z.map(ds_render::quantize_z),
+        // Tiles `reference_time` query parameter is a follow-up (#337 Phase 4);
+        // the handler always queries the engine's latest run for now.
+        reference_time: None,
     };
 
     let cache_control = cache_control_value(has_explicit_time);
--- a/crates/api-wms/src/capabilities.rs
+++ b/crates/api-wms/src/capabilities.rs
@@ -273,6 +273,32 @@ fn write_layer_metadata(writer: &mut Writer<Vec<u8>>, info: &RasterInfo) {
             let _ = writer.write_event(Event::End(BytesEnd::new("Dimension")));
         }
     }
+
+    // Forecast reference-time (model run) dimension — advertised only for
+    // forecast collections that retain multiple runs. The standard `time`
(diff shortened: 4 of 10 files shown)


mrauhala:
Thanks for the review. Receipts:

**#2 — `STYLES=` missing (fixed).** Added the mandatory-but-may-be-empty `STYLES=` to `FC_GETMAP_URI` per WMS 1.3.0 §7.3.3 (be8f1cd). Tests stay green.

**#1 — dead `if let Some` / sort contract (keeping as-is, with rationale).** Two parts:
- *Sort contract:* already documented — `RasterInfo.reference_times` carries `/// Available forecast model runs (reference times), ascending.` (ds-core `map_engine.rs`). So "latest last" is part of the field's contract, and `.last()` is the intended way to read the default run.
- *Dead branch:* the `if !empty { … if let Some(latest) = …last() { default= } }` shape is a deliberate copy of the two sibling dimension blocks immediately above it in the same function (`time` and `elevation`). Rewriting only the `reference_time` block to `.expect()` would make it the odd one out. If we want the `.expect()` form, it's worth doing to all three as a small consistency pass — out of scope for this PR, happy to file a follow-up.

mrauhala:
Round 2 receipts (f1d0461):

**Three `raster_info()` snapshots → one (fixed).** Hoisted the parameter/ELEVATION/reference_time validation onto a single `engine.raster_info()` snapshot. `raster_info()` clones its vecs, so per the CLAUDE.md hot-path discipline taking it once is the right call — good catch (I added the third).

**Redundant `dim_reference_time` alias (fixed).** Removed it; serde already matches the lowercase field name. Kept `DIM_REFERENCE_TIME` / `Dim_Reference_Time` for the case variants real clients send.

**`capabilities.rs` `if let Some(latest)` (keeping, re-affirming).** The vec's ascending-sort contract is documented on `RasterInfo.reference_times` (`/// …, ascending.`), and this `if !empty { … if let Some(latest) = …last() }` shape is a deliberate copy of the two sibling dimension blocks (`time`, `elevation`) directly above it in the same function. Converting only this one to `.expect()` makes it the odd block out; converting all three is a reasonable standalone consistency pass but is out of scope here.

mrauhala:
Round 3 receipt (bd6d455):

**Ascending-sort invariant documentation (strengthened).** Took the suggested doc route: `RasterInfo.reference_times` now spells out the contract explicitly — *"sorted ascending, so the latest run is `.last()`; engines build this from a reference-time-keyed `BTreeMap`, and consumers depend on the ordering (WMS advertises `.last()` as the dimension's `default`)."* That makes the invariant the WMS default rests on loud at the type, which was the substantive concern.

On the `if let Some(latest)` guard itself: keeping it for consistency with the two sibling dimension blocks (`time`, `elevation`) in the same function, which use the identical `if !empty { … if let Some(latest) = …last() }` shape. A pass to convert all three to `.expect()` is a fine standalone cleanup but is out of scope for this PR.

mrauhala:
Round 4 receipts (36f8172):

**Meta-tile path untested (fixed).** Added `getmap_metatile_path_selects_pinned_reference_time` — a `CRS=EPSG:3857` GetMap pinning a non-latest run, asserting the run reaches *every* `render_metatiled` `get_raster_tile` call. That covers both `TileKeyPrefix.reference_time` and the meta closure, which the CRS:84 direct-path tests didn't.

**Latest-run pin fragments the cache (fixed).** Normalised an explicit pin of the *current* latest run to `None` at the cache-key/engine-call site, so it shares entries with the no-dimension request (identical pixels). This matches the common client flow (echoing the capabilities `default=`, which is the latest run) and the repo's "cache key must match the access pattern" rule. A pin of an older run stays explicit. Added `getmap_explicit_latest_run_normalizes_to_none` to lock it in.

mrauhala:
Round 5 — verdict is "no blocking issues"; both new notes are non-actionable for this PR:

- **Case-insensitive param names (§6.8):** as noted, this is a codebase-wide WMS convention — every parameter (`TIME`/`ELEVATION`/`BBOX`/…) uses the same fixed upper/title/lower alias set, not a true case-insensitive layer. `DIM_REFERENCE_TIME` follows that established pattern exactly. A strict case-insensitive pass would be a separate cross-cutting change.
- **O(n) `contains` over runs:** negligible for the 4–40 runs engines retain. The vec is documented ascending, so `partition_point` is a clean drop-in if a very-large archive ever makes it matter — noted for a follow-up, not needed here.

No code change for these. Thanks for the thorough multi-round review — the PR is ready from my side.

Issue #348, opened 2026-06-08
3D Tiles Phase 2: ds-core VolumeEngine trait + VoxelGrid type + ds-3dtiles encoder crate


mrauhala:
Part of #346.

Promote the PoC's hand-rolled encoding into the codebase's normal shape: a domain type + trait in `ds-core`, and a framework-free encoder crate.

## Scope
- **ds-core**: `VoxelGrid` domain type (dims, cylinder/region bounds, per-quantity arrays + noData) and a **`VolumeEngine`** trait — `read_voxel_grid(bounds, lod) -> VoxelGrid` plus tileset/subtree descriptors. Framework-free (alongside `EdrEngine`/`FeatureEngine`/`MapEngine`).
- **ds-3dtiles** (new crate, like `ds-mvt`): encode `VoxelGrid` → `.pnts` (and later `.glb`), `tileset.json`, ECEF/ENU transforms, `region` bounding volume. No framework deps.
- **engine-odim**: `PolarVolumeSiteView` implements `VolumeEngine`, generalizing the PoC sampler from a 1-D cross-section path to a 3-D `(radius, azimuth, height)` lattice; honour the surveyed-elevation envelope (no fabricated data in the cone of silence).

## Notes
- Encode the gotchas from #346 as invariants (geometricError > 0; pnts ECEF-native; etc.).
- Multi-quantity: carry DBZH/VRADH/ZDR/RHOHV together as parallel arrays.

## Done when
- `engine-odim` produces the same tileset the PoC does, but via `VolumeEngine` + `ds-3dtiles` (PoC example becomes a thin driver).
- Unit/integration test over a fixture (per repo convention).

mrauhala:
Encoder hardening to fold in when building `ds-3dtiles` (surfaced by claude-review on the Phase-1 demo PR #352):

- **Read actual per-ray azimuths**, don't assume a uniform 360° sweep. The demo uses `deg_per_ray = 360 / nrays`, which is wrong for **sector scans** and volumes with dropped/duplicate rays. ODIM stores real azimuths in `how/startazA`/`stopazA` (or `startazT`/`stopazT`); `volume_engine.rs`'s trajectory query already reads these. A production encoder placing points at assumed azimuths is a silent geometry error.
[...]

Issue #346, opened 2026-06-08
Epic: OGC 3D Tiles for volumetric weather (radar polar volumes first)


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep as an epic; prioritize remaining children.

**Current assessment and challenge:** 3D Tiles points, meshes, voxels and viewer exist. Refresh phases and rank concrete remaining children.

The original roadmap below is historical context; shipped phases should not be counted as new work or as independent priority candidates. Remaining child issues retain their individual implementation and verification scope.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

OGC 3D Tiles (OGC 22-025r4) delivery of **volumetric** weather data, starting with ODIM radar **polar volumes**. Renders true 3D structure — storm cores, echo tops, the bright band, the cone of silence — that 2D WMS/Maps/Tiles output fundamentally cannot show.

A Phase-1 proof of concept already renders the **fivih (Vihti)** polar volume in CesiumJS (see *PoC findings* below).

## Why radar fits 3D Tiles unusually well

The emerging 3D Tiles **voxel** work defines a **cylinder** voxel shape whose grid dimensions are `[radius, angle, height]` — i.e. radar's native `(range, azimuth, elevation)`. The azimuth axis maps **1:1** (radar rays *are* the angle axis); only radius/height need resampling, via the engine's existing 4/3-Earth beam model (`slant_to_ground_height` in `engine-odim/src/volume_engine.rs`). The engine already carries the hard pieces: beam geometry, per-site catalog `ArcSwap` snapshot, lazy pixel cache, multi-quantity metadata, and the cross-section sampler that already inverts `(ground, height) → (slant, elangle)`.

## Two delivery tiers

**Tier A — standardized today (3D Tiles 1.1, OGC-ratified):** glTF 2.0 content + implicit `OCTREE` tiling + `EXT_structural_metadata`. Ship radar as point clouds / isosurfaces / CAPPIs that render in *any* 3D Tiles client now.

[...]

PR #352, opened 2026-06-08
feat(engine-odim): 3D Tiles point-cloud demo for radar polar volumes

### LINKED ISSUE #347: 3D Tiles Phase 1: static .pnts exporter for radar polar volumes (CesiumJS demo)
Part of #346.

Formalize the Phase-1 proof of concept: a reproducible static **`.pnts`** exporter that turns an ODIM polar volume into an OGC 3D Tiles point cloud loadable in CesiumJS. This is a standalone example/CLI (no engine trait changes yet) — it validates the geometry pipeline end-to-end before `ds-3dtiles` is built.

## Scope
- `crates/engine-odim/examples/gen_3dtiles.rs` — reads a PVOL, projects every echo cell via the 4/3-Earth beam model → geodetic → ECEF, writes `content.pnts` + `tileset.json` + a token-free **CARTO Dark Matter** viewer.
- `crates/engine-odim/examples/volume_stats.rs` — reflectivity probe to pick a good demo timestep.
- A short demo README documenting how to run it and the `.pnts`/`geometricError` gotchas (see epic).

## Done when
- `cargo run -p engine-odim --example gen_3dtiles -- <pvol.h5>` produces a tileset that renders in CesiumJS.
- Verified visually (377k points, fivih 2026-05-19 10:50Z).
- Committed + reproducible from a checked-in fixture.

Out of scope: serving over HTTP (#api-3dtiles), the `VolumeEngine` trait, time animation, true voxels.

mrauhala:
Phase 1 of the **OGC 3D Tiles for volumetric weather** epic (#346). Closes #347.

Standalone `cargo run` examples that turn an ODIM **polar volume** into an OGC
3D Tiles `.pnts` point cloud loadable in CesiumJS — validating the
`(slant_range, elevation, azimuth) → geodetic → ECEF` geometry pipeline before
the `VolumeEngine` trait + `ds-3dtiles` encoder (#348) are built. No engine
trait changes; nothing wired into the server.

## What's here
- **`examples/gen_3dtiles.rs`** — reads a PVOL, projects every echo cell via the
  engine's 4/3-Earth beam model → ECEF, writes `content.pnts` + `tileset.json` +
  a token-free **CARTO Dark Matter** CesiumJS viewer. Colours by an NWS dBZ ramp.
- **`examples/volume_stats.rs`** — per-sweep/volume reflectivity probe (coverage,
  max dBZ, convective fraction, echo top) for picking a demo timestep.
- **`examples/README.md`** — usage + the gotchas below.

## Try it
```bash
cargo run -p engine-odim --example gen_3dtiles            # default FMI Vihti fixture
cd target/3dtiles-fivih && python3 -m http.server 8777    # open localhost:8777
```

Renders the fivih (Vihti) 2026-05-19 10:50Z volume — 377k points: the radial
sweep fans, echo tops, convective cells, and the cone of silence above the
antenna, all against a dark basemap.

## Gotchas captured (these will inform `ds-3dtiles`, #348)
- **Use `.pnts`, not glb-with-POINTS** — a glb POINTS primitive goes through
  Cesium's `Model` renderer, draws fixed 1px points and **ignores both
  `pointSize` styling and vertex `COLOR_0`** (faint white). `.pnts` honours both.
- **Tileset top-level `geometricError` must be `> 0`** — at `0`, Cesium never
  refines to the root and **never requests the content tile** (nothing renders).
- **`.pnts` `POSITION` is ECEF-native** (offset from `RTC_CENTER`) — no glTF
  Y-up→Z-up flip.
- `region` bounding volume is geodetic and ignores the tile transform.

## Notes
- Default input is the FMI Vihti volume the integration tests already use; like
  those tests it is **not committed** (15 MB) and the example exits gracefully
  when absent. CI is unaffected (examples build under clippy but aren't run).
- `cargo fmt` + `cargo clippy -p engine-odim --examples` clean.

### DIFF
--- a/crates/engine-odim/examples/README.md
+++ b/crates/engine-odim/examples/README.md
@@ -1,0 +1,84 @@
+# engine-odim examples — 3D Tiles radar-volume demo
+
+Proof-of-concept tooling for **OGC 3D Tiles** delivery of radar polar volumes
+(epic [#346], Phase 1 = [#347]). These are standalone `cargo run` examples — they
+do **not** change any engine trait; they validate the geometry pipeline
+end-to-end before the `VolumeEngine` trait + `ds-3dtiles` encoder are built
+([#348]).
+
+## `gen_3dtiles` — polar volume → 3D Tiles point cloud
+
+Reads an ODIM polar volume, projects **every** echo cell through the engine's
+4/3-Earth beam model → geodetic → ECEF, and writes a 3D Tiles point cloud plus a
+token-free CesiumJS viewer.
+
+```bash
+# Uses the default FMI Vihti fixture (see "Fixture" below):
+cargo run -p engine-odim --example gen_3dtiles
+
+# Or point it at any ODIM polar volume, with an output dir + min dBZ threshold:
+cargo run -p engine-odim --example gen_3dtiles -- path/to/volume.h5 target/3dtiles 5.0
+```
+
+Output (default `target/3dtiles-fivih/`):
+
+| File | What |
+|------|------|
+| `content.pnts` | the point cloud (3D Tiles 1.0 Point Cloud format) |
+| `tileset.json` | the tileset (single tile, `region` bounding volume) |
+| `index.html`   | a self-contained CesiumJS viewer (CARTO Dark Matter basemap) |
+
+### Viewing
+
+CesiumJS needs the files served over HTTP (not `file://`):
+
+```bash
+cd target/3dtiles-fivih && python3 -m http.server 8777
+# then open http://localhost:8777/index.html
+```
+
+The colours are an NWS-style reflectivity ramp; the dark basemap makes them pop.
+You'll see the conical sweep structure, the echo tops, and the "cone of silence"
+directly above the antenna.
+
+## `volume_stats` — reflectivity probe
+
+Prints per-sweep + whole-volume reflectivity stats (coverage, max dBZ, convective
+fraction, approximate echo top). Used to pick a demo timestep with interesting
+weather.
+
+```bash
+cargo run -p engine-odim --example volume_stats -- path/to/volume.h5 [QUANTITY]
+```
+
+## Why `.pnts` and not `.glb` (and other gotchas)
+
+Hard-won while building the PoC — these inform how `ds-3dtiles` should encode:
+
+- **Use `.pnts`, not a `.glb` with POINTS-mode.** A glb point primitive routes
+  through CesiumJS's `Model` renderer, which draws fixed **1px** points and
+  **ignores both `pointSize` styling and vertex `COLOR_0`** (renders faint
+  white). The `.pnts` FeatureTable (POSITION + RGB) honours `pointSize` and
+  per-point colour.
+- **The tileset's top-level `geometricError` must be `> 0`.** With it `0`,
+  CesiumJS never refines to the root and **never even requests the content
+  tile** — nothing renders, no error. (This example uses tileset `100000`,
+  root `1000`.)
+- **`.pnts` `POSITION` is ECEF-native** (offsets from `RTC_CENTER`) — no glTF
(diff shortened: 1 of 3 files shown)


mrauhala:
Addressed the claude-review findings in 14476d7:

1. **Bug — silently-dropped `index.html` write** → now `.expect("write index.html")`, consistent with the other writes.
2. **Misleading hardcoded title/HUD** → `write_viewer` now takes a `hud` string built from the actual volume (site label + `vol.time` + dBZ threshold); verified the generated page reads `Vihti (fivih) polar volume — 2026-05-19 10:50Z · DBZH ≥ 5 dBZ` and updates for other inputs.
3. **Dead `h` param** → camera altitude is now `site.height + 90 km` (tracks the antenna rather than assuming sea level).
4. **`assert!` panic** → replaced with `eprintln!` + `process::exit(1)`, matching the file-read path.

`cargo fmt` + `cargo clippy -p engine-odim --examples` clean; regenerated and re-verified the demo renders in CesiumJS.

mrauhala:
Fixed the HTML-injection finding in 988e735: added `html_escape()` in `write_viewer` and applied it to the HUD before interpolation, so HDF5-sourced `PLC`/`NOD` can no longer inject markup into the `<title>`/`<div>`. (This vector was introduced by the previous commit's dynamic HUD — good catch.) `fmt`/`clippy` clean; normal output unchanged.

mrauhala:
Addressed the two hardening items from the latest review in 614ce7d:

1. **Silent u32 truncation in the .pnts header** → `u32::try_from(...).expect(...)` for all three length fields (loud failure instead of a corrupt >4 GB tile).
2. **Non-finite coords → broken JS / NaN positions** → guard `site.lon/lat/height.is_finite()` right after parse with a graceful `eprintln!` + `exit(1)`. Placed early so it protects both the ECEF projection and the viewer (its `lon`/`lat`/`cam_alt` derive from these), rather than only the `write_viewer` assert.

Intentionally **skipping** the minor `FOUR_THIRDS_EARTH_M` duplication — the two examples are standalone and the reviewer confirmed no correctness bug; a shared module just to dedupe one const isn't worth it for demo tooling.

`fmt`/`clippy` clean; regenerated and re-verified.

mrauhala:
Fixed both items from the latest review in 665f5c3 (the 'approve pending' set):

1. **NaN coverage in volume_stats** → guard `vtot == 0`: prints `no readable {QUANTITY}/TH sweeps` instead of `-inf` max + `0/0 = NaN` coverage.
2. **Backwards lon-range comment** → corrected to `[-180, 180)` (the formula maps +180 → -180).

`fmt`/`clippy` clean; both examples re-verified.

mrauhala:
Fixed both round-5 items in c9996d8 (the two the review listed as the only remaining ones):

1. **`-inf` max when sweeps exist but all cells are nodata** → guard now keys on `vvalid == 0` rather than `vtot == 0`, so that case prints `no valid …/TH samples (N cells, all nodata/undetect)` and returns.
2. **Header/count mismatch** → table headers now read `>=20%/>=35%/>=50%` to match the inclusive `>= 20.0` counts.

`fmt`/`clippy` clean; re-verified header/row alignment.

mrauhala:
Round 6 disposition:

- **Fixed here** (9d5bc53): the `>=20%` per-sweep column now has a matching volume-summary total (`precip (>=20)`).
- **Deferred to #348** (per the review's own "carry forward before building ds-3dtiles" framing): (1) read actual per-ray azimuths (`how/startazA`) instead of assuming a uniform 360° sweep — matters for sector scans; (2) finite-guard sweep-level metadata so a corrupt `.h5` can't emit `NaN`/`Infinity` into `tileset.json`. Both logged at https://github.com/mrauhala/meteocore/issues/348#issuecomment-4648622005.

The demo's geometry/encoding has been signed off across rounds; these two are forward-looking encoder concerns, not demo blockers. Considering this PR complete.

Issue #351, opened 2026-06-08
3D Tiles voxels: octree LOD and national mosaic after single-tile support


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** Single-tile cylindrical voxels shipped; octree LOD and mosaic are the remaining expansion, not absent voxel support.

**Next step / remaining scope:** Keep this open for octree LOD and mosaic requirements. Build on the existing single-tile cylindrical voxel implementation.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

Part of #346. The **Tier-B** high-fidelity track (draft spec, CesiumJS-only today).

Move from point clouds to **true volumetric voxels** + LOD + multi-radar mosaic.

## Scope
- **Cylindrical voxels**: encode `VoxelGrid` as glTF with `EXT_primitive_voxels` (cylinder shape, dims `[radius, angle, height]`) + `3DTILES_content_voxels` + `3DTILES_bounding_volume_cylinder`. The azimuth axis maps 1:1 from radar rays; radius/height via 4/3-Earth. Quantities → `EXT_structural_metadata` SCALAR properties with `noData`.
- **Implicit OCTREE tiling** + `.subtree` availability for sparse coverage.
- **National mosaic**: stitch N sites into one octree with `geometricError`-driven LOD; honour per-site coverage envelopes.
- Client-side isosurfaces / CAPPI slices / threshold (the payoff over point clouds).

## Risks / notes
- Draft extensions, **CesiumJS-only**, spec in flux (axis-reorder rule `[radius,angle,height]`↔`[radius,height,angle]` between the two extensions). Keep Tier-A `.pnts`/mesh as the always-works fallback.
- Bandwidth: quantize to UINT8/16, exploit noData sparsity, octree LOD for the mosaic.

## Done when
- A single site renders as voxels in a recent CesiumJS build.
- (Stretch) a multi-site mosaic with octree LOD.

---

## Status update — CesiumJS voxel support as of 1.142 (2026-06-09)

Checked the current state of the voxel path (was "≥1.127" in our notes; CesiumJS is now **1.142**, 2026-06-01). It has **matured and settled** — a burst of breaking refinements through 2025, then quiet:

| Version | Voxel change (relevant to this issue) |
|---|---|
[...]

Issue #349, opened 2026-06-08
3D Tiles Phase 3: api-3dtiles HTTP service (tileset.json + content + subtrees)


mrauhala:
Part of #346. Depends on the Phase-2 `VolumeEngine` + `ds-3dtiles`.

New HTTP API crate serving 3D Tiles from any `VolumeEngine`-capable collection.

## Scope
- **api-3dtiles** (new crate): routes
  - `GET /3dtiles/{collection}/tileset.json`
  - `GET /3dtiles/{collection}/content/{...}` (`.pnts`/`.glb`)
  - `GET /3dtiles/{collection}/subtrees/{...}` (once implicit tiling lands)
- Depends only on ds-core + ds-3dtiles (no engine crates). Engine registry keyed by collection id, same as the other APIs.
- Wire `"3dtiles"` into the `engine_type → supported_apis` allowlist in `server/src/admin.rs`; gate on the collection's `apis` config.
- ETag/caching aligned with the catalog `data_version`.

## Done when
- A configured `odim-volume` collection serves a working tileset over HTTP that loads in CesiumJS by URL.
- OpenAPI/landing wiring updated.

mrauhala:
Handler-layer requirements surfaced by the #348 review (PR #353) — must be honoured when wiring the route:

- **Run `VolumeEngine::read_point_cloud` off the request-serving runtime.** It's a sync trait method (like the other engine traits) that does HDF5 I/O (`blocking_pixel_handle`) + a long CPU loop over all cells. Per CLAUDE.md's concurrency rules, dispatch it via `spawn_blocking` on the dedicated runtime (the same pattern `api-{wms,maps,tiles}` use for `get_raster_tile`), never inline on a worker thread.
- **Point-cap is already enforced in the engine** (`MAX_POINTS = 8_000_000`, truncate + WARN) as defence-in-depth, but consider a per-collection configurable cap and/or a response-size limit at the API layer too.
- **Apply the #211 snapshot-cache to `volume_info()` + `raster_info()`** for the PVOL engine when these go on the per-request path (both currently clone parameters/times per call).

mrauhala:
Two more boundary/API decisions surfaced by #348 review (PR #353), to handle when wiring the route:

[...]

PR #353, opened 2026-06-08
feat(3dtiles): ds-core VolumeEngine + ds-3dtiles encoder crate (#348)

### LINKED ISSUE #348: 3D Tiles Phase 2: ds-core VolumeEngine trait + VoxelGrid type + ds-3dtiles encoder crate

mrauhala:
Phase 2 of the **OGC 3D Tiles** epic (#346). Closes #348. Promotes the Phase-1 demo's hand-rolled geometry + encoding (#347) into the codebase's normal shape: engines return a domain type, a framework-free crate turns it into bytes — exactly how `ds-render` does PNG and `ds-mvt` does MVT.

## What's here

- **ds-core**
  - `geo::geodetic_to_ecef` — WGS84 → ECEF (EPSG:4978), with tests pinning known points (equator, pole, 90°E).
  - new `volume` module — `VolumePoint` / `VolumePointCloud` domain types + the **`VolumeEngine`** trait (`read_point_cloud`, `volume_info`). Framework-free (chrono only), alongside the other engine traits.
- **ds-3dtiles** (new crate; deps `ds-core` + `ds-render`)
  - `encode_pnts(cloud, colormap) -> Vec<u8>` and `tileset_json(cloud, uri) -> String`.
  - Bakes in the Phase-1 invariants so every caller gets them: tileset **`geometricError > 0`** (or CesiumJS never requests the content), **ECEF-native `.pnts` POSITION** (RTC_CENTER, no Y-up flip), geodetic **`region`**, **`u32` bounds-checks**, and **rejection of non-finite geometry** (JSON built via `serde_json`, not `format!`). 4 unit tests.
- **engine-odim** — `impl VolumeEngine for PolarVolumeSiteView`: the cross-section sampler generalized from a 1-D path to the full `(sweep, ray, bin)` lattice, using the true **slant→ground 4/3-Earth** geometry (the 2-D render path uses the ground-range interim). Drops nodata cells (honours the surveyed envelope) and a `min_value` floor.
- **gen_3dtiles example** is now a **thin driver** over `PolarVolumeEngine` + `ds-3dtiles` — the hand-rolled projection/encoding is gone. Still renders the fivih volume in CesiumJS (verified end-to-end, 377k DBZH points; screenshot reproduced locally).
- **Integration test** samples the real fixture into a cloud and round-trips it through `ds-3dtiles` (same fixture-absent skip guard as the sibling PVOL tests, so CI stays green).

## Design notes

- The domain type is a **point cloud** (`VolumePointCloud`), which matches the `.pnts` format and radar's native polar sampling (one point per cell). A dense, regular **voxel grid** is the Tier-B `EXT_primitive_voxels` path, tracked in #351 — I named the type for what it is rather than the issue's tentative "VoxelGrid".
- **Deferred follow-ups** (neither modelled by the engine today): multi-quantity batch-table properties (carry DBZH/VRADH/ZDR/RHOHV together), and per-ray azimuths (`how/startazA`) for sector scans — the `Sweep` struct assumes uniform `360/nrays` spacing, as does the existing 2-D/EDR path.
- **Server/API wiring is #349** — no routes touched here; this is the library + engine + tests layer.

## Checks
`cargo clippy --workspace --all-targets` clean; `ds-3dtiles` (4), `ds-core` (195, incl. new ECEF test), `engine-odim` (134 unit + 28 integration, incl. the new VolumeEngine test) all green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -999,6 +999,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ds-3dtiles"
+version = "0.1.0"
+dependencies = [
+ "ds-core",
+ "ds-render",
+ "serde_json",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "ds-core"
 version = "0.1.0"
 dependencies = [
@@ -1141,7 +1151,9 @@ version = "0.1.0"
 dependencies = [
  "arc-swap",
  "chrono",
+ "ds-3dtiles",
  "ds-core",
+ "ds-render",
  "ds-storage",
  "hdf5-reader",
  "ndarray",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "crates/storage",
     "crates/render",
     "crates/ds-mvt",
+    "crates/ds-3dtiles",
     "crates/api-edr",
     "crates/api-features",
     "crates/api-wms",
--- a/crates/core/src/geo.rs
+++ b/crates/core/src/geo.rs
@@ -5,6 +5,24 @@ const WGS84_A: f64 = 6_378_137.0; // semi-major axis (meters)
 const WGS84_F: f64 = 1.0 / 298.257223563; // flattening
 const WGS84_E2: f64 = 2.0 * WGS84_F - WGS84_F * WGS84_F; // eccentricity squared
 
+/// Convert WGS84 geodetic coordinates to Earth-Centered, Earth-Fixed (ECEF)
+/// metres (EPSG:4978). `lon_deg`/`lat_deg` are degrees, `h` is metres above
+/// the WGS84 ellipsoid. Used to place 3D content (e.g. OGC 3D Tiles volumes)
+/// at its true global position; the inverse is not needed here.
+pub fn geodetic_to_ecef(lon_deg: f64, lat_deg: f64, h: f64) -> [f64; 3] {
+    let lon = lon_deg.to_radians();
+    let lat = lat_deg.to_radians();
+    let (sin_lat, cos_lat) = lat.sin_cos();
+    let (sin_lon, cos_lon) = lon.sin_cos();
+    // Radius of curvature in the prime vertical.
+    let n = WGS84_A / (1.0 - WGS84_E2 * sin_lat * sin_lat).sqrt();
+    [
+        (n + h) * cos_lat * cos_lon,
+        (n + h) * cos_lat * sin_lon,
+        (n * (1.0 - WGS84_E2) + h) * sin_lat,
+    ]
+}
+
 /// Coordinate reference system.
 ///
 /// Stores projection parameters and provides forward/inverse transforms
@@ -1848,5 +1866,25 @@ mod tests {
         let (rlon, rlat) = crs.forward(24.0, 60.0);
         assert!((rlon - 24.0).abs() < 0.01, "rlon={rlon}");
         assert!((rlat - 60.0).abs() < 0.01, "rlat={rlat}");
+    }
+
+    #[test]
+    fn ecef_known_points() {
+        // (0,0,0): on the equator at the prime meridian, ECEF = (a, 0, 0).
+        let [x, y, z] = geodetic_to_ecef(0.0, 0.0, 0.0);
+        assert!((x - WGS84_A).abs() < 1e-3, "x={x}");
+        assert!(y.abs() < 1e-6 && z.abs() < 1e-6, "y={y} z={z}");
+
+        // North pole: x=y=0, z = polar radius b = a(1-f) (+height).
+        let b = WGS84_A * (1.0 - WGS84_F);
+        let [x, y, z] = geodetic_to_ecef(123.0, 90.0, 0.0);
+        assert!(x.abs() < 1e-6 && y.abs() < 1e-6, "x={x} y={y}");
+        assert!((z - b).abs() < 1e-3, "z={z} b={b}");
+
+        // 90°E on the equator: ECEF = (0, a, 0); height adds along the radial.
+        let [x, y, z] = geodetic_to_ecef(90.0, 0.0, 100.0);
+        assert!(x.abs() < 1e-6, "x={x}");
+        assert!((y - (WGS84_A + 100.0)).abs() < 1e-3, "y={y}");
+        assert!(z.abs() < 1e-6, "z={z}");
     }
 }
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -14,3 +14,4 @@ pub mod ogc_extent;
 pub mod openapi;
 pub mod resample;
 pub mod vertical;
+pub mod volume;
--- a/crates/core/src/volume.rs
+++ b/crates/core/src/volume.rs
@@ -76,7 +76,10 @@ pub trait VolumeEngine: Send + Sync {
     ///
     /// - `quantity`: which parameter to sample (`None` → the default).
     /// - `time`: select the retained volume **nearest** this valid time
-    ///   (`None` → latest).
+    ///   (`None` → latest). There is **no staleness cap** — the nearest
+    ///   retained volume is always returned regardless of age (matching the
+    ///   `MapEngine` raster path); a caller needing freshness must check the
+    ///   returned time, or the engine config must bound retention.
     /// - `min_value`: drop points whose physical value is below this (`None` →
     ///   keep every non-nodata sample). E.g. a dBZ floor to cut clutter.
     /// - `reference_time`: forecast model run (`None` → latest; ignored by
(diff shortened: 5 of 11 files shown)


mrauhala:
Addressed the review in 0bce8f0:

1. **Non-finite antenna coords** → `read_point_cloud` now validates `site.{lon,lat,height}.is_finite()` up front and returns an error naming the antenna (restores the Phase-1 guard the engine path had dropped), instead of an opaque encoder `NonFinite("point offset")`.
2. **Empty cloud → invalid `.pnts`** → `encode_pnts` now *rejects* an empty cloud with `Tiles3dError::Empty` (spec requires `POINTS_LENGTH >= 1`); the test asserts rejection and the precondition is documented.
3. **`volume_info()` O(n) per call** → **deferred to #349** with an in-code note. Rationale: the sibling `raster_info()` in this same engine already clones identically, `volume_info` isn't on a per-request path until the API wiring lands, and the proper #211-style `ArcSwap` snapshot should fix *both* accessors together rather than half-fixing one now.

Minor: added the `asset.version = "1.1"` explanatory comment. Camera-altitude minor left as-is (demo-only, antenna height < 200 m). `fmt`/`clippy` clean; ds-3dtiles + engine-odim suites green.

mrauhala:
Round-2 review addressed in 358cb49 — all three:

1. **Needless alloc in `tileset_json`** → removed `region.to_vec()`; `[f64; 6]` serializes to a JSON array directly.
2. **Missing `elangle` finite-guard** → added `!sweep.elangle.is_finite()` to the sweep guard (matches the `rscale` precedent → clean sweep-level skip instead of NaN offsets).
3. **Camera-altitude regression** → `Location` carries no elevation, so I frame above the cloud's region min-height (≈ antenna altitude) + 90 km, restoring site-relative framing (90182 m for the fivih fixture). `fmt`/`clippy` clean; suites green.

mrauhala:
Round-3 review addressed in 7fa79fa:

1. **No cap on point-cloud size** → `volume_point_cloud` now bounds at `MAX_POINTS = 8_000_000` (≈120 MB; real volumes are ~0.38M) with a truncate-and-`WARN`, in the engine as defence-in-depth — before the HTTP path exists.
2. **Sync method blocks the runtime** → deferred to #349 (the trait is sync by design like the other engines; the fix is dispatching via `spawn_blocking` on the dedicated runtime in the handler). Logged as a requirement on https://github.com/mrauhala/meteocore/issues/349.
3. **`time` doc mismatch** → corrected to say it selects the *nearest* retained volume.

`fmt`/`clippy` clean; engine-odim 134 unit + 28 integration green.

mrauhala:
Round-4 review addressed in 22e7f83 — all three:

1. **`POINTS_LENGTH` u32 bounds-check** → `u32::try_from(count)` → `TooLarge("POINTS_LENGTH")` before the body build (the byte-length guards only caught it indirectly at ~4 GiB).
2. **Vec capacity** → reserve the true upper bound (matching sweeps' `nrays*nbins`, capped at `MAX_POINTS`) so the fill loop never reallocates.
3. **Memory-estimate comment** → corrected to 24 B/point → ~192 MB at the cap.

`fmt`/`clippy` clean; ds-3dtiles 4 + engine-odim 134/28 green.

mrauhala:
Round-5 review — one new item fixed, two already-deferred repeats:

1. **`encode_pnts` allocates body before the byte-length check** → fixed in dbfa92d: front-loaded the body byte-budget guard (`count*15 + padding` must fit u32) so an oversized cloud is rejected *before* any multi-GB allocation.
2. **Sync blocking I/O in `read_point_cloud`** → already deferred to #349 (the review agrees: handler-layer `spawn_blocking`, addressed with `get_raster_tile`'s same obligation). Tracked on #349.
3. **`volume_info()` per-call clone** → already deferred to #349 with the in-code NOTE + #211 snapshot-cache plan. Tracked on #349.

`fmt`/`clippy` clean; ds-3dtiles green.

mrauhala:
Round-6 review in 2fa4ac5:

1. **Truncated cloud region** — kept the tight region (documented). A 3D Tiles bounding volume must enclose the *tile's content*; when MAX_POINTS truncates, the dropped cells aren't in the tile, so the bbox of the included points is the correct (and tightest, best-for-culling) bound. The break happens before the push, so region encloses every included point. Enlarging it to the full sweep extent would advertise empty space and weaken culling. Expanded the comment to make this explicit for the truncated case (it previously only spoke to the empty case).
2. **Test gap** → added a `NonFinite("point offset")` case to `non_finite_geometry_is_rejected`.

`fmt`/`clippy` clean; ds-3dtiles 4 tests green.

mrauhala:
Round-7 review addressed in 606aba3:

1. **`volume_info()` O(n) clone** → **done now, not deferred.** `VolumeEngine::volume_info` now returns `Arc<VolumeInfo>`, and the PVOL engine caches a pre-built snapshot in `SiteMeta` (rebuilt on catalog swap) → O(1) Arc clone per call, no Vec clone, satisfying the #211 rule. Doing it here is the right call since the trait has no API consumers yet (callers read fields via `Deref`, no changes needed). #349's note is now resolved for `volume_info`; `raster_info` (the pre-existing MapEngine trait, returns by value) stays as-is.
2. **`rstart` finite-guard** → added to the sweep guard alongside `rscale`/`elangle`.
3. **`content_uri` constraint** → documented that callers must pass a server-controlled relative path.

Minor (CLAUDE.md capabilities table): deferred to #349 — 3D Tiles isn't a *served* API until the route lands, so listing it now would misrepresent the table's 'APIs' column.

`clippy --workspace --all-targets` clean; ds-3dtiles 4 / ds-core 195 / engine-odim 134+28 green.

mrauhala:
Round-8 review — the reviewer noted **no blocking correctness/security issues**; all four were minor/forward-looking.

- **Default-quantity inconsistency** → fixed in 3ac3ac9: `read_point_cloud` and `volume_info.default_quantity` now share one source (`quantities.first()`), so the advertised default == the rendered default.
- **Silent `volume_info()` fallback** → documented as intentional graceful degradation for a dropped-NOD reload race (a per-request `warn!` would spam).
- **`content_uri` runtime guard** and **truncation flag** → deferred to #349 (reviewer agreed both are API-boundary decisions); logged at https://github.com/mrauhala/meteocore/issues/349.

`clippy --workspace --all-targets` clean; engine-odim 134+28 green. From my side this PR is converged — eight rounds in, the findings are now minor and the remaining items are genuinely #349's.

mrauhala:
Round-9 review (reviewer: **no blocking issues**) — both notes addressed in f92b41f:

1. **MSRV** → declared `rust-version = "1.87"` on ds-3dtiles (the new `is_multiple_of`/`is_none_or` idioms, both clippy-preferred, set that floor); an older toolchain now fails clearly.
2. **Unknown quantity → 404** → fixed at the engine layer: `read_point_cloud` now returns `InvalidParameter` (→ 400) for an unrecognised quantity, matching `get_raster_tile`/`polar_sample`, so #349's handler gets the right status without extra work. Added an integration assertion.

`clippy --workspace --all-targets` clean; engine-odim 134 + 28 green.

mrauhala:
Round-10 review addressed in 2680e3e:

1. **`content_uri` runtime guard** → `tileset_json` now validates it (rejects empty / absolute / `://` / `..`) with a new `Tiles3dError::InvalidUri`, so the `pub` API is safe-by-default and a handler can't emit a traversal/redirect URL. Tested.
2. **Duplicate error label** → the pre-allocation budget guard is now `TooLarge("point data size")`, distinct from the post-build `TooLarge("featureTableBinary")`.
3. **Staleness cap** → documented the intentional "nearest, no cap" contract on the trait (matches `get_raster_tile`); a configurable `max_age` would be a follow-up feature, not a v1 gap.

`clippy --workspace --all-targets` clean; ds-3dtiles 5 / ds-core 195 / engine-odim 134+28 green.

PR #355, opened 2026-06-08
feat(api-3dtiles): bundled CesiumJS viewer with collection picker

mrauhala:
A committed, server-served CesiumJS viewer for the 3D Tiles API — the throwaway test page from the live demo, now a proper bundled page with a **collection + quantity picker**.

## What's here
- **`crates/api-3dtiles/viewer/index.html`** — a self-contained CesiumJS page: a top bar with a **Collection** dropdown (populated from `/3dtiles/collections`), a **Quantity** dropdown (from the collection doc, reflectivity-first), and a **min dBZ** threshold. CARTO Dark Matter basemap, token-free (no Cesium Ion). Out-of-order load guard for rapid switching.
- **Served at `GET /3dtiles/viewer`** (`include_str!`-baked into the binary). So once deployed it's live at e.g. **`https://meteocore.app.meteo.fi/3dtiles/viewer`**.
- **Same-origin API base by default** (derives `…/3dtiles` from its own URL, so it works on any host), with a **`?base=`** override for local dev (`/3dtiles/viewer?base=http://127.0.0.1:8131/3dtiles`).
- **`tileset.json` now accepts `min_value`** and carries it into the `content.uri`, so the viewer's threshold control actually reaches the `.pnts` fetch (cuts clutter + tile size).
- Landing page links the viewer (`rel=alternate`, `text/html`).

## Verified
- `GET /3dtiles/viewer` → `200 text/html` against the running server; landing advertises it.
- Tests: viewer page is served as HTML and references CesiumJS + the collections API; `tileset.json?...&min_value=5` → `content.pnts?quantity=DBZH&min_value=5`.

## Checks
`clippy` clean; api-3dtiles 1 unit + 11 integration green. CLAUDE.md documents the endpoint.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -243,10 +243,14 @@ framework-free `ds-3dtiles` crate encodes it to a `.pnts` tile + `tileset.json`;
   `PolarVolumeSiteView`; the cloud is bounded by `MAX_POINTS` (8M, truncate +
   WARN). Unknown quantity ⇒ `InvalidParameter` (→ 400).
 - **Routes** (mounted at `/3dtiles`): `GET /collections/{id}/tileset.json`
-  (`?quantity=&datetime=`) and `GET /collections/{id}/content.pnts`
+  (`?quantity=&datetime=&min_value=`) and `GET /collections/{id}/content.pnts`
   (`?quantity=&datetime=&min_value=`), plus `/` · `/collections` ·
-  `/collections/{id}`. The tileset's `content.uri` embeds the resolved quantity
-  (+ pinned time) so the `.pnts` fetch is deterministic.
+  `/collections/{id}` · **`/viewer`** (a bundled CesiumJS page with a
+  collection + quantity picker, `include_str!`-baked from
+  `crates/api-3dtiles/viewer/index.html`; same-origin API base by default, so it
+  works on any deployment, with a `?base=` override). The tileset's
+  `content.uri` embeds the resolved quantity (+ pinned time + `min_value`) so
+  the `.pnts` fetch is deterministic.
 - **Concurrency:** `read_point_cloud` is sync (blocking HDF5 I/O + a long CPU
   loop), so the handler bounds it with the shared render semaphore and runs it
   via `spawn_blocking` — never inline on a request worker (the same pattern the
--- a/crates/api-3dtiles/src/handlers.rs
+++ b/crates/api-3dtiles/src/handlers.rs
@@ -169,7 +169,12 @@ pub async fn get_tileset(
         query.push_str(&format!("&datetime={}", dt.format("%Y-%m-%dT%H:%M:%SZ")));
     }
     if let Some(min) = params.min_value {
-        // f64 Display is URL-safe for a dBZ-range value (digits, `.`, `-`).
+        // `serde_urlencoded` parses "NaN"/"inf" as valid f64; a non-finite
+        // threshold filters every point → a silently-empty tile. Reject → 400.
+        if !min.is_finite() {
+            return Err(Tiles3dError::BadRequest("min_value must be finite".into()));
+        }
+        // f64 Display is URL-safe for a finite value (digits, `.`, `-`).
         query.push_str(&format!("&min_value={min}"));
     }
     let content_uri = format!("content.pnts?{query}");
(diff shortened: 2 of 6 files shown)


mrauhala:
Addressed all three in c08d65a:

1. **SRI on CesiumJS** → added `integrity=sha384-…` + `crossorigin=anonymous` to the version-pinned script + CSS (computed from the live CDN; cesium.com sends `ACAO: *` so SRI loads cleanly). Verified the hashes are in the served page. Documented that Cesium's runtime workers/WASM remain un-SRI'd — self-hosting would be the full fix.
2. **`min_value=NaN`/`inf` → silent empty tile** → now `400` in both `tileset.json` and `content.pnts` (verified live: `?min_value=NaN` → 400). Test added for NaN/inf/-inf on both routes.
3. **Public-instance fallback** → kept (intentional for the opened-from-disk case you asked for) but documented the base-resolution priority + the third-party trust boundary.

`clippy` clean; api-3dtiles 1 unit + 12 integration green.

mrauhala:
Round-2 review addressed in d5ce7f3:

- **GPU leak (real bug)** → `loadTileset` now calls `ts.destroy()` when a newer load supersedes the in-flight one, freeing CesiumJS GPU buffers on rapid collection/quantity switching.
- **SRI verification** (informational) → the sha384 hashes were computed directly from the live CDN with `openssl dgst -sha384 | openssl base64` (the exact spot-check), so they protect rather than silently break.
- **Public-URL fallback** (informational, no change requested) → kept + documented in-code; it only fires for the opened-from-disk case.

api-3dtiles tests green.

mrauhala:
Round-3 review in 318f778:

1. **CSP header** → **deferred, with reasoning.** A correct CSP for CesiumJS needs `'unsafe-inline'` (the viewer's inline JS), `'wasm-unsafe-eval'` (Cesium compiles WASM), `worker-src blob:` (Cesium workers), and permissive `connect-src`/`img-src` (cross-origin `?base` + CARTO tiles) — the suggested snippet omits the WASM/worker bits and would silently blank the page. Getting any directive wrong = blank viewer with no JS error, and I can't browser-verify a CSP in this environment. The complete fix is **self-hosting CesiumJS** (already flagged in the viewer comment); worth its own issue rather than a risky blind CSP here.
2. **`?base=` validation** → now only accepts an `http(s)` override (rejects `javascript:`/`data:`/etc.); falls back to same-origin otherwise.
3. **SRI verifiability** → added `scripts/check-cesium-sri.sh` (re-runnable, used on version bumps). Ran it: the pinned 1.124 hashes **match the live CDN** ✓ — so the SRI is protecting, not silently breaking.

api-3dtiles tests green.

mrauhala:
Round-4 review in df7b847:

1. **"GPU leak on the replaced-current path"** → not a leak: `viewer.scene.primitives` has `destroyPrimitives: true` (the default), so `.remove(current)` already destroys the old tileset and frees its GPU buffers. The superseded-load path needs explicit `destroy()` only because *that* tileset was never added to the collection — calling `destroy()` after `remove()` would double-destroy and throw. Added a comment documenting this so it's not 'fixed' into a double-destroy later.
2. **Hardcoded `dBZ` label** → relabeled to neutral "min value" (min_value applies to any quantity, not just reflectivity).
3. **Hardcoded fallback URL** → already documented (rounds 1/3) as an intentional opened-from-disk trust boundary.

api-3dtiles tests green.

mrauhala:
Round-5 review in 34fbf14:

1. **`?base=` not normalised** → now validated/normalised with `new URL()` (resolves `..`, throws on malformed) requiring http(s), replacing the scheme-prefix regex.
2. **SRI not enforced in CI** → documented as a **mandatory step** right at the SRI tags (recompute via `scripts/check-cesium-sri.sh` on a version bump). Deliberately *not* a hard CI gate: the check curls the cesium CDN, and a CDN-reachability dependency in CI is the same flaky failure mode that already red-X'd a Docker run here.
3. **Hardcoded fallback** (design note, not blocking, 4th mention) → kept intentionally — the standalone page is *meant* to target the public API when opened from disk (the original request). The reviewer's alternative (error instead of fallback) would break that.

api-3dtiles tests green. From my side the PR is converged — 5 rounds, now polish on a demo viewer.

Issue #357, opened 2026-06-08
3D Tiles: isosurface (triangle-mesh) representation — storm shells / echo tops via marching cubes


mrauhala:
Part of #346.

Add an **isosurface (triangle-mesh)** representation for volumetric data, alongside the point cloud (today) and the planned voxels (#351). For a volumetric scalar field, the natural mesh is a **marching-cubes isosurface** at a threshold — and for radar these map onto operational products.

## Why it's compelling for radar

- **Storm-cell shells** — the 40/50 dBZ surface shows convective cores as solid 3D objects.
- **Echo tops** — the height surface of a reflectivity threshold (a standard nowcasting product).
- **Hail / ZDR-column** and **bright-band / melting-layer** surfaces — dual-pol signatures as 3D shapes.
- **CAPPI** — a textured mesh at a constant altitude.

## Why a mesh, vs the other representations

- **Surfaces read as solid 3D; point clouds don't.** A mesh is **lit, occluding, depth-sorted** — a supercell rendered as a shaded isosurface looks like a 3D object; the point cloud is unlit speckle with no occlusion.
- **Small payload** — a surface is far less data than a volume of points (~13 MB `.pnts` → a few hundred KB of mesh).
- **Standards-clean, non-deprecated** — a triangle mesh is **glTF 2.0**, first-class 3D Tiles 1.1 content, with none of the glTF-POINTS rendering gotcha. So this also sidesteps the `.pnts` deprecation (#356) for this representation.

## How it fits the three representations (complementary, by question)

- **Voxels (#351):** the full continuous field — "what's the reflectivity everywhere" (ray-march/transparency).
- **Isosurface mesh (this issue):** discrete features — "where are the storm cores and how tall" — legible, shareable, lit, small.
- **Point cloud (today):** the simplest faithful dump; good default, least communicative.

## Tradeoffs / risks

- **It's a reduction:** one surface = one threshold; you lose the continuous field. Nested translucent shells (e.g. 35/45/55 dBZ) bring transparency-sorting fiddliness.
- **Needs structured input + extraction:** resample the polar volume to a regular `(radius, azimuth, height)` lattice (the same one voxels need), then marching cubes — more server compute than dumping points, but smaller output.
- **Gaps/noise:** radar is holey (cone of silence, beam spreading, attenuation). Isosurfaces can be jagged, and the mesher **must not fabricate surface across unscanned gaps** — respect the coverage envelope, like the current sampler.

## Scope (sketch)

[...]

PR #358, opened 2026-06-08
feat(3dtiles): cylindrical VoxelGrid resampling foundation (#351 Phase 5a)

mrauhala:
**Phase 5a** of the 3D Tiles epic (#346): the structured voxel substrate for #351, built ahead of the volatile glTF voxel *encoding*.

## What's here
- **ds-core `VoxelGrid`** — a regular **cylindrical** grid (`radius` = ground range, `angle` = azimuth, `height`), one `f32`/cell with **`NaN` = nodata**, plus `index()` (height-fastest order) and `valid_count()`. New `VolumeEngine::read_voxel_grid(quantity, time, dims, reference_time)`.
- **engine-odim `read_voxel_grid`** on `PolarVolumeSiteView` — resamples the polar volume into the grid by inverting each cell centre to `(slant_range, elevation)` via the 4/3-Earth beam model and sampling through the **existing envelope-guarded `sample_polar_slant`**, so the cone of silence and unscanned gaps stay `NaN` (no fabricated data). Default `128×360×48` (1° azimuth ≈ radar rays), capped at `MAX_VOXELS` (32M cells).
- Extracted a shared `select_entry_and_quantity` (quantity validation/default + nearest-time volume selection + antenna-finite guard) now used by **both** `read_point_cloud` and `read_voxel_grid` (dedup).

## Why this split
The actual voxel *tile* encoding (`EXT_primitive_voxels` + `3DTILES_content_voxels` + the cylinder bounding volume) is **draft and actively in flux** — the bounding volume was just renamed `3DTILES_bounding_volume_cylinder` → `EXT_implicit_cylinder_region`, and there are *open* CesiumJS PRs fixing voxel bounds/transforms. Rendering also needs CesiumJS ≥1.127 voxel support, which can't be render-verified in CI. So this PR ships the **verifiable, stable** half (the resampling + domain type, also shared with isosurfaces #357), and the glTF encoding lands as a focused follow-up once the spec settles and can be visually verified.

## Tests
- ds-core unit (CI-always): `index()` ordering + `valid_count()`.
- engine-odim integration: resamples the **real fivih fixture** — asserts dims tile the grid, `valid_count` is non-zero **and** less than total (NaN gaps from the cone of silence), and every sampled value is a sane dBZ.
- api-3dtiles `MockVolume` implements the new trait method.

`clippy --workspace --all-targets` clean; ds-core 196, engine-odim 134+29, api-3dtiles 12 green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -237,11 +237,18 @@ framework-free `ds-3dtiles` crate encodes it to a `.pnts` tile + `tileset.json`;
 (same rule as `MapEngine`→`ds-render`).
 
 - **Trait:** `VolumeEngine::read_point_cloud(quantity, time, min_value, reference_time)`
-  → `VolumePointCloud`, and `volume_info() -> Arc<VolumeInfo>` (O(1) cached
-  snapshot, #211 — carries quantities, times, default quantity, and a coverage
-  `region` for the tileset bounding volume without sampling). Implemented by
-  `PolarVolumeSiteView`; the cloud is bounded by `MAX_POINTS` (8M, truncate +
-  WARN). Unknown quantity ⇒ `InvalidParameter` (→ 400).
+  → `VolumePointCloud`; `read_voxel_grid(quantity, time, dims, reference_time)`
+  → `VoxelGrid` (a regular **cylindrical** grid — `radius`×`angle`×`height`,
+  `NaN`=nodata — the substrate for true voxels #351 and isosurfaces #357);
+  and `volume_info() -> Arc<VolumeInfo>` (O(1) cached snapshot, #211 — carries
+  quantities, times, default quantity, and a coverage `region` for the tileset
+  bounding volume without sampling). Implemented by `PolarVolumeSiteView` (which
+  shares `select_entry_and_quantity` across both samplers); the cloud is bounded
+  by `MAX_POINTS` (8M) and the grid by `MAX_VOXELS` (32M cells). Both resample
+  via the envelope-guarded `sample_polar_slant` (no fabricated data across the
+  cone of silence). Unknown quantity ⇒ `InvalidParameter` (→ 400). The
+  `EXT_primitive_voxels` glTF *encoding* of a `VoxelGrid` is a follow-up (#351) —
+  draft spec, render-verify against CesiumJS ≥1.127.
 - **Routes** (mounted at `/3dtiles`): `GET /collections/{id}/tileset.json`
   (`?quantity=&datetime=&min_value=`) and `GET /collections/{id}/content.pnts`
   (`?quantity=&datetime=&min_value=`), plus `/` · `/collections` ·
--- a/crates/api-3dtiles/tests/integration.rs
+++ b/crates/api-3dtiles/tests/integration.rs
@@ -57,6 +57,31 @@ impl VolumeEngine for MockVolume {
         })
     }
 
+    fn read_voxel_grid(
+        &self,
+        _quantity: Option<&str>,
+        _time: Option<chrono::DateTime<chrono::Utc>>,
+        dims: Option<[usize; 3]>,
+        _reference_time: Option<chrono::DateTime<chrono::Utc>>,
+    ) -> Result<ds_core::volume::VoxelGrid, DataServerError> {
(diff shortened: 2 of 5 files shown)


mrauhala:
Round-1 review addressed in becf606:

1. **`valid_count()` doc/impl mismatch** → doc now says "finite (excludes NaN *and* ±∞)", matching `is_finite()`.
2. **`read_voxel_grid` no default** → added a default impl returning a "not supported" error. It's the *optional* structured capability (vs `read_point_cloud`, the baseline), so a new `VolumeEngine` opts in rather than failing to compile. Used `DataServerError::Engine` (no enum blast radius) rather than adding a `NotImplemented` variant; the api-3dtiles mock now drops its unused impl and uses the default.
3. **Dead test block** → removed the `let _ = apex` cone-of-silence block (asserted nothing; `index()` is covered by the ds-core unit test).

Forward note (spawn_blocking for the voxel handler) → noted; it lands with the 5b HTTP route, same pattern as `read_point_cloud`. `clippy` clean; ds-core 196, engine-odim 134+29, api-3dtiles 12 green.

mrauhala:
Round-2 review addressed in d1f821c: the default `read_voxel_grid` now returns `LocationNotFound` (→ 404), not `Engine` (→ 500). A voxel-unsupported engine is a capability gap (resource-not-found-for-this-collection), not a server fault — so the future voxel route answers 404, matching `read_point_cloud`'s no-data path. ds-core green.

mrauhala:
Round-3 review addressed in 4af6f58:

1. **Duplicated index formula** → extracted `VoxelGrid::index_of(dims, i_r, i_a, i_h)` (associated fn, usable before the grid exists). `index(&self)` delegates to it, and the engine fill loop now calls it — one source of truth for the axis order (which is in flux per #351).
2. **No bounds assertion** → `index_of` now `debug_assert!`s all three axes in range, with axis context, so an off-by-one surfaces immediately in dev/test.

`clippy` clean; ds-core 196, engine-odim 134+29 green.

mrauhala:
Round-4 review addressed in 67b473b:

1. **`radius_max` missed `rstart` finiteness** → the fold now filters `rscale.is_finite() && rscale>0 && rstart.is_finite()`, so a NaN `rstart` can't silently drop from `f64::max` and under-compute the radial extent (mirrors the sweep-extent helper).
2. **Double O(N) scan** → `voxel_grid_from_volume` counts sampled cells in the fill loop and returns `(grid, valid)`; `read_voxel_grid` uses that instead of a second `valid_count()` pass over up to 32M cells.

(spawn_blocking handler note → for the 5b HTTP route.) `clippy` clean; ds-core 196, engine-odim 134+29 green.

mrauhala:
Round-5 review addressed in 960aee5:

1. **`valid` vs `valid_count()` discrepancy** → the fill loop now writes/counts only when `(v as f32).is_finite()`. A `Some(NaN)` (non-finite gain/offset in a malformed file) would otherwise pass the `valid>0` 404 guard while `valid_count()` saw 0 — now consistent, and NaN is correctly treated as nodata (also guards an f64→f32 overflow-to-inf).
2. **`derive(Clone)` on a ~128 MB struct** → removed (nothing clones it; move it into the encoder). YAGNI per CLAUDE.md.
3. **`reference_time` dropped** → documented on the shared helper (PVOL has no model runs; accept-and-ignore).

`clippy` clean; ds-core 196, engine-odim 134+29, api-3dtiles 12 green.

mrauhala:
Round-6 review — one item, and the reviewer scopes it to the **voxel route PR (5b)**: the default `read_voxel_grid` 404 ("not supported") will be indistinguishable from the impl's `valid==0` 404 ("no data") once a route calls it on a non-voxel engine. The fix is a route-layer `supports_voxel_grid` capability check.

**Deferred to 5b** (where the route lives) and recorded as a requirement on #351 — adding an unused `supports_voxel_grid` flag in this foundation PR would contradict the round-5 YAGNI fix (removing the unused `Clone`). No 5a code change.

From my side this PR is converged: the resampling foundation is correct + tested, and the remaining item is a route-layer concern with no consumer yet. MERGEABLE/CLEAN.

Issue #350, opened 2026-06-08
3D Tiles Phase 4: time-dynamic tilesets (per-timestep animation)


mrauhala:
Part of #346. Depends on Phase 3.

Animate radar volumes over time.

## Scope
- Per-timestep tilesets: `GET /3dtiles/{collection}/{time}/tileset.json` (time = catalog volume timestamp; default = latest).
- A manifest/index of available times for the client to scrub (reuse `RasterInfo`/catalog times).
- Client demo: swap tileset URIs per frame (CesiumJS), looping over the retained volumes.

## Done when
- The fivih fixtures (or a longer pulled sequence) animate smoothly in the demo viewer.
- No per-frame full reload stalls (preload/next-frame strategy documented).

PR #359, opened 2026-06-09
feat(3dtiles): isosurface mesh — encoder, API (both representations), viewer toggle (#357)

mrauhala:
## What

Continuing the OGC 3D Tiles work (epic #346) on the merged `VoxelGrid` foundation (#358): a **marching-tetrahedra isosurface mesher** that turns a cylindrical `VoxelGrid` into a glTF 2.0 `.glb` triangle mesh — e.g. "the 20 dBZ shell" of a radar volume.

### Why isosurface (and not the draft voxel path)
The true `EXT_primitive_voxels` voxel path (#351) is blocked on an unsettled draft spec that needs experimental CesiumJS ≥1.127 to render — not something I can verify at encode time. An **isosurface is a plain glTF triangle mesh**: it renders in *any* 3D Tiles 1.1 client, so it's the verifiable way to show a radar volume's 3-D structure, and it builds directly on the same `VoxelGrid`. The voxel encoder stays tracked in #351 for when the spec settles.

## Changes

- **`ds-core`** — `geo::destination_point` + `EARTH_RADIUS_M`: shared spherical great-circle so the encoder maps polar `(range, azimuth)` samples the *same way* the volume engines do (one formula, one earth radius). Pinned against analytic references (meridian/equator/antimeridian-wrap).
- **`ds-3dtiles`** — `isosurface.rs`: `encode_isosurface_glb(grid, threshold, color)` + `tileset_json_glb(region, uri, rtc_center)`.
  - **Marching *tetrahedra*, not cubes.** A tetrahedron is the complete graph K4, so the surface crosses exactly `|inside|·|outside|` edges (3 → triangle, 4 → quad) and the topology is **correct by construction** — no 256-case edge/triangle table to mis-transcribe. Deliberate, since the output can't be eyeballed at encode time. Cube → 6 tets via the Kuhn/Freudenthal split (watertight across cube boundaries).
  - **NaN-aware:** any tet touching a nodata corner is skipped → no fabricated surface across the cone of silence / beyond the beam fan.
  - **Geometry:** fractional cell index → ground/azimuth/height (same cell-centre convention the engine sampled with) → `destination_point` + `geodetic_to_ecef`, stored antenna-relative and pre-flipped Z-up→Y-up (a runtime re-applies Y-up→Z-up to **glTF** content — the flip `.pnts` skips). Tile `transform` = antenna ECEF; geodetic `region` unaffected. `doubleSided` PBR shell coloured at the threshold.
- **`engine-odim`** — `gen_isosurface` example: real fivih PVOL → `.glb` + `tileset.json` + a token-free CesiumJS viewer (CARTO dark).
- **`CLAUDE.md`** — documents the isosurface path + the marching-tet rationale.

## Verification

- **Unit tests (13 in `ds-3dtiles`, + geo):** glb container well-formed (magic/chunks/alignment); **the surface lands at the exact expected height** over a height-ramp field — this verifies the index→ECEF→Y-up mapping end-to-end without a renderer; empty / NaN-corner / non-finite-threshold guards; tileset transform + region + URI/region validation.
- **Real-data demo** on the committed-locally fivih fixture: 253,636 finite cells → **178,022 triangles**, independently re-parsed — all positions finite, **all normals unit-length (0 bad)**, geodetic region + antenna-ECEF transform correct.

## Render-verify (the one thing I can't check here)
The math is verified by the height test, but I can't drive CesiumJS in this environment. To eyeball:
```
cargo run -p engine-odim --example gen_isosurface
# then serve target/3dtiles-iso-fivih/ and open index.html
```
The shell should stand upright over Vihti. **If it renders tipped 90°**, that's the glTF up-axis convention for direct 1.1 `.glb` content (a one-line fix: store ECEF-native + `asset.gltfUpAxis: "Z"`).

## Follow-ups
- API route (serve `.glb` isosurface from `/3dtiles`, threshold query param, viewer toggle) — encoder + demo ship first, mirroring `.pnts` (#347 demo → #349 API).
- Angular-seam wrap (a one-cell gap at azimuth 0 today) and vertex sharing (mesh is non-indexed) — noted in code.

Part of epic #346. Refs #357, #351, #358.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -265,7 +265,16 @@ into a glTF `.glb` triangle mesh.
   cell-centre convention as the engine sampler) → `destination_point` +
   `geodetic_to_ecef` (both now in `ds_core::geo`), stored antenna-relative and
   pre-flipped Z-up→Y-up because a runtime re-applies Y-up→Z-up to **glTF**
-  content (the flip the `.pnts` path skips — pnts isn't glTF). `tileset_json_glb`
+  content (the flip the `.pnts` path skips — pnts isn't glTF). **`background`
+  sealing (load-bearing for radar):** the grid stores `NaN` for *both* clear-air
+  (`undetect`) and unmeasured (`nodata`/cone of silence) — `RawPixels::sample`
+  masks both identically (reader.rs). `encode_isosurface_glb(grid, threshold,
+  color, background)` with `background=Some(bg<threshold)` treats `NaN` as
+  no-echo so the surface **closes into solid blobs**; `None` skips `NaN`-touching
+  tets and the echo→clear-air boundary renders as open vertical *curtains*. The
+  demo seals with `Some(-32.0)` (the dBZ floor). Splitting `undetect` from
+  `nodata` in the engine sampler (so only the real cone of silence stays open) is
+  a follow-up. `tileset_json_glb`
   carries the antenna ECEF as the tile **`transform`** (glTF content has no
   embedded origin, unlike `.pnts` `RTC_CENTER`); the geodetic `region` is
   unaffected by it. Demo: `cargo run -p engine-odim --example gen_isosurface`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -234,7 +234,9 @@ Volumetric weather as OGC 3D Tiles — radar polar volumes today. The chain:
 cell, placed at its true ECEF position via the 4/3-Earth beam model); the
 framework-free `ds-3dtiles` crate encodes it to a `.pnts` tile + `tileset.json`;
 `api-3dtiles` serves it over HTTP. Engines return the domain type, never bytes
-(same rule as `MapEngine`→`ds-render`).
+(same rule as `MapEngine`→`ds-render`). `ds-3dtiles` has two encoders: the
+`.pnts` point cloud, and (#357) an **isosurface** mesher that turns a `VoxelGrid`
+into a glTF `.glb` triangle mesh.
 
 - **Trait:** `VolumeEngine::read_point_cloud(quantity, time, min_value, reference_time)`
   → `VolumePointCloud`; `read_voxel_grid(quantity, time, dims, reference_time)`
@@ -249,6 +251,29 @@ framework-free `ds-3dtiles` crate encodes it to a `.pnts` tile + `tileset.json`;
   cone of silence). Unknown quantity ⇒ `InvalidParameter` (→ 400). The
   `EXT_primitive_voxels` glTF *encoding* of a `VoxelGrid` is a follow-up (#351) —
   draft spec, render-verify against CesiumJS ≥1.127.
+- **Isosurface meshing (#357, `ds-3dtiles/src/isosurface.rs`):**
+  `encode_isosurface_glb(grid, threshold, color)` extracts a constant-value
+  shell (e.g. "the 20 dBZ surface") from a `VoxelGrid` as a glTF 2.0 `.glb`
+  triangle mesh — a **plain glTF mesh that renders in any 3D Tiles 1.1 client**
+  (the verifiable alternative to the draft `EXT_primitive_voxels` voxel path).
+  Uses **marching tetrahedra**, not marching cubes: a tet is K4, so the surface
+  crosses exactly `|inside|·|outside|` edges and the topology is correct by
+  construction (no 256-case table to mis-transcribe — chosen because the output
+  isn't render-checkable at encode time). Cube → 6 tets (Kuhn split); any tet
+  touching a `NaN` corner is skipped (no surface across the cone of silence).
+  Surface vertices map fractional cell index → ground/azimuth/height (same
+  cell-centre convention as the engine sampler) → `destination_point` +
+  `geodetic_to_ecef` (both now in `ds_core::geo`), stored antenna-relative and
+  pre-flipped Z-up→Y-up because a runtime re-applies Y-up→Z-up to **glTF**
+  content (the flip the `.pnts` path skips — pnts isn't glTF). `tileset_json_glb`
+  carries the antenna ECEF as the tile **`transform`** (glTF content has no
+  embedded origin, unlike `.pnts` `RTC_CENTER`); the geodetic `region` is
+  unaffected by it. Demo: `cargo run -p engine-odim --example gen_isosurface`
+  (writes `.glb` + `tileset.json` + a token-free CesiumJS viewer). **API route
+  is a follow-up** — the encoder + demo ship first (mirrors how `.pnts` went,
+  demo #347 → API #349). **One thing to confirm visually:** that the runtime
+  applies Y-up→Z-up to direct 1.1 `.glb` content (the standard default); if the
+  shell renders tipped 90°, that's the up-axis convention and a one-line fix.
 - **Routes** (mounted at `/3dtiles`): `GET /collections/{id}/tileset.json`
   (`?quantity=&datetime=&min_value=`) and `GET /collections/{id}/content.pnts`
   (`?quantity=&datetime=&min_value=`), plus `/` · `/collections` ·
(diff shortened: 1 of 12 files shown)


mrauhala:
## Render-verified live in CesiumJS (1.124) — and fixed a real artifact

Drove the demo in an actual CesiumJS session (served `target/3dtiles-iso-fivih/`). Findings:

✅ **Placement/orientation correct** — the shell sits exactly over Vihti, southern Finland (bounding-sphere centre lon 24.50, lat 60.59), lies flat over the terrain (**not** tipped 90°), so the Y-up→Z-up flip is right for direct 1.1 `.glb` content. Colour = radar_dbz at 20 dBZ. Loads with no console errors.

⚠️ **Found "curtains"** — storm→clear-air boundaries rendered as open vertical walls/blades instead of closed blobs. Root-caused to `RawPixels::sample` masking `nodata` *and* `undetect` identically → `NaN`, so the mesher couldn't tell clear air from unmeasured space and skipped those tets.

🔧 **Fixed** (commit `d0ba4f7`) — added `background: Option<f64>` to `encode_isosurface_glb`; `Some(bg)` seals `NaN` as no-echo so the surface closes into solid blobs (never invents echo — assumes its absence). Demo seals with `Some(-32.0)`. Re-verified live: clean closed precipitation blobs. The shallow, flat appearance that remains is physical (widespread stratiform precip ~few km thick), not an artifact.

Engine-side follow-up to split `undetect` from `nodata` (so even the real cone of silence stays open): #360.

mrauhala:
## API + viewer for both representations (commit `7acecbd`)

Per the follow-up ask — point cloud **and** isosurface mesh now both available from the API and choosable in the viewer, both as proper OGC 3D Tiles tilesets.

- **`tileset.json?representation=points|isosurface`** — `points` → `.pnts` (region-only, `RTC_CENTER` self-places); `isosurface` → `.glb` + a tile `transform` = the antenna ECEF (built from `VolumeInfo.origin`, no sampling).
- **`GET /content.glb?quantity=&datetime=&threshold=`** — `read_voxel_grid` + `encode_isosurface_glb` on a blocking thread (shared render semaphore), content-ETag, empty→404. Seal floor = the colormap's `domain()` min.
- **Discovery:** collection JSON advertises a `representations` array (driven by `VolumeInfo.supports_voxel_grid`); the viewer reads it to populate a **View** dropdown (Point cloud / Isosurface), relabelling the numeric field between *min dBZ* and *threshold dBZ*.
- **ds-core:** `VolumeInfo` += `supports_voxel_grid` + `origin`. **ds-render:** `ColorMap::domain()`.

Verified over HTTP against the live fivih fixture: `representations: ["points","isosurface"]`, isosurface tileset carries the antenna-ECEF transform, `content.glb` is valid glTF (`model/gltf-binary`), points path unchanged. 16 api-3dtiles integration tests green.

Next per plan: complete review rounds on this PR, then **#360** (split `undetect`/`nodata` in the engine sampler so the seal becomes exact).

mrauhala:
## Review round 1 — addressed (commit `3be290c`)

| # | Finding | Resolution |
|---|---------|------------|
| 1 | **Blocker:** `background >= threshold` not validated (inverted surface returned `Ok`) | `encode_isosurface_glb` now returns the new `Tiles3dError::BackgroundNotBelowThreshold`. Test: `background_not_below_threshold_is_rejected`. |
| 2 | **Perf:** `Vec::with_capacity(4)` ×2 per `march_tet` (millions of allocs) | Replaced with fixed `[usize; 4]` + length counters. Byte-identical output (same 17,152,904-byte glb on fivih). |
| 3 | **Consistency:** `get_content_glb` validates quantity inside the blocking task | Now validated against `info.quantities` **before** `spawn_blocking`, matching the `.pnts` handler. |
| 4 | **Cleanup:** `tileset_json_glb` round-trips through a JSON string | Shared `tileset_value_for_region` (`serde_json::Value`); the glb path injects `transform` on the `Value` directly — no serialize-then-reparse. |
| 5 | **Doc:** `crossing` says `t ∈ [0,1)` but clamp allows `1.0` | Corrected to `[0, 1]`. |
| 6 | **Note:** winding not oriented; `doubleSided` is load-bearing | Comment added in `march_tet` (and the material block already notes it). |

Acknowledged-as-follow-up (not changed here): angular-seam gap at azimuth 0, non-indexed mesh (no vertex sharing), and the `undetect`/`nodata` split (tracked in #360 — next up after this PR's review settles).

15 ds-3dtiles + 16 api-3dtiles tests green; clippy clean.

mrauhala:
## Review round 2 — addressed (commit pushed)

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Missing isosurface `links` entry (discoverability) | `collection_doc` now adds a second `rel:"3dtiles"` link with `?representation=isosurface` when `supports_voxel_grid`. Test asserts both links present. |
| 2 | Seal `floor` uses the shared reflectivity colormap for all quantities | Added a comment: v1 uses the collection-level colormap regardless of quantity (same limitation as the `.pnts` path) — geometrically fine, semantically approximate for non-reflectivity quantities; per-quantity colormaps (#350) fix it. |

Observations (non-blocking, left as tracked follow-ups): angular-seam gap at azimuth 0 (always North for Finnish radars — noted for #357 follow-up), non-indexed mesh (vertex sharing would cut ~13 MB → ~2 MB).

mrauhala:
## Review round 3 — addressed (commit pushed)

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Winding inconsistency → inverted lighting on ~half the surface | `march_tet` now derives the local **outward** direction (inside-centroid → outside-centroid, toward lower values) and `MeshBuilder::push` flips winding+normal to match, so every normal points outward. `doubleSided` kept as belt-and-suspenders. New test `isosurface_normals_point_outward`. |
| 2 | Demo duplicates region logic instead of `VolumeInfo::region` | `gen_isosurface` now uses `info.region` (the production value), falling back to the approximation only when `None`. |
| 3 | No test for `supports_voxel_grid=false` + isosurface → 400 | New `isosurface_on_unsupported_collection_is_400` (a `MockNoVoxel` advertises only `points`, 400s on both isosurface routes). |

16 ds-3dtiles + 17 api-3dtiles tests green; clippy clean. The normal-orientation fix is verified by the new unit test (height-ramp ⇒ all normals point down/outward); I can also render-verify it in CesiumJS if you'd like.

mrauhala:
## Review round 4 — addressed (commit pushed)

| # | Finding | Resolution |
|---|---------|------------|
| 1 | **Bug:** `TooLarge` → HTTP 500 in `get_content_glb` | Now mapped to **400** ("raise the threshold") — it's a client-driven condition, not a server fault. |
| 2 | **Design:** `supports_voxel_grid:true` + `origin:None` representable (→ 500) | Collapsed the two fields into one `VolumeInfo.voxel_grid: Option<VoxelGridCaps { origin }>` — capability + origin are atomic, the bad state is **unrepresentable**, and the isosurface tileset can no longer 500 on a missing origin. |
| 3 | **Perf:** two ECEF round-trips per tet for orientation | Cut to **one**: `push` already has the triangle's own centroid, so only the outside-corner centroid needs projecting (`outward = out_ref − tri_centroid`). Kept world-space rather than the index-space gradient — the cylindrical→ECEF map's handedness isn't something I can render-verify, and a wrong sign would invert every normal. |

Output byte-identical; 16 ds-3dtiles + 17 api-3dtiles tests green; clippy clean.

mrauhala:
## Review round 5 — addressed (commit pushed)

| # | Finding | Resolution |
|---|---------|------------|
| 1 | `BackgroundNotBelowThreshold` falls through to 500 | Explicit match arm → **400** (a bad parameter combination). Unreachable today (the floor is clamped `< threshold`), but the hole is now closed against a future floor-formula change. |
| 2 | `(2,2)` quad always fans from `q0` (mesh-quality, non-blocking) | Added a code note: shorter-diagonal selection is a follow-up tied to vertex sharing; negligible at radar grid scale. |

Minor observations (default threshold for non-reflectivity quantities, non-indexed mesh, angular seam) are all already noted in code and tracked.

16 ds-3dtiles + 17 api-3dtiles tests green; clippy clean. This was the round-5 verdict's only required change — I believe the review is settled.

PR #369, opened 2026-06-09
feat(3dtiles): echo-top-height draped-surface mesher + demo (#362)

mrauhala:
## What

Prototype of the **echo-top-height (ETH)** product — the classic forecaster view of *storm depth*, in 3D (the idea raised in discussion). Collapses the cylindrical `VoxelGrid` to a 2-D height field (per `(radius, azimuth)` column, the highest cell ≥ a threshold, threshold-crossing interpolated for a smooth top) and drapes it as a **height-coloured** glTF `.glb`.

Closes #362 (encoder + demo; API route is a follow-up, demo-first like the isosurface).

## Changes
- **`ds-3dtiles/src/echo_top.rs`** — `encode_echo_top_glb(grid, threshold, height_colormap)`. A regular **indexed grid mesh** (no marching cubes) — a quad is emitted only where all four corner columns have an echo top, so clear-air regions are **holes**, not a sheet at height 0. Per-vertex `COLOR_0` (normalized u8 VEC4) coloured by height + smooth (face-accumulated, up-oriented) normals. Reuses the isosurface's `index_to_gltf_pos` (now `pub(crate)`) and `tileset_json_glb` (antenna-ECEF tile transform).
- **`engine-odim/examples/gen_echo_top.rs`** — fivih demo → `.glb` + `tileset.json` + token-free viewer.

## Verification
- **Render-verified live** (CesiumJS): a draped surface over southern Finland graded **blue/cyan → green/yellow by storm depth**. Just **0.27 MB** (a 2-D surface vs ~17 MB for the isosurface volume). Heights **0–6.5 km (median 2.5 km)** — physically sane for the May-19 precip.
- A real bug was caught during render-verify and fixed: colouring by a *builtin* colormap (`Temperature`, whose stops are in °C) over a 0–15 km range collapses to one colour — the first cut rendered uniformly **red**. Fixed by building the ramp with stops **at height values** (`LutColorMap::from_stops`); documented as a gotcha.
- **Tests:** glb container well-formed (POSITION/NORMAL/COLOR_0/indices); the echo top lands at the **exact expected height** end-to-end (same style as the isosurface height test); empty / non-finite guards. 20 ds-3dtiles tests green, clippy clean.

## Screenshot
A draped echo-top surface over southern Finland, coloured by height (blue shallow → green/yellow deeper) — render-verified in CesiumJS during development.

## Follow-ups
- API route (serve from `/3dtiles` with a representation toggle) — mirrors how the isosurface went (demo first).
[...]

Issue #362, opened 2026-06-09
3D Tiles: echo-top-height (ETH) draped surface mesh


mrauhala:
Derived 2-D height field draped as a colored glTF surface — the classic **Echo Top Height** product, in 3D.

For each `(radius, azimuth)` column of the cylindrical `VoxelGrid`, take the highest altitude where reflectivity ≥ a threshold (the standard is ~18 dBZ; threshold-crossing interpolated for a smooth top). Triangulate the `(r, az)` grid into a draped surface at those heights, **per-vertex coloured by height** (low=blue → high=red), and encode to glTF `.glb` in `ds-3dtiles`.

- Reuses the `#357` isosurface's index→ground/azimuth/height→ECEF→Y-up mapping and the `tileset_json_glb` transform.
- New bit: a regular indexed grid mesh (no marching cubes) + a `COLOR_0` per-vertex colour attribute.
- The draped-grid mesher is **reused by VIL** (sibling issue).
- Tier A (standards-clean glTF, renders in any 3D Tiles 1.1 client).

Prototype in progress (demo-first, like the isosurface went). Part of #346.

PR #370, opened 2026-06-09
feat(3dtiles): echo-top API representation + viewer toggle + reflectivity-scaled point size

mrauhala:
## What

Two 3D Tiles enhancements (builds on the merged echo-top encoder #362):

### 1. Echo-top as a third API representation + viewer toggle
`tileset.json?representation=echotop` serves the extruded **echo-top columns** (the 3-D storm-depth bar field) as a glTF `.glb` + the antenna-ECEF tile `transform`. The two glTF products share `content.glb`, disambiguated by `?representation=echotop` (default = isosurface). Echo-top uses **full `[512,360,64]` resolution** (`ECHO_TOP_DIMS` — ≈ native radial; ~12 s first compute, ETag-cached after) coloured by a height ramp. The collection JSON's `representations` array and the viewer's **View** dropdown gain `echotop`, so points / isosurface / echo-top are all selectable live.

### 2. Point-cloud dot size by reflectivity
`encode_pnts` now writes a per-point `value` into the `.pnts` **batch table** (no `BATCH_ID` ⇒ per-point properties), and the viewer styles `pointSize: clamp((${value} − 5)·0.16 + 1, 1, 10)` — weak echo → ~1 px, strong → ~10 px.

## Verification (render-verified live in CesiumJS)
- **Point cloud:** dots scale with reflectivity — strong cores pop (bigger), weak clutter recedes (1 px); the radial beam structure is clearly visible.
- **Echo top:** the 3-D bar field renders via the **View → Echo top** dropdown (full-res bars, blue→green/yellow by depth).
- **Toggle:** all three representations selectable; the numeric field relabels (min dBZ / threshold dBZ).
- Endpoints over HTTP: `content.pnts` carries the value batch table; `content.glb?representation=echotop` → 200, valid glTF (28 MB); isosurface unchanged.

21 ds-3dtiles + 19 api-3dtiles tests green (incl. a new echo-top route test + the batch-table assertion), clippy clean.

## Notes / follow-ups
- **Echo-top first-compute is ~12 s** (the 11.8 M-cell resample); ETag caches repeats. A precompute/coarser-default or background cache is a candidate if this becomes a hot path.
- The deferred isosurface `TooLarge` test (from #369) still wants a home; not exercised here (needs a >4 M-triangle grid).

Part of epic #346 (#362 follow-up).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -315,23 +315,27 @@ into a glTF `.glb` triangle mesh.
   origin }>` — `Some` ⇒ the mesh products are available and the origin is present
   (tileset built without sampling; "supports but no origin" is unrepresentable,
   never a 500). It drives the collection JSON's `representations` array → the
-  viewer's toggle. The isosurface seals (`background=Some(-32)`); echo-top uses
-  full `[512,360,64]` resolution (`ECHO_TOP_DIMS`, ~12 s first compute — ETag
-  caches repeats) coloured by a height ramp. A too-low threshold → 400, an empty
-  result → 404.
+  viewer's toggle. The isosurface seals (`background=Some(-32)`); echo-top is
+  coloured by a height ramp. Both mesh products take a `?resolution=` detail tier
+  (`Resolution` enum → voxel-grid dims; `low` `[128,360,48]` ≈2.2 M cells, `med`
+  `[256,360,56]` ≈5.2 M *default*, `high` `[512,360,64]` ≈11.8 M / ~12 s first
+  compute — ETag caches repeats); the tier is echoed into the `content.uri`. The
+  point cloud is native-resolution and ignores it. A too-low threshold or a bad
+  resolution token → 400, an empty result → 404.
 - **Point sizing:** `encode_pnts` writes a per-point `value` (the physical value)
   into the `.pnts` **batch table** (no `BATCH_ID` ⇒ per-point properties), so the
   viewer styles `pointSize` by `${value}` — weak echo → ~1 px, strong → ~10 px.
 - **Routes** (mounted at `/3dtiles`): `GET /collections/{id}/tileset.json`
-  (`?representation=&quantity=&datetime=&min_value=&threshold=`),
+  (`?representation=&quantity=&datetime=&min_value=&threshold=&resolution=`),
   `GET /collections/{id}/content.pnts` (`?quantity=&datetime=&min_value=`),
-  `GET /collections/{id}/content.glb` (`?representation=&quantity=&datetime=&threshold=`),
+  `GET /collections/{id}/content.glb` (`?representation=&quantity=&datetime=&threshold=&resolution=`),
   plus `/` · `/collections` · `/collections/{id}` · **`/viewer`** (a bundled
-  CesiumJS page with collection + quantity + **representation** pickers,
-  `include_str!`-baked from `crates/api-3dtiles/viewer/index.html`; same-origin
-  API base by default, with a `?base=` override). The tileset's `content.uri`
-  embeds the resolved quantity (+ pinned time + `min_value`/`threshold` +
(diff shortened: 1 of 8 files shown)


mrauhala:
Receipts for the latest review round (on `20aecf7`), now in `aecfcb0`:

- **`ds-3dtiles/src/lib.rs` byte-budget `+64`** → bumped to `+128` with an honest comment (it's a conservative ballpark for HEADER + the two JSON blobs + section alignments; the authoritative overflow guard remains the post-build `u32::try_from(total)`). No functional change.
- **`content.glb` absent→isosurface coupling** → `get_tileset` now emits `&representation=isosurface` explicitly for the isosurface tileset too (not just `echotop`), so the `content.uri` is unambiguous and no longer relies on the handler default. The handler keeps the absent→isosurface fallback for direct callers. Test updated.

**Deferring the legend title/unit finding (`handlers.rs` `collection_doc`) to #350**, with reasoning: v1 uses **one shared reflectivity colormap for every quantity** (documented limitation, #350). So for a velocity layer the *points themselves* are already coloured by the dBZ reflectivity ramp — the legend `"Reflectivity (dBZ)"` accurately describes the colormap that is actually applied, and is consistent with what's rendered. Deriving a per-quantity unit (e.g. `m/s` for `VRADH`) while the colours stay the reflectivity ramp would be *more* misleading, not less. Per-quantity colormaps **and** their matching legends land together with #350.

Note: several inline comments re-anchored to `20aecf7` with earlier timestamps (11:4x–12:14) are re-cites of findings already fixed in `2ee8585`/`20aecf7` (`{product}` message, pnts v2 assert, echo-top resolution assert, `legend_stops` n==1 guard, dropped `medium` alias).

mrauhala:
Receipts for the review round on `aecfcb0`, now in `2349b4f`:

- **`legend_stops` per-request allocation** → the legend is built once in a `LazyLock` (`POINT_LEGEND`); `GET /collections/{id}` now references it (a metadata accessor, O(1) per #211). Consolidated the colormap into `default_point_colormap()` (was duplicated inline at all 4 `TilesState3d` sites) so the encoded points and advertised legend share one source; admin/preview use it.
- **echo-top colormap rebuilt per request** → now a `LazyLock` (`ECHO_TOP_COLORMAP`), built once instead of inside `spawn_blocking` on every echo-top GLB request.
- **missing echo-top link assertion** → `collections_list_and_doc` now asserts the echo-top tileset link too.

⚠️ **Wire-format note for integrators (informational):** as of this PR, `.pnts` responses carry **non-zero** `batchTableJSONByteLength` / `batchTableBinaryByteLength` (previously always `0`) — the per-point `value` that drives point sizing + client-side filtering. Spec-compliant clients (CesiumJS 1.124 verified) handle this; any downstream consumer that snapshot-tests the raw `.pnts` header bytes or asserts the old zero-length batch-table fields must update its byte-level expectations.

mrauhala:
Receipts for the review round on `2349b4f`, now in `7b023a8`:

- **`?resolution=` only validated in the mesh arm** → `get_tileset` now parses it up front (before the representation match), so a bad tier is a 400 for *any* representation; `?representation=points&resolution=typo` no longer slips through as 200. Added a test.
- **Isosurface default `[256,360,56]` vs the engine's old `None`** → confirmed **intentional** and documented on the `Resolution::Med` variant: adding the tiers deliberately bumped the no-tier-specified isosurface/echo-top quality from the engine's historical `[128,360,48]` (= the `Low` tier) to `Med`, so a pre-tier bookmarked content URI renders one step sharper, not differently-wrong.
- **Reflectivity-specific echo-top threshold/floor** → noted on `ECHO_TOP_DEFAULT_THRESHOLD` (it joins the isosurface path + the no-echo-floor guard, which already reference #350); per-quantity floors land with per-quantity colormaps (#350).

PR #372, opened 2026-06-09
feat(3dtiles): time-dynamic playback — times manifest + viewer scrubber (#350)

### LINKED ISSUE #350: 3D Tiles Phase 4: time-dynamic tilesets (per-timestep animation)

mrauhala:
Closes #350 (3D Tiles Phase 4: time-dynamic tilesets). Part of epic #346.

## What

Animate radar volumes over time. Per-timestep tilesets already worked via `?datetime=<volume time>` (the engine's `select_entry_and_quantity` picks the volume nearest the requested time; `None` ⇒ latest), so this fills the two gaps from the issue:

### API — a `times` manifest
The collection JSON (`GET /3dtiles/collections/{id}`) now advertises `times`: `VolumeInfo.times` formatted RFC 3339 `…Z`, **sorted ascending + deduped**. Each value round-trips straight back as `?datetime=` on `tileset.json` / `content.*` — the time axis a client scrubs.

### Viewer — a time scrubber with no-stall playback
A play/pause + slider time bar, shown only when a collection has **>1** volume. It **preloads one hidden tileset per timestamp** and animates by toggling `.show`:
- `preloadWhenHidden: true` is **load-bearing** — without it a hidden Cesium tileset fetches no tiles, so the first reveal of each frame would stall. With it, every frame's content streams in the background and showing it is an instant visibility flip.
- Scrub/play is therefore a **pure visibility toggle — zero network per frame**.
- The point-cloud style (reflectivity size + client-side `min dBZ` filter) is applied across **all** frames so the whole loop is consistent.
- Refactored the single-`current`-tileset model into a `frames[]` array; a non-time-dynamic collection is the degenerate 1-frame case (time bar hidden).

## Verification (live, against `radar-fmi-vihti-h5`)

The committed 3-volume fivih sequence (`202605150000/0005/0010_fivih_PVOL.h5`):
- `times` manifest → `["2026-05-15T00:00:00Z","…00:05:00Z","…00:10:00Z"]`
- each `?datetime=` serves a **distinct** volume (different bytes + ETags)
- viewer preloads all 3 frames, time bar appears, defaults to latest (3/3)
- scrubbing to frame 0 → `frames[].show == [true,false,false]` (exactly one visible)
- play loops and advances; pause works; **`0` tileset/content fetches during scrub + play**

## Done when (#350)
- [x] Per-timestep tilesets (`?datetime=`; manifest exposed)
- [x] Manifest of available times (collection JSON `times`)
- [x] Looping client demo (the bundled `/3dtiles/viewer`)
- [x] No per-frame full-reload stalls — preload-all + `preloadWhenHidden`; a **sliding window** for much longer sequences is a noted follow-up

api-3dtiles 21 tests green (adds a times-manifest + ascending-sort assertion on an out-of-order mock), clippy clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -335,13 +335,27 @@ into a glTF `.glb` triangle mesh.
   carries a `legend` (the point colormap sampled into `#rrggbb` stops over a dBZ
   range via `legend_stops`) which the viewer renders as a gradient bar for the
   point cloud.
+- **Time-dynamic playback (#350, done):** per-timestep tilesets already work via
+  `?datetime=<volume time>` on tileset.json/content.* (the engine selects the
+  volume nearest the time; `None` ⇒ latest). The collection JSON advertises a
+  `times` manifest (`VolumeInfo.times` → RFC 3339 `…Z`, sorted ascending — each
+  value round-trips straight back as `?datetime=`). The viewer **preloads one
+  hidden tileset per timestamp** (`preloadWhenHidden: true` is load-bearing — a
+  hidden tileset otherwise fetches no tiles, so the first reveal would stall) and
+  animates by toggling `.show` — scrub/play is a pure visibility flip, **zero
+  network per frame** (verified). A play/pause + slider time bar shows only when
+  the collection has >1 volume; the point-cloud style (size + client-side filter)
+  is applied across all frames. A non-time-dynamic collection is the degenerate
+  1-frame case. Preloading-all suits the engine's bounded retained-volume count; a
+  sliding window for much longer sequences is a follow-up.
 - **Routes** (mounted at `/3dtiles`): `GET /collections/{id}/tileset.json`
   (`?representation=&quantity=&datetime=&min_value=&threshold=&resolution=`),
   `GET /collections/{id}/content.pnts` (`?quantity=&datetime=&min_value=`),
   `GET /collections/{id}/content.glb` (`?representation=&quantity=&datetime=&threshold=&resolution=`),
   plus `/` · `/collections` · `/collections/{id}` · **`/viewer`** (a bundled
   CesiumJS page with collection + quantity + **representation** + **resolution**
-  pickers, `include_str!`-baked from `crates/api-3dtiles/viewer/index.html`;
+  pickers and a **time scrubber** (play/pause + slider) for multi-volume
+  collections, `include_str!`-baked from `crates/api-3dtiles/viewer/index.html`;
   same-origin API base by default, with a `?base=` override). The tileset's
   `content.uri` embeds the resolved quantity (+ pinned time +
   `min_value`/`threshold` + `representation` + `resolution`) so the content fetch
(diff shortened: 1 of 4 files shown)


mrauhala:
Pushed two follow-ups since the first review:

- **`d43bc1c` — preload safety cap.** The `odim-volume` engine's `max_files` defaults to **unbounded**, so a real S3 feed can retain hundreds of volumes (a day ≈ 288 at 5-min cadence) and preloading all of them would mean hundreds of concurrent fetches + GPU-resident tilesets. The viewer now animates the most-recent **`MAX_FRAMES` (48)** of the manifest (`frameTimes = times.slice(-MAX_FRAMES)`); the status notes the window (e.g. "latest 48 of 288"). A true sliding window stays a follow-up; operators bound the source with `[odim] max_files`/`time_window`.
- **`9513eb5` — review fixes (all-frames-failed case).** Both `startPlaying` and `syncTimeBar` now also require `frames.some(Boolean)`, so when every frame's `fromUrl` fails (`frames = [null,…]`) the timebar hides and play won't start (no more dead slider / forever-looping-on-nothing). Thanks for catching it.

mrauhala:
Addressed the latest review round in `7dcd636`:

- **Race condition (loadFrames after `zoomTo`)** — fixed: rechecks the load token immediately after the async `zoomTo`, so a superseded call can't overwrite the live "loading…" status. Good catch.
- **Times manifest: redundant per-request sort** — fixed: `VolumeInfo.times` is already sorted-ascending + deduped by the engine (`volume_engine.rs:1252-1254` does `sort_unstable()` + `dedup()`), so `collection_doc` now emits it directly (no clone/re-sort — metadata accessor, #211), via `to_rfc3339_opts(SecondsFormat::Secs, true)`. Test mock now supplies ascending times to match the contract.
- **"CLAUDE.md violation: multi-line comment block"** — respectfully declining: that rule ("one short line max") is **not in this project's CLAUDE.md** (I grepped both the repo and the user-global CLAUDE.md — only a SQL-safety mention of multi-line `format!`). Multi-line comments for non-obvious invariants are the codebase norm (e.g. `engine.rs`/`volume_engine.rs` carry 5–7-line blocks), and the harness guidance is to match surrounding comment density. The `preloadWhenHidden`/`MAX_FRAMES` comments document load-bearing rationale, so I've kept them.

PR #373, opened 2026-06-09
feat(3dtiles): true cylindrical voxels — EXT_primitive_voxels volume rendering (#351)

mrauhala:
Part of epic #346. The **Tier-B** true-voxel track: radar polar volumes as a CesiumJS ray-marched **volume**, alongside the always-works `.pnts`/isosurface/echo-top products.

## What

A `VoxelGrid` → `EXT_primitive_voxels` glTF `.glb` + `3DTILES_content_voxels` / `3DTILES_bounding_volume_cylinder` tileset that CesiumJS 1.142 renders with a `VoxelPrimitive`. **Render-verified live** (screenshot in the thread): the reflectivity field ray-marches as a true 3-D volume over southern Finland with coherent precipitation structure and visible vertical depth.

### Commits
1. **CesiumJS 1.124 → 1.142** — voxels need WebGL2 (≥1.135) and the `EXT_primitive_voxels` realignment (1.127); also picks up point-cloud filtering perf, the circles→squares fix, and shader-metadata correctness. SRI recomputed; existing reps re-verified.
2. **Encoder** (`ds-3dtiles/voxels.rs`) — `encode_voxels_glb` + `tileset_json_voxels` + `voxel_subtree_json`.
3. **API** — `…/voxel/{tileset.json,subtrees/*,content/*}` routes; `VoxelGridCaps` gains `radius_m`/`height_m` so the bounding cylinder matches the grid without a sample.
4. **Viewer** — "Voxels" in the View dropdown via `Cesium3DTilesVoxelProvider` + a reflectivity transfer-function `CustomShader`.
5. **ENU transform fix** — found in render-verify (see below).

## Authoritative-source note

`EXT_primitive_voxels` is **not in the Khronos glTF registry** — it's a CesiumGS draft with **CesiumJS as the only implementation**. So I encoded against the live CesiumJS `VoxelCylinder3DTiles` **fixtures**, which caught that the draft README is **stale**: cylinder `mode` is `2147483650` (0x80000002), not the README's `2147483647`. The mode encodes the shape (box 0x80000000, ellipsoid 0x80000001).

## Load-bearing gotchas (all in `voxels.rs` doc comments)

- **Axis swap**: content `[radius, angle, height]` → glTF `dimensions [radius, height, angle]`; data laid out **radius-fastest → height → angle-slowest** (our `VoxelGrid` is the transpose, so the encoder reorders).
- `EXT_structural_metadata` (schema + `propertyAttributes`) at the glTF **top level**, not the primitive; embedded BIN buffer; `NaN` → a declared `noData` sentinel.
- **Implicit OCTREE tiling + a constant `.subtree`** is required (CesiumJS's voxel traversal); for one tile that's a constant-availability subtree.
- **The tile `transform` must be the real ENU→ECEF frame**, not identity-rotation — identity works for the mesh products (absolute-ECEF vertices) but tilts the *parametric* cylinder by the latitude (a ~60° slanted slab, which is exactly the first render). This was the one bug render-verify surfaced.

## Scope (v1 MVP)

Single tile, latest time. **Follow-ups** (tracked in #351): implicit octree LOD over multiple tiles, the national mosaic, time-dynamic voxels, per-quantity transfer functions. Tier-A `.pnts`/mesh remain the portable fallback for non-CesiumJS clients.

## Tests

`ds-3dtiles` (24) structural tests against the fixture shape; `api-3dtiles` (22) incl. a voxel tileset/subtree/content route test; `ds-core`/`engine-odim` green; clippy clean. Plus the live CesiumJS 1.142 render-verify.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -375,9 +375,14 @@ into a glTF `.glb` triangle mesh.
   (`VoxelCylinderShape`, verified in the 1.142 source) uses default angle bounds
   **-π..+π**, angle-index 0 → -π, angle **counter-clockwise from local +X**
   (=East via the ENU transform). The encoder remaps each output angle slot `s`
-  to the radar bin at compass bearing `90° - (-π + (s+0.5)/nA·2π)`
+  to the radar bin at compass bearing **`270° - φ`** where `φ = -π + (s+0.5)/nA·2π`
   (`grid_azimuth_index`); the mesh/point products skip this (they map azimuth →
-  ECEF directly). **Unmeasured (`NaN`) cells → the no-echo floor (-32 dBZ), NO
+  ECEF directly). **The `+180°` (270 not the 90 the stated convention implies) is
+  render-verified, not derived:** CesiumJS's effective cylinder-angle origin sits
+  opposite (-X) to where the -π bound + +X=East transform predict, so the naive
+  `90° - φ` leaves the volume 180°-rotated. ALWAYS render-verify a voxel-cylinder
+  azimuth mapping against the point cloud — the handedness/offset can't be trusted
+  from the spec alone. **Unmeasured (`NaN`) cells → the no-echo floor (-32 dBZ), NO
   `noData`** — an extreme sentinel trilinearly-interpolates into hard
   walls/floors at the data boundary; the floor (faded out by the transfer
   function) keeps boundaries soft, at the cost of a dense volume (no empty-space
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -305,17 +305,19 @@ into a glTF `.glb` triangle mesh.
   first cut rendered uniformly red). Demo: `cargo run -p engine-odim --example
   gen_echo_top` (render-verified). API route is a follow-up; the mesher will be
   reused for VIL (#365).
-- **Three representations are served from the API** (selected by
-  `?representation=points|isosurface|echotop` on `tileset.json`): `points` →
+- **Four representations are served from the API.** Three are selected by
+  `?representation=points|isosurface|echotop` on `tileset.json`: `points` →
   `.pnts` (region-only tileset, `RTC_CENTER` self-places), `isosurface` /
   `echotop` → `.glb` plus a tile **`transform`** = the antenna ECEF. The two glTF
   products share `content.glb`, disambiguated by `?representation=echotop` in the
-  content URI (default = isosurface). Capability + the origin both glTF products
-  need are coupled in one field: `VolumeInfo.voxel_grid: Option<VoxelGridCaps {
-  origin }>` — `Some` ⇒ the mesh products are available and the origin is present
-  (tileset built without sampling; "supports but no origin" is unrepresentable,
-  never a 500). It drives the collection JSON's `representations` array → the
-  viewer's toggle. The isosurface seals (`background=Some(-32)`); echo-top is
+  content URI (default = isosurface). The fourth — **`voxels`** (#351) — has its
+  own `…/voxel/` sub-path (below), not a `?representation=` variant. Capability +
+  the origin/extents all of them need are coupled in one field:
+  `VolumeInfo.voxel_grid: Option<VoxelGridCaps { origin, radius_m, height_m }>` —
+  `Some` ⇒ the mesh + voxel products are available and the cylinder origin/extents
+  are present (tileset built without sampling; "supports but no origin" is
+  unrepresentable, never a 500). It drives the collection JSON's `representations`
+  array → the viewer's toggle. The isosurface seals (`background=Some(-32)`); echo-top is
   coloured by a height ramp. Both mesh products take a `?resolution=` detail tier
   (`Resolution` enum → voxel-grid dims; `low` `[128,360,48]` ≈2.2 M cells, `med`
   `[256,360,56]` ≈5.2 M *default*, `high` `[512,360,64]` ≈11.8 M / ~12 s first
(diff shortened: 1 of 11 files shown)


mrauhala:
Addressed all four review findings in `8741f76`:
- **Subtree handler status/guard consistency** — `get_voxel_subtree` now uses the same gate as the tileset/content handlers (no voxel grid → 400, no coverage → 404), so the three voxel routes agree.
- **Unchecked `count * 4`** — now a checked multiply before the u32 bound.
- **GLSL `quantity` interpolation** — guarded with `^[A-Z][A-Z0-9_]*$` before building the shader.
- **`geometricError: 0`** — commented why it's correct (single-tile implicit voxel set, nothing to refine to), so it isn't 'fixed' to match the mesh `>0` rule.

mrauhala:
## Review findings addressed — `b8b9f61`

All five inline findings from the latest review round are fixed:

| # | Finding | Resolution |
|---|---------|------------|
| 1 | `contentAvailability` must be an array (3D Tiles 1.1) | Subtree JSON now emits `[{ "constant": 1, "availableCount": 1 }]`; integration assertion → `[0]["constant"]` |
| 2 | Missing ETag on `get_voxel_tileset` | Routes through `binary_response` (strong ETag + 304), matching `content.pnts`/`content.glb` |
| 3 | `expect` panics in `Result`-returning fns | New `Tiles3dError::Serialize`; `assemble_glb` + `tileset_json_voxels` use `.map_err` |
| 4 | Missing `content_uri`/`subtree_uri` validation | Extracted `validate_uri`; applied to both URIs in `tileset_json_voxels` (+ rejection test) |
| 5 | `"voxels"` advertised when `radius_m = 0` | `collection_doc` now gates voxels on `supports_voxels` (`radius_m > 0 && height_m > 0`), matching the tileset route; mesh products stay on `supports_mesh` |

**Note on #5's suggested test:** pinning the `radius_m = 0` → 404 path needs a mock `VolumeEngine` registry (heavier scaffolding than this MVP edge case warrants); the gate fix mirrors `get_voxel_tileset` exactly, so advertisement and tileset availability can no longer diverge.

ds-3dtiles 25 unit tests + api-3dtiles 22 integration tests green; clippy clean.

mrauhala:
## Second review round addressed — `06cc9f2`

The `get_voxel_subtree` round (06:48) findings:

| Finding | Resolution |
|---------|------------|
| Missing ETag / conditional-GET on `get_voxel_subtree` | Now routes through `binary_response` (constant body → constant ETag → 304 on `If-None-Match`), like the tileset/content handlers |
| Tile coordinates silently ignored (any `/voxel/subtrees/*` & `/voxel/content/*` path returns the root's constant data with 200) | `is_root_voxel_tile` rejects anything but `0/0/0/0` with 404 (single-tile implicit set). Unit test (`only_the_root_voxel_tile_is_accepted`) + integration 404 assertions for both routes |

Plus closing the prior round's **#5** suggested test: `MockZeroRadiusVoxel` (voxel grid present, `radius_m = 0`) pins that the collection doc omits `"voxels"` (mesh products still advertised) **and** that `voxel/tileset.json` 404s — advertisement and availability can't diverge.

api-3dtiles 23 integration + 2 unit tests green; clippy clean.

mrauhala:
## Third review round addressed — `15f9c99`

| Finding | Resolution |
|---------|------------|
| `statistics.min`/`max` must be bare scalars for a SCALAR property (EXT_structural_metadata) | `tileset_json_voxels` now emits `"min": x, "max": y` (was `[x]`/`[y]`); unit test → `json!(70.0)` |
| `get_voxel_content` skips the coverage guard, burning a render slot + `spawn_blocking` before a 404 | Now checks `radius_m > 0 && height_m > 0` up front (404), matching the tileset/subtree handlers — all three agree |
| Voxel time scrubber shown-but-inert on multi-volume collections | Voxels branch pins `frameTimes = [null]` (latest volume) so `syncTimeBar()` keeps the bar hidden — matches the MVP intent (voxel time-dynamic playback is a #351 follow-up) |

ds-3dtiles 25 + api-3dtiles 23 integration + 2 unit tests green; clippy clean. The summary also confirmed all prior-round findings (ETag, `validate_uri`, `Serialize`, `is_root_voxel_tile`, `MockZeroRadiusVoxel`, `supports_voxels`) are resolved.

mrauhala:
## Fourth round addressed — `66a038a`

The "approved with minor findings" round:

| Finding | Resolution |
|---------|------------|
| Medium (robustness): `count`/`grid.values` could disagree → OOB panic in `spawn_blocking` | `debug_assert_eq!(grid.values.len(), count)` before the transpose (zero release cost) |
| Low (perf): `smooth_grid` radius sweep had `r` (largest stride) innermost → cache misses | Reordered `r` outermost / `h` (stride 1) innermost — identical result, cache-friendly |
| Low (cosmetic): `validate_uri` scanned `..` over the whole URI, missing `a/..?t=1` | Isolates the path (drops `?query`/`#fragment`) before the segment scan; allows a harmless `..` in a query value. Test added |

ds-3dtiles 26 + api-3dtiles 23 integration + 2 unit tests green; clippy clean.

mrauhala:
## Fifth round addressed — `a077482`

The "well-executed MVP" round:

| Finding | Resolution |
|---------|------------|
| `assemble_glb` is the third (and best) glb-assembly copy | Promoted to `pub(crate) assemble_glb`; `isosurface`/`echo_top` `build_glb` now use it (both return `Result`) — they inherit the `u32::try_from` overflow guard + `Serialize` error path, dropping their `total as u32` casts and `expect` panics. Both mesh BINs are always 4-aligned, so output is **byte-identical** (all wellformed tests pass) |
| `3DTILES_content_voxels.dimensions` order undocumented | Added a comment: content/radar order `[n_r,n_a,n_h]` here vs the glb's axis-swapped `[n_r,n_h,n_a]`; CesiumJS uses only the glb field |
| `voxelMin()` fallback hardcodes `5` | Now `parseFloat(REP_META.voxels.numDefault)` — single source of truth |

ds-3dtiles 26 + api-3dtiles 23 integration + 2 unit tests green; clippy clean.

mrauhala:
## Sixth round addressed — `b7fdab6`

| Finding | Resolution |
|---------|------------|
| Bug: non-finite legend `min`/`max` → `"NaN"`/`"Infinity"` in GLSL → dark volume | `voxelShader` fallback now also requires `Number.isFinite(min) && Number.isFinite(max)` |
| `high`-res voxel (~12 s) held the shared render semaphore, blocking raster renders | Voxel content moved to a dedicated process-global `VOXEL_SEMAPHORE` (¼ cores ≥1); it no longer touches the raster pool. Mesh products (<1 s) stay shared |
| No test for the load-bearing 270° azimuth offset | `azimuth_remap_pins_the_render_verified_270_offset`: pins `grid_azimuth_index` at 4 points (`(449−slot) mod 360` for n_a=360) + asserts a bijection over all 360 bins |

Note: the actual mapping values are 89/359/269/179 (not the 90/180 in the inline suggestion — those were approximations; the test pins the render-verified values).

ds-3dtiles 27 + api-3dtiles 23 integration + 2 unit tests green; clippy clean.

mrauhala:
## Seventh round addressed — `86469d1`

| Finding | Resolution |
|---------|------------|
| Perf: `smooth_grid` angle sweep had `a` (stride n_h) innermost | Reordered to `h` (stride 1) innermost, neighbour indices hoisted — consistent with the height/radius sweeps; byte-identical |
| Consistency: `voxel_subtree_json` `expect`-panicked on serialize | Now returns `Result<String, Tiles3dError>` via `map_err` (handler maps to a contextual 500) — aligns with the `Serialize` variant this PR added |
| Cosmetic: Cyrillic `Т` (U+0422) in 6 `glTF` doc comments | Replaced all with Latin `T` |

ds-3dtiles 27 + api-3dtiles 23 integration + 2 unit tests green; clippy clean.

Issue #293, opened 2026-06-02
engine-odim: PVOL lazy cache re-downloads the whole .h5 per moment — batch-decode on miss


mrauhala:
## Problem

The lazy PVOL pixel cache (`PIXEL_CACHE`, #289 / PR #290) is keyed per
`(file_id, dataset_path)` — one entry per **moment**. But `Pixels::moment`
fetches and parses the **whole `.h5` file** to extract that one dataset (the
pure-Rust `hdf5-reader` has no slice / byte-range API; `read_moment_pixels`
parses the entire buffer).

So for a polar volume with `Q` cold-cached moments, the file is downloaded and
parsed `Q` times. At ~27 MB/file and ~4 typical quantities that is ~108 MB of
S3 transfer per cold timestep-file; a time-range trajectory / profile query
over `N` files multiplies it by `N × Q`.

(Flagged in PR #290 review round 3. Local sources are cheap — the re-read hits
the OS page cache — so this is specifically an **S3 / remote** cost.)

## Remedy

On the **first** miss for a file, fetch it once and **batch-decode all moments
of all sweeps** (for the requested quantities) into `PIXEL_CACHE`, so sibling
moment lookups for the same file hit the cache. The metadata-only volume in the
catalog already enumerates every `(sweep, moment)` dataset path, so the batch
set is known.

Prefer batch-decode over caching the raw 27 MB bytes per file — caching raw +
decoded would double the resident memory and undercut the bounded-RAM goal.

## Notes
- Keep the byte-bounded eviction; batch-decoding a file just primes multiple
  entries at once.
- The negative cache (PR #290 r2) already prevents *failed* moments from
  re-downloading; this issue is the **success** path.

mrauhala:
The 3D Tiles hot-path audit (PR #378) confirmed this is the dominant remaining S3 cost: `Pixels::moment` re-downloads the **whole** `.h5` once per cold moment (Q moments → Q full-object GETs, ~tens of MB each), and a 48-frame viewer preload touches 48 files. With the #378 content + voxel-grid caches the recompute side is fixed, so first-touch S3 transfer is now the long pole for cold animation loads. Batch-decoding all moments of a file on the first miss (this issue) is the highest-leverage S3 fix left.

PR #378, opened 2026-06-10
perf(3dtiles): content + voxel-grid caches — cached repeats 165× faster

mrauhala:
## Summary

A hot-path audit of the 3D Tiles APIs found that **every** content request — including `If-None-Match` revalidations that end in a 304 — paid the full engine read → polar resample → encode pipeline, and the bundled viewer multiplies that by up to 48 preloaded animation frames on every load, reload, and control change. This PR adds the two cache layers the audit ranked highest (P1 + P2):

### 1. Encoded-content cache + single-flight (`api-3dtiles/src/cache.rs`)
- Process-global byte-bounded LRU of the final `.pnts`/`.glb`/voxel-`.glb` bytes + strong ETag.
- Key = (collection, product, quantity, datetime, params, dims) **+ a data-version** hashed from `VolumeInfo.times` — ingesting/dropping a volume invalidates "latest" and nearest-time selections naturally, with no duplication of engine selection logic at the API layer.
- **Single-flight:** concurrent identical requests share one compute via a per-key async mutex; only the computing request acquires the render/voxel semaphore, so a 48-frame viewer preload no longer queues 48-deep on the shared semaphore (which also starved WMS).
- A 304 revalidation is now two cache lookups, not a recompute.
- `MC_3DTILES_CONTENT_CACHE_MB` (default 512, `0` disables).

### 2. Shared voxel-grid cache + column-resolved resampler (`engine-odim`)
- `VolumeEngine::read_voxel_grid` now returns **`Arc<VoxelGrid>`** served from a global LRU keyed (file, quantity, dims) — the isosurface, echo-top, and voxel products (and threshold changes, which only re-mesh) share one multi-million-cell resample. `MC_PVOL_VOXEL_GRID_CACHE_MB` (default 512).
- The resample now resolves each `(radius, height)` column **once** (`ColumnTarget`): the sweep choice, envelope/range guards, moment lookup, range bin, and pixel-cache lookup (which heap-allocated two `Arc<str>` per cell) are all azimuth-independent. Per-cell work becomes a table index + `sample_class` — at `med` that's 14 k resolutions instead of 5.2 M, guard-for-guard identical to `sample_polar_slant_class`.

### Cache-Control policy
Content/tilesets pinned to an explicit `?datetime=` are archived volumes → `public, max-age=86400, immutable` (the existing WMS explicit-`TIME` policy). The viewer pins every animation frame, so a page reload serves all frames straight from the browser cache. "Latest" keeps `max-age=60`.

### Observability
New `/metrics` series: `tiles3d_content_cache_{hits,misses,bytes,capacity_bytes}` and `pvol_voxel_grid_cache_{hits,misses,bytes,capacity_bytes}`, emitted only when the relevant collections are loaded.

## Smoke test (local fivih PVOL fixture, debug build)

| Request | Before-equivalent (cold) | Cached |
|---|---|---|
| isosurface `med` `.glb` (35 MB) | 4.93 s | **0.030 s** |
| echo-top `med` `.glb` after isosurface | ~4.9 s | **0.28 s** (shares the cached grid, only re-meshes) |
| 304 revalidation | ~4.9 s | **0.0008 s** |
| `.pnts` | 0.53 s | **0.005 s** |

Repeat bytes verified identical; `immutable` vs `max-age=60` headers verified; both cache metric series live.

## Tests
- New unit tests for the content cache (hit/miss, version invalidation, concurrent coalescing to one compute, error-not-cached + gate release, capacity-0 disable).
- New integration tests: repeated `.pnts` request hits the cache without touching the engine (counting mock), `.glb` 304 without recompute, pinned-vs-latest `Cache-Control`.
- `cargo test --workspace`, `cargo clippy --workspace --tests -- -D warnings`, `cargo fmt` all clean.

Follow-ups (from the same audit, not in this PR): viewer progressive preload + per-representation tileset reuse, dedicated points/mesh semaphore, #293 batch moment decode (S3 Q× transfer), isosurface field smoothing + smooth normals (visual quality).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -428,10 +428,11 @@ into a glTF `.glb` triangle mesh.
   changes share one polar resample (`MC_PVOL_VOXEL_GRID_CACHE_MB`, default 512);
   the resample itself resolves each `(radius, height)` column once
   (`ColumnTarget`) instead of per-cell sweep/moment/pixel-cache lookups.
-  **`Cache-Control`:** content/tilesets pinned to an explicit `?datetime=` get
-  `max-age=86400, immutable` (the viewer pins every animation frame, so reloads
-  re-use the browser cache); "latest" keeps `max-age=60`. A 304 revalidation
-  costs two cache lookups, not a recompute. Cache metrics are in `/metrics`
+  **`Cache-Control`:** content/tilesets whose `?datetime=` **exactly matches an
+  advertised volume time** get `max-age=86400, immutable` (the viewer pins every
+  animation frame from the `times` manifest, so reloads re-use the browser
+  cache); a between-volumes datetime (nearest-selection can change) and "latest"
+  keep `max-age=60`. A 304 revalidation costs two cache lookups, not a recompute. Cache metrics are in `/metrics`
   (`tiles3d_content_cache_*`, `pvol_voxel_grid_cache_*`). `content_uri` is
   validated (no `..`/absolute/scheme) in `ds-3dtiles`.
 - **Config:** add `"3dtiles"` to a collection's `apis` (only `odim-volume`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -414,8 +414,25 @@ into a glTF `.glb` triangle mesh.
   handlers bound them with the shared render semaphore and run via
   `spawn_blocking` — never inline on a request worker (the same pattern the
   raster APIs use for `get_raster_tile`).
-- **Caching:** content-derived ETag on both `.pnts` and `.glb` (deterministic
-  bytes ⇒ cheap 304s, shared `binary_response` helper); `content_uri` is
+- **Caching (two layers, hot-path audit 2026-06):** (1) `api-3dtiles/src/cache.rs`
+  — a process-global byte-bounded LRU of the **encoded content bytes + ETag**
+  (`.pnts`/`.glb`/voxel `.glb`), keyed by (collection, product, quantity,
+  datetime, params, dims) **plus a data-version hashed from `VolumeInfo.times`**
+  (new volume ⇒ new version, so "latest" and nearest-time selection invalidate
+  without duplicating engine selection logic), with per-key **single-flight**
+  coalescing (concurrent identical requests share one compute; the semaphore is
+  only acquired by the computing request). `MC_3DTILES_CONTENT_CACHE_MB`
+  (default 512, 0 disables). (2) engine-side `VOXEL_GRID_CACHE` in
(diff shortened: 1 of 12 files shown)


mrauhala:
Review receipts (all three findings addressed in 9db5b3f):

1. **`immutable` for nearest-time selection** — fixed. `cache_control(pinned)` now takes `pinned = exact_volume_time(&info, time)` (binary search over the sorted `VolumeInfo.times`): only a `?datetime=` that exactly matches an advertised volume time gets `max-age=86400, immutable`; a between-volumes datetime (nearest-selection can change) falls back to `max-age=60`. Applied to all three content routes and both tileset routes. New test `non_manifest_datetime_is_not_immutable` covers it.

2. **`VOXEL_GRID_CACHE` single-flight** — fixed properly rather than just re-qualifying the claim: reads now go through `quick_cache`'s `get_or_insert_with`, whose placeholder guard runs the resample for exactly one caller per key and blocks concurrent callers until it finishes (same mechanism `PixelCache::mark_bad` already relies on). Concurrent first-time isosurface + echo-top + voxel requests for the same `(volume, quantity, dims)` now share one resample. Blocking the waiter thread is safe — every caller is on a `spawn_blocking` pool by contract (noted in the comment).

3. **Stale grids after volume retirement** — documented on `DEFAULT_VOXEL_GRID_CACHE_MB`: entries for volumes retired by `max_files`/`time_window` are reclaimed by LRU pressure only (no data-version in the key since volume files are immutable); operators should bound source retention so retired grids cycle out.

`cargo test -p api-3dtiles -p engine-odim`, clippy `-D warnings`, fmt all clean.

mrauhala:
Review r2 receipts (9beb50b):

1. **`data_version` over-invalidation of pinned-time entries** — fixed, without the trait/interface change: an **exact advertised volume time** already fully determines the bytes (nearest-selection of a zero-distance match is that volume, and the file is immutable once scanned), so exact-time keys now get a constant `content_version` while "latest"/between-volume keys keep the full time-axis hash. New volume arrivals no longer evict the preloaded animation window. Retirement is still safe: once the time drops from `VolumeInfo.times`, the exactness check flips false and the same request re-keys to `data_version` — a retired time can't serve stale bytes. Regression test `exact_time_entry_survives_new_volume_arrival` asserts a pinned frame survives a manifest change (engine read count unchanged) while "latest" re-keys and recomputes.

2. **Capacity-0 doc wording** — fixed, and made more precise than suggested: at capacity 0 the per-key gate only **serializes** duplicates (each waiter re-checks the cache under the gate, misses because nothing is admitted, and recomputes in turn) — it bounds the stampede to one compute at a time but does not share results, since the in-flight winner's result isn't retained for the re-check.

Issue #360, opened 2026-06-09
engine-odim: distinguish radar undetect (clear air) from nodata (unmeasured) in the voxel grid


mrauhala:
## Context
`RawPixels::sample` (`crates/engine-odim/src/reader.rs`) masks **both** `nodata` ("outside coverage" — cone of silence / beyond range) and `undetect` ("the radar looked and saw nothing" — clear air) by returning `None`. In the cylindrical `VoxelGrid` (`read_voxel_grid`, #358) both therefore become `NaN`, so a clear-air cell is indistinguishable from a truly-unmeasured one.

This surfaced while building the 3D Tiles isosurface (#357): the surface couldn't close where echo meets clear air, rendering open vertical "curtains" instead of solid blobs. PR #359 worked around it in the **encoder** with a `background` seal (treat all `NaN` as no-echo), which closes the blobs but also caps the *real* cone of silence (a minor, unavoidable artifact given the conflation).

## Proposal
Plumb the `undetect` vs `nodata` distinction through the voxel sampling path so the grid can represent them differently:
- **`undetect` (clear air, in-coverage)** → a finite "below-detection floor" (e.g. the quantity's min), so consumers treat it as a real low value.
- **`nodata` / out-of-coverage / out-of-envelope** → stay `NaN` (genuinely unknown).

Then the isosurface (and future true voxels, #351) can seal against clear air **without** capping the actual cone of silence — only the genuinely-unmeasured cone stays open.

### Sketch
- `reader.rs`: add a classified sampler returning `Value(f64) | Undetect | Masked` (reimplement `sample()` on top, no behaviour change for raster/EDR callers).
- `volume_engine.rs` `voxel_grid_from_volume` / `sample_polar_slant`: fill `Undetect` cells with the floor, `Masked`/out-of-envelope with `NaN`.
- Benefits both the isosurface (#357) and the voxel path (#351).

Refs #357 #358 #351

PR #388, opened 2026-06-10
feat(3dtiles): smooth the scalar field before isosurface marching (#381)

### LINKED ISSUE #381: 3D Tiles isosurface: smooth the scalar field before marching (reuse the voxel blur)
From the 3D Tiles visual audit (epic #346). The user-visible complaint: the isosurface shell looks blocky/aliased.

The voxel representation dissolves the cellular radar field with a 4-pass separable blur before encoding (`smooth_grid`, `ds-3dtiles/src/voxels.rs` ~202 — sigma ≈ 1.4 cells), but the isosurface mesher (`ds-3dtiles/src/isosurface.rs`, `encode_isosurface_glb`) marches the **raw** grid. Radar echo is cellular (each cell a local max), so the extracted shell inherits the cell lattice: stair-step artifacts along the contour, especially at range where 1° azimuth bins are km-wide.

Proposed: apply 1-2 passes of the same `[0.25, 0.5, 0.25]` separable kernel (angle periodic, radius/height clamped) to a copy of the grid values before marching — factor `smooth_grid` out of `voxels.rs` with a pass-count parameter. NaN handling needs care: blur only across finite cells (or after the no-echo-floor sealing decision) so the smoothing doesn't erode the `Masked`/open-boundary semantics (#360).

Render-verify against the fivih fixture (`cargo run -p engine-odim --example gen_isosurface`). Note the grid resample is now cached engine-side (#378), so this re-mesh-only change stays cheap to iterate on.

mrauhala:
## Summary

From the 3D Tiles visual audit (epic #346): the isosurface shell looked blocky/aliased because the mesher marched the **raw** voxel grid while the voxel representation already dissolves the cellular radar field with a multi-pass separable blur. Radar echo is cellular (each cell a local max), so the extracted shell inherited the cell lattice — stair-steps along the contour, worst at range where 1° azimuth bins are km-wide.

- **Shared `smoothing` module** (`ds-3dtiles/src/smoothing.rs`): the `[0.25, 0.5, 0.25]` separable sweep factored out of `voxels.rs`, with a pass-count parameter. The generic sweep is monomorphized per kernel, so the voxel path (4 passes) keeps its cache-tuned loop orders unchanged.
- **`encode_isosurface_glb` smooths before marching**: 2 passes (sigma ≈ 1 cell) — enough to round the contour without erasing storm-core structure (a mesh shows its lattice less than a ray-marched volume seen from inside).
- **NaN handling follows the sealing decision** (preserves the #360 open-boundary semantics):
  - `background = Some(bg)` (the API default): seal NaN → bg **first**, then the dense blur — the same order the voxel path uses.
  - `background = None`: a NaN-aware blur keeps unmeasured cells `NaN` (still skipped) and renormalizes kernel weights over finite neighbours — never bleeds the boundary into real echo, never fabricates values inside the unmeasured region.

## Test updates

- New unit tests for the smoothing module: linear-field invariance away from clamped ends, periodic angle wrap, NaN preservation/non-bleed, zero-pass identity.
- `isosurface_sits_at_the_expected_height` moved its crossing to the grid interior (n_h 5 → 9): the kernel preserves a linear ramp exactly, but the clamped-boundary perturbation travels one cell inward per pass, so the assertion now pins the *smoothed* mesher to the same exact height.
- The sealed-blob test core grew 2×2×2 → 4×4×4 (a 2-cell core smooths below threshold against the −32 floor; physical echoes span many cells).

## Render verification

`cargo run -p engine-odim --example gen_isosurface` against the fivih fixture, viewed in CesiumJS: the 20 dBZ shell is smooth and organic with no visible cell lattice, including at range (the far blobs ~100+ km from the antenna), sitting correctly over southern Finland.

Closes #381

### DIFF
--- a/crates/api-3dtiles/tests/integration.rs
+++ b/crates/api-3dtiles/tests/integration.rs
@@ -72,16 +72,20 @@ impl VolumeEngine for MockVolume {
                 return Err(DataServerError::InvalidParameter(format!("unknown {q}")));
             }
         }
-        let dims = dims.unwrap_or([4, 8, 4]);
+        let dims = dims.unwrap_or([16, 16, 16]);
         let [n_r, n_a, n_h] = dims;
         // Clear air is the finite no-echo floor (matching the post-#360 engine
-        // fill), with a finite >threshold echo core in the middle — so the
-        // isosurface (which the handler seals with `background=Some(floor)`)
-        // produces a non-empty mesh.
+        // fill), with a finite >threshold echo core spanning the middle half of
+        // the radius/height axes (full angle ring) — so the isosurface (which
+        // the handler seals with `background=Some(floor)` and then smooths
+        // before marching, #381) produces a non-empty mesh: the core scales
+        // with the requested grid and its interior survives the blur against
+        // the floor, like a physical echo spanning many cells (a fixed
+        // 2-cell-wide ring would smooth entirely below the threshold).
         let mut values = vec![ds_core::volume::NO_ECHO_FLOOR_DBZ; n_r * n_a * n_h];
-        for i_r in 1..n_r.min(3) {
+        for i_r in n_r / 4..(n_r * 3 / 4) {
             for i_a in 0..n_a {
-                for i_h in 1..n_h.min(3) {
+                for i_h in n_h / 4..(n_h * 3 / 4) {
                     values[VoxelGrid::index_of(dims, i_r, i_a, i_h)] = 40.0;
                 }
             }
--- a/crates/ds-3dtiles/src/isosurface.rs
+++ b/crates/ds-3dtiles/src/isosurface.rs
@@ -344,6 +344,11 @@ fn march_tet(
 /// mesh shaded a single `color` (`[r, g, b, a]`, 0–255) — typically the
 /// colormap colour at `threshold`, i.e. "the 20 dBZ shell".
 ///
+/// The scalar field is lightly smoothed before marching (#381 — see
+/// `crate::smoothing`), so the shell follows the echo, not the cell lattice;
+/// expect extracted values to deviate from the raw grid by up to the
+/// cell-to-cell contrast near sharp gradients and clamped grid edges.
+///
 /// ## `background` — sealing the surface against clear air
 ///
 /// A radar volume stores **`NaN` for both** "the radar looked and saw nothing"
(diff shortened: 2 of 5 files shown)


mrauhala:
Review round 1 receipts (11488cc):

- **Non-finite `background` silently empties the GLB** — not reachable: `encode_isosurface_glb` already rejects a non-finite `background` at entry (`Tiles3dError::NonFinite("background")`, isosurface.rs:389-392 — a pre-existing guard, with the `Some(f64::INFINITY)` case covered by `non_finite_threshold_or_background_is_rejected`), so `bg as f32` in the seal step is always finite. Added the suggested `debug_assert!` in `smooth_grid` anyway as defence-in-depth for future callers.
- **`background = None` + NaN-corner skipping test gap** — added `open_surface_meshes_crossings_inside_finite_region_only`: a 40 dBZ core in a finite 0 dBZ shell surrounded by NaN meshes the internal crossing under `None` (NaN-aware smoothing keeps it intact; NaN-touching tets still skipped).
- **`smooth_grid` contract assert** — added (debug-only full scan).

Also in this round: 07ad522 fixes the CI integration-test failures — the mock volume's fixed 2-cell echo ring smoothed below threshold; it now scales with the grid.

mrauhala:
Review round 2 receipts (latest commit):

- **`bg as f32` unchecked** — the non-finite half was already rejected at entry (the round-1 receipt covers it); the genuinely new sliver — a *finite* f64 beyond f32 range casting to ±inf — is now rejected by the same guard (`bg.abs() > f32::MAX`), with a `Some(-1e39)` test case.
- **Implicit `SMOOTH_PASSES` margin in the open-surface test** — the "crossing ≥ SMOOTH_PASSES finite cells from any NaN" invariant is now stated explicitly in the test comment, with instructions to widen the shell if the pass count is bumped.

mrauhala:
Review round 3 receipts (latest commit):

- **Renormalization arithmetic untested** — added `nan_aware_renormalizes_weights_exactly`: one pass over a `[1,1,3]` column `[NaN, 10, 30]` pins the exact weighted averages (50/3 at the NaN boundary, 25 at the clamped end, NaN preserved). A uniform plateau couldn't catch a wrong weighting; this does.
- **`debug_assert` release-invisible** — keeping it debug-only, deliberately. `smooth_grid` is a `pub(crate)` helper with exactly two call sites, and both construct the dense field on the immediately preceding lines (the voxel no-echo fill, the isosurface seal — which now also rejects non-finite/f32-unrepresentable backgrounds at entry). A regression in either is caught by the unit + integration tests CI runs in debug mode. A release-mode finiteness scan would tax every encode on a hot, perf-tracked path (up to 32M cells) to guard a failure that can only come from a local edit in the same file — the assert + tests are the proportionate guard.

mrauhala:
Round 4 receipt: no blocking findings. The `SMOOTH_PASSES` symbolic-export observation is acknowledged and deferred — it's a private tuning constant with one consumer and one test margin note; exporting it would widen the module surface for a value that changes (if ever) together with the test by design. No code change.

PR #392, opened 2026-06-11
fix(odim): SMHI signed-i16 PVOL in 3D Tiles + raster

mrauhala:
## Summary

Swedish (SMHI) ODIM_H5 radar polar volumes failed across **3D Tiles / WMS / EDR** — a 404 with no error in the logs, the quantity menu populating but every content fetch empty. This fixes the full stack so SMHI sites render end-to-end.

Two stacked causes, plus two quality fixes:

### 1. SMHI moments are signed `i16`
SMHI stores DBZH (and friends) as **signed 16-bit** integers (`gain=0.01`, sentinels `nodata=-32768`/`undetect=-32767` — only representable as signed). The ODIM reader only handled `u8`/`u16`/`f64`, so every moment hit `UnsupportedPixelType`. Added a `RawPixels::I16` variant.

### 2. Upstream `hdf5-reader` B-tree bug (32-bit superblocks)
Once the i16 path was reachable it exposed an upstream bug: `hdf5-reader` read version-1 B-tree raw-data **chunk-key dimension offsets** as `size_of_offsets` bytes, but the HDF5 spec fixes them at **8 bytes** regardless. SMHI files use a 32-bit superblock (`size_of_offsets = 4`), so it mislocated the chunk address and failed to DEFLATE-decompress every chunk (`h5dump`/libhdf5 read the same files fine). Fixed in our fork and pinned via `[patch.crates-io]`:
- Upstream PR: roteiro-gis/netcdf-rust#54
- Bumps the engine to `hdf5-reader` 0.6 (API-compatible for our `from_bytes` usage).

### 3. dtype dispatch instead of type-probing
hdf5-reader 0.6 matches a typed `read_array::<T>()` on **element byte-size only**, not signedness — so a signed `i16` array reads *successfully* but bit-reinterpreted as `u16` (e.g. `-32767` → `32769`). The old "probe u8 → u16 → … in order" approach would silently pick `u16` and return garbage. New `reader::read_raw_pixels_2d` branches on the dataset's **actual** `Datatype`; both the COMP and PVOL paths share it.

### 4. Default quantity = reflectivity
The 3D Tiles / raster default was `quantities.first()` (alphabetical). For a dual-pol volume whose moments sort `CCORH < DBZH` (SMHI) that defaulted to a clutter-correction quality field — defined across the *whole* volume, near-zero values washing out on the dBZ colormap. Now prefers `DBZH`, then `TH`. Every quantity stays selectable.

## Testing
- `cargo test -p engine-odim` — 166 tests pass (DMI u8 COMP, FMI u16 PVOL, new SMHI i16 PVOL regression test that skips gracefully when the uncommitted fixture is absent).
- Server smoke test against the real SMHI fixture: `tileset.json` / `content.pnts` / `content.glb` / voxel `tileset.json` all 200 with valid payloads.
- Live render in the 3D Tiles viewer: real dBZ values (min −32, max 57.7, mean 10.2), proper cone of silence, precipitation structure resolves cleanly.

## Notes
- Ships the SMHI collection config (`collections.d/radar-se-volume-local-h5.toml`).
- The ~21 MB Kiruna test fixture is **not committed** (113 MB for the full set; matches the FMI PVOL fixture convention — tests skip when absent).
- `[patch.crates-io]` is interim; drop it once a `hdf5-reader` release > 0.6.0 carries the fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1714,9 +1714,8 @@ dependencies = [
 
 [[package]]
 name = "hdf5-reader"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d8840fbc5ba7a53641310daaa48dfbd2b37d9f5892269fe58d45cc16573c093e"
+version = "0.6.0"
+source = "git+https://github.com/mrauhala/netcdf-rust.git?rev=767add81eb12d0fd123f3550637fc001c2f40ff7#767add81eb12d0fd123f3550637fc001c2f40ff7"
 dependencies = [
  "flate2",
  "lru",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,3 +68,18 @@ debug = false
 # `cargo update` can't silently follow a moved branch tip.
 [patch.crates-io]
 icechunk = { git = "https://github.com/earth-mover/icechunk.git", rev = "5f56b13605745ea06ec6558bcd8ed4a8f5805a3e" }
+
+# Interim patch until upstream releases the fix.
+#
+# hdf5-reader's version 1 B-tree raw-data chunk-key parser read each
+# per-dimension chunk offset as `size_of_offsets` bytes, but the HDF5 format
+# spec fixes those key dimension offsets at 8 bytes regardless of the
+# superblock's offset size. On 32-bit-superblock files (`size_of_offsets = 4`)
+# — e.g. SMHI ODIM_H5 radar polar volumes — this misaligned the cursor, located
+# a garbage chunk address, and made every chunked-dataset read fail to
+# decompress ("corrupt deflate stream"), even though libhdf5/h5dump read the
+# same files fine. Fixed in our fork (read the dim offsets as fixed u64); PR
+# filed upstream (roteiro-gis/netcdf-rust). Pinned to an immutable `rev` so a
+# `cargo update` can't follow a moved branch. Remove once a crates.io release
+# > 0.6.0 carries the fix.
+hdf5-reader = { git = "https://github.com/mrauhala/netcdf-rust.git", rev = "767add81eb12d0fd123f3550637fc001c2f40ff7" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,7 +79,7 @@ icechunk = { git = "https://github.com/earth-mover/icechunk.git", rev = "5f56b13
 # a garbage chunk address, and made every chunked-dataset read fail to
 # decompress ("corrupt deflate stream"), even though libhdf5/h5dump read the
 # same files fine. Fixed in our fork (read the dim offsets as fixed u64); PR
-# filed upstream (roteiro-gis/netcdf-rust). Pinned to an immutable `rev` so a
+# filed upstream (roteiro-gis/netcdf-rust#54). Pinned to an immutable `rev` so a
 # `cargo update` can't follow a moved branch. Remove once a crates.io release
-# > 0.6.0 carries the fix.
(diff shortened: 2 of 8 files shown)


mrauhala:
Addressed both review findings in b6d23f8:

1. **Config FMI leftovers** — retexted `radar-se-volume-local-h5.toml` header, base-id/NOD example (`sekrn`/`sehem`), dropped the S3-sibling pointer, and de-FMI'd the VRADH Nyquist comment. Added a note on the signed-i16 / 32-bit-superblock provenance.
2. **`size: 1` signedness** — the u8 arm now matches `signed: false, ..`, so a signed `i8` dataset falls through to `UnsupportedPixelType` instead of bit-reinterpreting as `u8`. Matches the `size: 2` arms.

mrauhala:
Round 2 addressed in ac2163a:

1. **Unbounded retention** — added `max_files = 24` (≈2 h at 5-min cadence) to the SMHI `[odim]` table, with a comment pointing operators at `time_window` for a longer window. Keeps the whole static fixture; bounds a production copy.
2. **byte-order wildcard** — added a comment in `read_raw_pixels_2d` explaining that `read_array::<T>()` handles byte-swapping internally, so the `(size, signed)` match is sufficient and the `signed` branch is the load-bearing part.

mrauhala:
Round 3 addressed in 24984e7 (review was "approve with two minor findings"):

1. **Unsupported dtype drops silently** — the wildcard arm now binds the dtype and `tracing::debug!`s it (`ODIM unsupported pixel dtype {dt:?} at {path}`) before returning, so an operator sees *which* type was rejected.
2. **f32 not supported** — non-blocking and not a regression; filed as follow-up **#393** (deferred pending an f32 ODIM fixture to test against, per the engine-test convention) with a code note at the call site.

mrauhala:
Round 4:

**1. `debug!` → `warn!` on unsupported dtype** — applied in cd67a36. An unsupported dtype means the data won't load, so it warrants a production-visible `warn!` (the dtype it names is the actionable detail the caller's generic error lacks), matching the old probe chain.

**2. Committed config → uncommitted fixture (degraded on clean checkout)** — pushing back on this one. It's the **established convention** in this repo: `collections.d/radar-fi-volume-local-h5.toml` is identically committed-active with its `testdata/radar-fmi-pvol` fixture uncommitted, so the FMI local collection is *already* degraded on a clean checkout. A degraded collection is not a failure (`/health` reflects it; only an all-failed registry returns 503). These `*-local-h5.toml` files are dev/reference configs that activate when an operator drops the fixtures in place.

Renaming to `.toml.disabled` would (a) diverge from the FMI precedent and (b) break the active workflow this PR is for — `cargo run -p server -- --collections=…,radar-se-volume-local-h5`, which is exactly how the SMHI volumes are exercised locally.

This is ultimately the same decision as committing the 113 MB fixture (which fully resolves the degradation): that call is with the repo owner. If we decide to commit the fixtures (or just the one 21 MB Kiruna file the test uses), the degradation goes away and the config stays active and consistent with FMI.

mrauhala:
Round 5 (LGTM) notes addressed in 456725f:

1. **Config fixture note** — added a comment to `radar-se-volume-local-h5.toml` making explicit the fixture dir isn't committed and the collection reports `degraded` without it (with the drop-in / `.toml.disabled` options).
2. **Fork tracking issue** — opened #394 (remove the `hdf5-reader` `[patch.crates-io]` pin once a release > 0.6.0 carries the fix) and referenced it in the Cargo.toml patch comment, mirroring icechunk #340.
3. **f32** — already tracked in #393, no action (acknowledged).

PR #401, opened 2026-06-12
perf(engine-geotiff): park STAC metadata-load waiters on a condvar (#213)

### LINKED ISSUE #213: perf(engine-geotiff): wait_for_concurrent_load busy-waits with blocking sleep, holds render permit up to ~9s
## Problem

`wait_for_concurrent_load()` is the loser-side of the STAC metadata-load deduplication in the GeoTIFF engine. When two requests need the same not-yet-loaded STAC entry, one loads it and the other calls `wait_for_concurrent_load()`, which **busy-polls with a blocking sleep**:

```rust
// crates/engine-geotiff/src/lib.rs:611-631
fn wait_for_concurrent_load(&self, timestamp: &DateTime<Utc>, path: &Path) -> bool {
    for attempt in 0u64..20 {
        std::thread::sleep(Duration::from_millis(100 * (attempt + 1).min(5)));
        ...
    }
}
```

That is up to 20 iterations of 100 ms → 500 ms (capped) ≈ **9 seconds worst case** of polling.

## Evidence

- Busy-wait loop: `crates/engine-geotiff/src/lib.rs:611-631`
- Reached from `ensure_metadata()` (`lib.rs:699`), which is called on the render path (`lib.rs:1250`) and the metadata path (`lib.rs:1428`).
- `ensure_metadata()` only does work for STAC stub entries — `lib.rs:684-687` returns early for non-stub (local) entries. So **local-`data_path` collections never reach this code**; only STAC-backed GeoTIFF collections do.
- The render path runs inside `tokio::task::spawn_blocking` (`crates/api-wms/src/handlers.rs:239`), so the blocking sleep does **not** stall an async worker thread — but it does occupy a `spawn_blocking` pool thread **and a render-semaphore permit** for the entire wait.

## Impact

For STAC-backed GeoTIFF collections, a request that loses the dedup race holds a render-semaphore permit (a scarce, ~`2×CPU` resource — see #209) while doing nothing but sleeping, for up to ~9 s. Under concurrent load on a cold catalog this needlessly drains the render pool and inflates tail latency. Does not affect the local radar collection in #201.

## Suggested fix

Replace the busy-wait with proper synchronization so the waiter is parked, not polling, and does not hold a render permit while waiting:
- Use a `tokio::sync::OnceCell` / `Notify` (or a `Shared` future) per in-flight key so concurrent callers `await` the single load instead of sleeping; or
- Acquire the render-semaphore permit *after* metadata load completes, so a waiting request isn't burning a permit.

Tighten the worst-case bound while at it — 9 s is far longer than any reasonable metadata fetch.

## Labels
`priority: medium`, `performance`, `reliability` — milestone v0.3

---
_Surfaced during the #201 GeoTIFF radar WMS performance audit (hunter-concurrency, finding C5). Correctly excluded from that epic because it only affects STAC-backed collections, not the local-`data_path` radar collection — filed separately here._

mrauhala:
## Summary

Fixes #213 — `wait_for_concurrent_load()` busy-polled with a blocking sleep (up to ~9 s in 100–500 ms steps) while occupying a `spawn_blocking` pool thread and a render-semaphore permit. STAC-backed GeoTIFF collections only; local `data_path` collections never reach this path.

## Changes

- **`InFlightLoads` (Mutex + Condvar) replaces the bare in-flight `HashSet`.** Losers of the load race now park on the condvar and wake immediately when the loader finishes — the wait lasts exactly as long as the actual load instead of sleep-poll quantums.
- **`InFlightGuard` notifies on drop**, so success, error, and panic all wake waiters (previously an errored loader left waiters sleeping out their backoff steps).
- **Fixes a TOCTOU race the issue didn't cover:** the "is another thread loading?" check and the "claim the load" insert were in separate lock scopes, so two losers could both fall through and download the same STAC asset twice. `wait_then_claim` does the done-check + claim atomically under one lock.
- **Worst-case bound:** the 10 s cap remains only as a stuck-loader fallback (a STAC asset fetch downloads the whole file, so the bound stays generous); on timeout the waiter takes over the load rather than failing the request, matching the old liveness behavior.

Not done from the issue's alternatives: moving render-permit acquisition after metadata load — that's an API-layer restructure, and with waiters now parked for exactly the load duration the permit hold matches what the loading request itself holds.

## Testing

- 3 new unit tests: waiter wakes on loader success, waiter takes over the claim on loader failure, unclaimed path claims immediately / done work short-circuits.
- `cargo test --workspace` green; `cargo clippy -p engine-geotiff --tests` and `cargo fmt` clean.

### DIFF
--- a/crates/engine-geotiff/src/lib.rs
+++ b/crates/engine-geotiff/src/lib.rs
@@ -56,20 +56,30 @@ impl InFlightLoads {
     }
 
     /// Park until no other thread is loading `path`, then atomically either
-    /// observe the work done or claim the path. Returns `true` if `is_done`
-    /// reports the work complete (nothing left to do); `false` if the caller
-    /// is now the loader (path claimed — pair with `InFlightGuard`). If the
-    /// in-flight loader exceeds `CONCURRENT_LOAD_WAIT` the caller claims the
-    /// path anyway rather than failing the request.
+    /// observe the work done or claim the path. Returns `None` if `is_done`
+    /// reports the work complete (nothing left to do); `Some(guard)` if the
+    /// caller is now the loader — dropping the guard (on success, error, or
+    /// panic) releases the claim and wakes waiters, so the pairing is
+    /// compiler-enforced. If the in-flight loader exceeds
+    /// `CONCURRENT_LOAD_WAIT` the caller claims the path anyway rather than
+    /// failing the request.
     ///
     /// `is_done` is invoked while `self.paths` is held; it must not acquire
     /// that lock (directly or transitively) or the thread deadlocks itself.
-    fn wait_then_claim(&self, path: &Path, is_done: impl Fn() -> bool) -> bool {
+    fn wait_then_claim(
+        &self,
+        path: &Path,
+        is_done: impl Fn() -> bool,
+    ) -> Option<InFlightGuard<'_>> {
         let mut in_flight = self.paths.lock().unwrap_or_else(|e| e.into_inner());
         let deadline = std::time::Instant::now() + CONCURRENT_LOAD_WAIT;
         while in_flight.contains(path) {
             let now = std::time::Instant::now();
             if now >= deadline {
+                // Stuck loader: deliberately claim anyway (a duplicate fetch)
+                // rather than failing the request. The insert below is a
+                // no-op and whichever load finishes first clears the path —
+                // same liveness behavior as the old polling code.
                 break;
             }
             let (guard, _) = self
@@ -79,10 +89,11 @@ impl InFlightLoads {
             in_flight = guard;
         }
         if is_done() {
-            return true;
+            return None;
         }
         in_flight.insert(path.to_path_buf());
-        false
+        drop(in_flight);
+        Some(InFlightGuard::new(self, path.to_path_buf()))
     }
 }
 
@@ -761,11 +772,7 @@ impl GeoTiffEngine {
         timestamp: &DateTime<Utc>,
         asset_url: &str,
         file_size: u64,
-        path: &Path,
     ) -> Result<(), DataServerError> {
-        // RAII guard ensures path is removed from in-flight set even on panic
-        let _guard = InFlightGuard::new(&self.loading_in_flight, path.to_path_buf());
-
         let stub = catalog::StacStub {
             bbox: None, // Not needed for loading
             asset_url: asset_url.to_string(),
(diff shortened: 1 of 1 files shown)


mrauhala:
Review receipts (addressed in 7123624):

1. **`is_done` under the `paths` lock** — added the suggested doc note on `wait_then_claim`: the closure runs while `self.paths` is held and must not acquire that lock.
2. **Tests not guaranteed to exercise the condvar wakeup** — added a `Barrier` sync before the waiter's call in both threaded tests, plus an elapsed bound (`< CONCURRENT_LOAD_WAIT / 2`) after the guard drop: a broken `notify_all` would leave a parked waiter sleeping the full 10 s cap and fail the assertion, instead of the test passing via the not-yet-parked fast path.

mrauhala:
Round-2 receipts (addressed in the latest commit):

1. **Guard pairing not compiler-enforced** — adopted the suggestion: `wait_then_claim` now returns `Option<InFlightGuard>` (`Some` = you're the loader, `None` = already done). `do_load_metadata` no longer constructs its own guard; the claim is released by the returned guard's drop on success, error, or panic. Tests updated to the new API.
2. **Timeout edge case (duplicate fetch past a >10 s stuck loader)** — intentional, same liveness behavior as the old polling code; now documented at the `break` site.

mrauhala:
Round-3 receipts (all three were explicitly flagged by the reviewer as non-regressions):

1. **Stuck-loader fan-out can exceed one duplicate** — accepted; the comment at the `break` site now states dedup is best-effort past the timeout and can fan out to several parallel fetches (02cc3db).
2. **`is_done` lock contract not compile-enforced** — declining for now. Passing the held `MutexGuard` into the closure would make the helper's signature borrow-gymnastics for one private call site; the doc note plus the deadlock being immediate and obvious in any test run (the thread self-deadlocks deterministically, not racily) keeps the risk acceptable for a module-private API.
3. **50 ms sleep is load-bearing for exercising the parked-wakeup path** — acknowledged as informational; the reviewer confirms assertions hold either way, so no false CI failures. Left as is.

mrauhala:
Round-4 receipt: **shared Condvar wakes waiters on all paths** — acknowledged, leaving as is. Waiters re-check `in_flight.contains(path)` in the wait loop, so spurious wakeups are correct-by-construction; at STAC's concurrency (a handful of in-flight metadata loads) per-path condvars would be complexity without measurable benefit. If a large-fan-out use case appears, the suggested `HashMap<PathBuf, Arc<Condvar>>` is the right upgrade path.

Issue #367, opened 2026-06-09
PVOL storm cells: remaining 3D and Features presentation (#407 / #408)


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** Core extraction/tracking and raster overlay shipped. Remaining 3D/Features presentation is tracked by #407/#408.

**Next step / remaining scope:** Use #407 for 3D presentation and #408 for PVOL footprint/track Features; extraction, tracking and raster overlay are implemented.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

Segment connected echo regions above a threshold into discrete **storm cells**; render each as a labelled bounding object (box / ellipsoid) carrying attributes (max dBZ, echo-top, VIL, centroid) via `EXT_structural_metadata`. The basis for **time tracking** — cell-centroid trajectories as 3D polylines across volume scans (pairs with time-dynamic #350).

Analytics tier (3-D connected-components + feature attributes), larger than the rendering issues above.

Part of #346.

</details>

mrauhala:
## #367 storm-cell series — status

Broken into 4 sequential PRs (foundation → render surfaces). Tracking:

| Stage | What | PR / Issue | State |
|---|---|---|---|
| 1 | Core: `ds_core::cells` extraction + tracking, `VolumeEngine::read_cells`, engine-odim cell cache | #404 | ✅ merged |
| 2 | WMS/Maps/Tiles derived `CELLS` overlay layer | #405 | ✅ merged |
| — | Fix: column-max 2-D segmentation + morphological closing (3-D CC produced nested/overlapping footprints) | #406 | open |
| 3 | 3D Tiles `representation=cells` (labelled boxes + track polylines + `EXT_structural_metadata`) | #407 | 📋 issue |
| 4 | OGC API Features (cell polygons + track LineStrings; adds `Geometry::LineString`) | #408 | 📋 issue |

Shipped today across the merged PRs: 3-D→2-D segmentation, attributes (max dBZ, echo top/base, volume, area, column-max VIL), deterministic CCW footprints, gated centroid tracking with per-cell motion vectors (the optical-flow seed), and the CELLS map overlay live in nexus on all PVOL sites.

This umbrella issue (#367) stays open until #407 + #408 land.

PR #404, opened 2026-06-12
feat(ds-core,engine-odim): storm-cell extraction + tracking core (#367, 1/4)

mrauhala:
First of four PRs implementing #367 (storm-cell object extraction + tracking), generalized beyond 3D Tiles: the same cell/track product will be surfaced on WMS/Maps/Tiles (PR 2), 3D Tiles `representation=cells` (PR 3), and OGC API Features (PR 4). This PR is the engine-side foundation only — **no API surface changes**.

### `ds_core::cells` (new module — pure geometry/physics, framework-free)
- **`extract_cells(grid, time, opts)`**: 3-D connected components on the cylindrical `VoxelGrid` (6-connectivity; the azimuth axis wraps on full-circle grids, guarded for future sector scans) at `value >= threshold` (default 35 dBZ, the classic TITAN cell threshold). Per cell: max dBZ + its position, linear-Z-weighted centroid, echo top/base (MSL), **radius-dependent** voxel volume (`Δr·(r_c·Δa)·Δh` — constant cell volume would overweight far range ~4×), footprint area, column-max VIL (`Σ 3.44e-6·Z^(4/7)·Δh`, dBZ capped at 56 for hail), a geographic footprint ring, and a local-ENU AABB (for the future 3D bounding-box product).
- **Footprint ring**: exact boundary trace of the cell's `(radius, azimuth)` column mask along voxel-column edges — seam-spanning storms are rotated seam-free before tracing so they yield **one** ring; largest ring wins (holes dropped, v1); CCW per RFC 7946; Douglas–Peucker simplified at half a radial cell. (Plan note: this replaces the planned marching squares — same goal, but the traced boundary is *exact* for a binary column mask, with no interpolation cases to get wrong.)
- **`track_cells(history, opts)`**: greedy minimum-cost centroid matching between consecutive scans with constant-velocity prediction and a `max_speed·Δt + base_gate` gate; births/deaths on no match; merges/splits resolve as nearest-wins (documented v1). Track ids are deterministic (`{first-scan-stamp}-{label}`); **known limitation** (documented): the id changes when the retention window slides past a track's origin scan. Per-track `motion_ms (u, v)` + `speed_direction()` are first-class — the seed for the future motion/optical-flow field product.

### `VolumeEngine::read_cells` (default-implemented — the *generic* hook)
Any engine whose `volume_info()` advertises voxel-grid capability gets cells + tracks for free: the default impl walks `times` back from the target scan, reads one grid per scan, extracts and tracks. Empty scans are **valid empty results** (Features will need "no cells now" as an empty collection; 3D Tiles maps all-empty to 404) — unlike `read_voxel_grid`, whose empty case is an error; `LocationNotFound` is reserved for capability/no-volumes gaps.

### engine-odim
- `read_voxel_grid`'s cache fill refactored into a **handle-parametric** `voxel_grid_cached(entry, quantity, dims, handle)` so the cells path can run from *either* runtime context: `Some(handle)` on a `spawn_blocking` pool thread (3D Tiles/raster), `None` on a request worker (the Features path, where `block_in_place` is the valid bridge) — per the established `Pixels::handle` rules. Behaviour of the existing path is unchanged.
- New `CELL_SET_CACHE`: process-global, count-bounded memo of per-volume segmentations keyed `(file, quantity, dims, opts-bits)` — volume files are immutable, so entries never go stale (the `VOXEL_GRID_CACHE` argument). In steady state only the newest volume pays resample + extraction. Byte-weighted like `VOXEL_GRID_CACHE` (footprint rings dominate the entry size): `MC_PVOL_CELL_SET_CACHE_MB` (default 64 MB; `0` rejects every entry).
- Cells are gated to **dBZ-unit quantities** (linear-Z/VIL math is reflectivity physics) — anything else is a clean 400.
- `/metrics`: `pvol_cell_set_cache_{hits,misses}_total` + `pvol_cell_set_cache_entries`.

## Plan deviations (called out)
- Footprint contour = exact column-edge boundary trace, not marching squares (see above).
- The `[odim]` config defaults (`cell_threshold_dbz` etc.) are deferred to the Features PR where the engine first *consumes* them — shipping unread config fields in this PR would be dead surface. The 3D Tiles/raster PRs pass options from the API layer (query params + API defaults), matching how `threshold` works today.
- The fivih integration test pins the segmented peak at ≈43 dBZ (40–50 band), not the memory'd "59 dBZ": 59 is the native-resolution point peak; cell-centre resampling onto the test grid smooths it (≈46 at default dims). Verified by probing the actual grid.

## Tests
- `ds-core` unit suite on synthetic grids: blob separation + volume ordering, **azimuth-seam blob = one cell/one ring**, hand-checked single-voxel volume/area/VIL/centroid/echo-top, VIL hail cap, speckle floor + `max_cells` cap, CCW ring; tracking: expected u/v + speed/direction on a moving blob, gate rejection, death on empty scan, merge=nearest-wins; default `read_cells` window/empty-scan/no-caps behaviour via a stub engine.
- `engine-odim` integration (fivih fixture, skips if absent): ≥1 cell with sane physics, closed in-coverage footprint, cache miss-then-hit, 400 on `VRADH`.

Refs #367 (do not close — 3 follow-up PRs).

### DIFF
--- a/crates/core/src/cells.rs
+++ b/crates/core/src/cells.rs
@@ -252,7 +252,11 @@ pub fn extract_cells(
         origin,
         cells: Vec::new(),
     };
-    if total == 0 || grid.values.len() != total || opts.max_cells == 0 {
+    // `u32::MAX` guards the packed BFS frontier indices; any real grid is
+    // orders of magnitude below it (engines cap cell counts in the tens of
+    // millions).
+    if total == 0 || grid.values.len() != total || opts.max_cells == 0 || total > u32::MAX as usize
+    {
         return set;
     }
 
@@ -269,13 +273,24 @@ pub fn extract_cells(
     let a_centre = |i_a: usize| grid.angle_range[0] + (i_a as f64 + 0.5) * da;
     let h_centre = |i_h: usize| grid.height_range[0] + (i_h as f64 + 0.5) * dh;
 
-    let threshold = opts.threshold as f32;
-    let member = |idx: usize| grid.values[idx] >= threshold; // NaN compares false
+    // Compare in f64: an f32-narrowed threshold (e.g. 35.1 → 35.09999847…)
+    // would silently lower the effective cutoff. NaN compares false.
+    let member = |idx: usize| (grid.values[idx] as f64) >= opts.threshold;
 
     // --- Pass 1: BFS connected components, scalar accumulation -------------
     let mut labels = vec![0u32; total];
     let mut comps: Vec<Comp> = Vec::new();
-    let mut queue: VecDeque<(usize, usize, usize)> = VecDeque::new();
+    // The frontier holds packed flat indices (u32, valid up to MAX_VOXELS),
+    // not (usize, usize, usize) triples — 4 B/slot instead of 24 keeps the
+    // worst case (a storm spanning a large fraction of a `high`-tier grid)
+    // at tens of MB transient instead of hundreds.
+    let mut queue: VecDeque<u32> = VecDeque::new();
+    let unpack_idx = |idx: u32| {
+        let idx = idx as usize;
+        let i_h = idx % n_h;
+        let rest = idx / n_h;
+        (rest / n_a, rest % n_a, i_h)
+    };
 
     for seed_r in 0..n_r {
         for seed_a in 0..n_a {
@@ -298,9 +313,10 @@ pub fn extract_cells(
                     max_ih: seed_h,
                 };
                 labels[seed_idx] = raw_label;
-                queue.push_back((seed_r, seed_a, seed_h));
-                while let Some((i_r, i_a, i_h)) = queue.pop_front() {
-                    let idx = VoxelGrid::index_of(grid.dims, i_r, i_a, i_h);
+                queue.push_back(seed_idx as u32);
+                while let Some(packed) = queue.pop_front() {
+                    let (i_r, i_a, i_h) = unpack_idx(packed);
+                    let idx = packed as usize;
                     let v = grid.values[idx] as f64;
                     let rc = r_centre(i_r);
                     comp.count += 1;
@@ -323,7 +339,7 @@ pub fn extract_cells(
                         let nidx = VoxelGrid::index_of(grid.dims, r, a, h);
                         if labels[nidx] == 0 && member(nidx) {
                             labels[nidx] = raw_label;
-                            queue.push_back((r, a, h));
+                            queue.push_back(nidx as u32);
                         }
                     };
                     if i_r > 0 {
--- a/crates/core/src/cells.rs
+++ b/crates/core/src/cells.rs
@@ -1,0 +1,1313 @@
+//! Storm-cell segmentation + tracking on cylindrical voxel grids (#367).
+//!
+//! [`extract_cells`] segments the connected echo regions of a [`VoxelGrid`]
+//! at or above a reflectivity threshold into discrete [`StormCell`]s, each
+//! carrying physical attributes (max dBZ, echo top/base, volume, area, VIL,
+//! centroid) and a geographic footprint ring. [`track_cells`] then matches
+//! cell centroids across consecutive volume scans into [`Track`]s with a
+//! per-step motion vector — the seed for a future motion/optical-flow field
+//! product (the per-cell `(u, v)` is exactly what a flow-field raster would
+//! interpolate).
+//!
+//! Everything here is pure geometry + physics on the domain type — no
+//! framework dependencies, no I/O — so any [`crate::volume::VolumeEngine`]
+//! with voxel-grid capability gets cells for free (see
+//! [`crate::volume::VolumeEngine::read_cells`]), and every API surface
+//! (3D Tiles, Features, WMS/Maps/Tiles) renders one shared product.
+//!
+//! Values are assumed to be radar reflectivity in dBZ: the linear-Z centroid
+//! weighting and the VIL integral are reflectivity physics. Callers gate the
+//! quantity (the ODIM engine only computes cells for dBZ-unit quantities).
+
+use crate::geo::{destination_point, EARTH_RADIUS_M};
+use crate::volume::VoxelGrid;
+use chrono::{DateTime, Utc};
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+/// Liquid-water cap (dBZ) for the VIL integral: reflectivity above this is
+/// treated as hail contamination and clamped, per the standard NSSL practice.
+const VIL_DBZ_CAP: f64 = 56.0;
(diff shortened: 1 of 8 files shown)


mrauhala:
Review round 1 receipts (1f5ed14):
- `volume.rs` `read_cells`: added the explicit **quantity contract** doc — the default impl can't gate units generically, so the dBZ obligation (linear-Z / VIL are reflectivity physics) is now spelled out for future implementors; the ODIM override already rejects non-dBZ quantities with 400.
- `engine-odim/cells.rs`: comment now states the truth — `MC_PVOL_CELL_SET_CACHE=0` builds a 1-entry LRU (quick_cache has no zero capacity), i.e. *effectively* disabled, not a bypass.

mrauhala:
Review round 2 receipts (10805e5):
- **Footprint seam-rotation bug: confirmed real and fixed.** `vertex_polar` now maps a shifted boundary back through the inverse rotation `(y + n_a - shift) % n_a` — the same `orig_a` mapping the occupancy lookups use (the modulo only on full-circle grids; sector grids never rotate). Verified the reviewer's example arithmetic: old mapping put the 36-column seam blob's vertices at 280–320°, fixed mapping at the true 340°–20°. The seam unit test now pins every ring vertex's **bearing and ground range** to the blob's actual span, so it fails against the old mapping.
- **VIL_COEFF unit comment**: corrected to kg m⁻³ (3.44e-6 is the kg form of the textbook 3.44e-3 g m⁻³).
- **Track-id stability across window slides**: acknowledged — will be resolved in the Features PR (persistent ids or explicit documentation of the non-compliance), noted in the follow-up plan.

mrauhala:
Review round 3 receipts (HEAD):
- **Perf (cells.rs area sweep): fixed.** Footprint area now accumulates in the pass-2 voxel loop on a column's first claim (`!d.mask[…]` guard) — the per-rank full-mask sweep and its O(max_cells × n_r × n_a) cost are gone. The hand-checked single-voxel area test pins the result unchanged.
- **CELL_SET_CACHE sizing doc: reworded** per suggestion (pathological ~400 KB/entry → ~1.6 GB ceiling at 4096 entries; count-bounded, unlike the byte-weighted voxel-grid cache; tune via env var).

mrauhala:
Review round 4 receipts (f133425):
- **Cache bound (option a taken): CELL_SET_CACHE is now byte-weighted**, mirroring `VOXEL_GRID_CACHE` — entries weighted by approximate resident bytes (footprint rings dominate), bounded by `MC_PVOL_CELL_SET_CACHE_MB` (default 64 MB). Operators get an exact MB knob regardless of storm activity; `=0` now genuinely rejects every entry (1-byte budget) instead of keeping one ~400 KB object. `/metrics` switches to `pvol_cell_set_cache_{bytes,capacity_bytes}`.
- **Default `read_cells` perf note added**: remote-storage engines are warned off the N-sequential-`read_voxel_grid` pattern and told to override with per-scan caching (as ODIM does).
- **Both `(#404 review)` comment references dropped** per the comment style rule.

mrauhala:
Note on the latest review run: its four findings (count-bounded cache, the two `(#404 review)` comment refs, the missing default-`read_cells` perf note) re-cite the pre-f133425 tree — the run raced my push. All four are already addressed in f133425; see the round-4 receipts above.

mrauhala:
Review round 5 receipts (1b18531):
- **Annulus fallback now simplifies**: the full-circle ring runs through the same ENU Douglas–Peucker as the normal path (361 raw vertices → a few dozen at the chord tolerance).
- **Single shoelace pass**: signed area computed once per ring — magnitude selects the outer ring, sign settles orientation; no repeated traversals or intermediate allocations in `max_by`.
- **PR description corrected**: the cache env var is `MC_PVOL_CELL_SET_CACHE_MB` (byte-bounded, default 64 MB), matching the code.

mrauhala:
Review round 6 receipts (HEAD):
- **Hash-collision risk in the cache key: eliminated** — the key now carries the extraction options verbatim (`(threshold_bits, min_volume_bits, max_cells)` tuple), no FNV fold, so distinct options cannot share an entry even in principle.
- **f64 threshold comparison**: membership now compares `value as f64 >= threshold` — no f32 narrowing of non-representable thresholds like 35.1.
- **BFS frontier**: packed u32 flat indices (4 B/slot vs 24) with a `total > u32::MAX` guard at entry — high-tier worst-case frontier drops from hundreds of MB to tens.
- **Edge-adjacency flat table: considered, declined** — the proposed per-vertex `Vec` is O((n_r+1)×(n_a+1)) allocation *per cell footprint*, while the perimeter being chained is only O(n_r + n_a) edges; the table would cost more than the HashMap it replaces. Happy to revisit if profiling ever shows the chaining on a hot path (it runs once per cached volume).

mrauhala:
Review round 7 receipts (HEAD):
- **Non-deterministic ring output: fixed** — each ring now starts at the smallest remaining vertex id instead of HashMap iteration order; a new test extracts the same grid twice (fresh RandomState each time) and requires byte-identical cell sets.
- **`unwrap_or` window fallback: removed** — a failed target-index lookup is now a surfaced `Engine` error (the target entry comes from the same slice, so a miss is a logic bug, not a condition to paper over).
- **`track_scans` cap: added** — shared `ds_core::cells::MAX_TRACK_SCANS` (24) clamps the window in both the trait default and the ODIM override.

mrauhala:
Review round 8 receipts (bf73a8c):
- **Annulus fallback CW ring: fixed** — boundaries now walk in decreasing azimuth (compass azimuth increases clockwise on the ground, so the CCW exterior traces the circle backwards), and the previously-unexercised annulus path now has a test: echo in every azimuth column ⇒ closed CCW ring with every vertex on the outer radius.

Going forward this PR applies a significance gate per the author: further rounds' findings will be fixed only if they are real correctness/security/perf issues; doc polish and micro-optimizations will be summarized for the maintainer instead of auto-applied.

mrauhala:
Final triage (significance gate — summarized, not auto-fixed):

1. **`footprint_ring` can return `Vec::new()`** — unreachable for any BFS-produced mask (every component claims ≥1 column ⇒ boundary edges exist ⇒ ≥1 closed ring). Will be handled at the natural place: the PR-3/PR-4 encoders that consume footprints will skip/guard empty rings explicitly, rather than adding dead defensive code here.
2. **Default `read_cells` sequential reads** — known and deliberately traded: the default is what makes cells generic across engines; it is doc-flagged and clamped to `MAX_TRACK_SCANS`. Removing the default would defeat the design goal; a runtime WARN can be revisited if a remote-backed `VolumeEngine` ever relies on it.
3. **dBZ contract is doc-only in the default** — restated from round 1; the doc contract plus the ODIM 400-gate is the accepted resolution (ds-core has no unit registry to enforce against).

CI green across the board; all substantive findings from rounds 1–8 are fixed. Settling here.

Issue #397, opened 2026-06-12
feat(ds-render): shared vector polygon → raster fill for alert/hazard map layers (CAP, GeoJSON, IWXXM)


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Completed in current code; close.

**Current assessment and challenge:** Shared fill_polygon primitive and CAP consumer shipped. GeoJSON consumer is separately tracked by #398.

**Closure boundary:** #429 delivered the shared fill/projection primitives, and CAP now calls the shared fill. That satisfies the final consumer criterion identified in the previous comment. GeoJSON and IWXXM consumers remain #398 and #399.

**Source:** [crates/render/src/rasterize.rs:62](https://github.com/mrauhala/meteocore/blob/524a1df934d8da97c34af943f43df892f97fb704/crates/render/src/rasterize.rs#L62) · [crates/engine-cap/src/engine.rs:766](https://github.com/mrauhala/meteocore/blob/524a1df934d8da97c34af943f43df892f97fb704/crates/engine-cap/src/engine.rs#L766)

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Summary

Introduce a **shared vector polygon → raster fill primitive in `ds-render`** so that any vector source (alert/hazard areas) can be rendered as a colored WMS/Maps/Tiles layer — not just exposed as Features. Today **every `MapEngine` impl samples a pre-existing gridded raster**; none rasterize vector geometry. Multiple engines now need polygon fill, so build it **once, reusable**, rather than per-engine.

**Planned consumers:**
- **CAP** alerting engine — #396 (first consumer; severity-shaded alert zones).
- **GeoJSON `MapEngine`** — #398 — let `engine-geojson` (currently Features-only) also render its polygon features as map layers (categorical/attribute-driven fill).
- **IWXXM** aviation weather — #399 — SIGMET/AIRMET hazard areas, volcanic-ash clouds, tropical-cyclone zones are GML polygons; identical Features + Maps surface to CAP.

This issue is a **dependency of #396** and a foundation for the GeoJSON-Maps and IWXXM work.

## Why a shared primitive (not per-engine)

[...]

PR #405, opened 2026-06-12
feat(engine-odim,ds-core): derived CELLS storm-cell overlay for WMS/Maps/Tiles (#367, 2/4)

mrauhala:
Second of four PRs for #367, building on #404 (core extraction + tracking). This one surfaces the storm-cell product on the **raster map APIs**: every reflectivity-capable PVOL site collection now advertises a derived **`CELLS`** parameter layer — footprint outlines, centroid `+` markers, and track polylines, painted at each cell's max dBZ so the ordinary radar colormap styles the overlay for stacking on the DBZH layer.

### `ds_core::raster_paint` (new, framework-free)
Value-space paint primitives on a raster's `Vec<Option<f64>>` *before* colorization: `Canvas::{stroke_segment, stroke_polyline, stroke_ring, paint_marker}` — Bresenham strokes with **Liang–Barsky clipping** (off-tile geometry costs nothing), thickness, non-finite skip, last-writer-wins. This is deliberately the stroke-side seed of the #397 rasterization story (the fill side stays #397's scanline scope).

### `OutputCrs::world_to_fraction` (new)
The inverse of `project_node` for all three variants (linear lon/lat, Mercator-Y, projected-metres via `Crs::forward`). For **per-vertex** use only — overlay geometry is a few hundred points; the per-pixel direction stays `project_node` + `ProjectionGrid` (#203). Roundtrip-tested against `project_node` at multiple fractions per variant.

### engine-odim
- `PolarVolumeSiteView::cells_raster_tile`: computes the cached `CellProduct` (low-dims grid, 6-scan tracking window, defaults from `CellExtractionOptions::default()`), projects ring/track/centroid vertices via `world_to_fraction`, paints. Empty footprints (the unreachable degenerate case flagged in #404's final triage) are explicitly harmless here — `stroke_ring` skips `< 2` points.
- `get_raster_tile` intercepts `parameter == "CELLS"` **before** quantity resolution; `time` selects the scan (WMS TIME animation works), `z` is ignored (cells are 3-D objects, not per-sweep slices).
- `CELLS` is appended to `SiteMeta.parameters` **after** `VolumeInfo` captures its clone — it is a map layer only, never an EDR quantity nor a 3D Tiles quantity (both asserted in the integration test). Only sites with a dBZ-unit moment advertise it (same gate as `cell_product`).
- Styling: inherits the collection fallback colormap; the bundled `radar-fi-volume-local-h5` config gains a documented optional `[[wms.parameters]] name = "CELLS"` entry.

## Render-verified live (fivih fixture)
Server run + GetMap smoke: **WMS in CRS:84, EPSG:4326 (swapped axis), and EPSG:3857**, plus **Maps** and **Tiles** (`6/18/36`) all return painted overlays; visual check confirms the orange/yellow cell outlines + centroid markers sit exactly over the DBZH layer's convective cores (NNE cluster + SW band), and an off-coverage viewport renders transparent. `GetCapabilities` advertises `…fivih/CELLS` titled "Vihti — Storm cells". Track polylines need ≥2 retained volumes, so the single-volume fixture exercises the skip branch; the painting itself is unit-tested at the `Canvas` level and the tracking logic was integration-tested in #404.

## Tests
`raster_paint` unit suite (endpoints, clipping, ring closure, thickness, non-finite, marker); `world_to_fraction` roundtrip for every variant incl. off-tile behaviour; engine-odim integration: CELLS advertised on the layer list but absent from EDR + 3D Tiles lists, painted dBZ values over coverage in WGS84 + WebMercator, empty far away. `cargo test --workspace` green; clippy clean.

Refs #367 (PRs 3/4: 3D Tiles `representation=cells`, Features).


mrauhala:
Triage (significance gate — summarized, not auto-fixed; both are one-liners that can fold into the next PR of this series if wanted):

1. **`world_to_fraction` doc note** — `wgs84_bbox` is intentionally unused in the `Projected` arm (the variant carries its own projected bbox, mirroring `project_node`'s documented behaviour); reviewer asks for an explicit comment in that arm. Doc polish, no behaviour change.
2. **Integration-test upper bound** — `(35.0..70.0)` pins the fixture's accident; the code's real invariant is only `>= threshold`. A ≥70 dBZ hail cell with a swapped fixture would spuriously fail. Test-only robustness.

Everything else reviewed clean (Liang–Barsky/Bresenham correctness, ring-closure float equality, runtime-context handling, the parameters-ordering invariant). CI green; settling here.

PR #406, opened 2026-06-12
fix(ds-core): segment storm cells on the column-max projection (#367)

mrauhala:
Fixes the user-reported rendering mess on the merged CELLS overlay (#405): footprint rings nested inside each other and fragmented into confetti.

## Root cause
Full-**3-D** connected components on raw radar data have two map-hostile behaviours:
1. Echo split vertically by a sub-threshold layer (elevated core above low-level rain) → two components whose footprints **stack** → nested rings.
2. Patchy 35 dBZ fields → one physical rain area fragments into many touching-but-separate components → confetti.

## Fix — TITAN-style column projection
- Mask = **column-maximum projection** (a `(radius, azimuth)` column is in when any voxel reaches the threshold).
- **One-cell morphological closing** (3×3 dilate + erode; azimuth wrap; foreground border padding so closing never shrinks the original mask) bridges single-column speckle gaps.
- 4-neighbour **2-D** connected components.

Footprints can no longer nest or overlap **by construction**, and a vertically split storm is one cell. Attributes are unchanged physics, still computed from the 3-D member voxels (raw ≥ threshold) within the footprint columns — closing widens only the footprint/area. The 2-D labels also simplify per-column VIL (exactly one label per column) and drop the 3-D label buffer + packed BFS frontier.

## Verification
- New tests: stacked-layer echo = **one** cell spanning both layers (volume counts members only); 1-column gap bridges while a 3-column gap stays two cells.
- All previous tests (seam blob, annulus, hand-checked volume/area/VIL/centroid, determinism, tracking) pass **unchanged** — closing is identity on isolated blobs, so the attribute pins still hold.
- **Render-verified** against the fivih fixture: the NNE cluster collapses from dozens of nested/overlapping rectangles into coherent storm outlines; SW fragments merge into one elongated cell.

Rode along (the two review nits from #405): `world_to_fraction` now documents that the `Projected` arm ignores `wgs84_bbox`, and the CELLS integration test asserts only the `>= threshold` invariant (no fixture-specific upper bound).

Refs #367.


mrauhala:
Triage (significance gate — summarized, not auto-fixed):

1. **`bbox_enu_m` excludes closing columns** — this is the *intended* semantics: the box bounds the cell's actual **echo** (3-D member voxels — it's the future 3D Tiles cell box), while the footprint ring deliberately includes the one-cell morphological padding. Growing the 3-D box to cover mask padding would draw boxes around empty air. Agreed the field doc should state "bounds member voxels; may be up to one cell tighter than the closed footprint" and that culling consumers should use the footprint — will ride along with the next push to this area (the 3D Tiles PR touches exactly this field).
2. **`sum_w > 0.0` guard redundant** — true (every component holds ≥1 pre-closing column ⇒ ≥1 member); kept as a cheap belt-and-braces against a future mask-producer change. No action.

CI green; settling here.

PR #409, opened 2026-06-13
fix(engine-odim,ds-render): neutral, connected storm-cell track trails (#367)

mrauhala:
Fixes two user-reported problems with the merged `CELLS` overlay (#405/#406): track trails were **disconnected** (lines for cells no longer present) and **the same colour as the cell outlines**, making the image messy.

## Fixes

**1. Orphan trails → connected only.** A trail is now drawn only when its last observation is the *rendered* (target) scan, so every trail terminates at a visible cell outline. Previously every track in the 6-scan window was drawn, leaving dangling lines for cells that had died or weakened out of the latest scan.

**2. Trail colour → distinct neutral.** Trails are painted at a reserved sentinel value (`CELLS_TRACK_SENTINEL`); the CELLS layer's style wraps whatever colormap it resolves to in a new **`ds_render::OverlayColorMap`** that maps that one sentinel to a fixed neutral colour (`CELLS_TRACK_COLOR`, dark grey) and delegates every other (dBZ) value to the inner radar ramp. So a single colormap-driven raster carries two symbol classes — intensity-coloured outlines + markers, neutral trails — without a second layer. Trails are also 1 px thinner than outlines to read as subordinate.

The overlay wrap sits **outermost** (over any integer-LUT or `[[wms.parameters]]`-override colormap), so the sentinel is matched before any inner clamping. Outlines, centroid markers, projection handling, and the inherited/overridden dBZ styling are unchanged.

## Verification

- **`OverlayColorMap` unit test**: sentinel → overlay colour; every other value → inner colormap; nodata/NaN follow the inner nodata colour (sentinel match is finite-exact, never collides).
- **Render-verified** against the 3-scan `dkste` DMI fixture (15:00/15:05/15:10) — the one local multi-volume PVOL series — in **CRS:84** and **EPSG:3857**: trails now dark grey, 1 px, each connecting to a live cell outline + `+` marker; no orphan lines. (The fivih fixture is single-volume so has no trails; its integration test still passes — outlines only.)

Refs #367. (Doesn't change the algorithm from #406 — purely the WMS/Maps/Tiles trail rendering.)


mrauhala:
Triage (significance gate): the flagged line is a trivially-true self-comparison `color(-9998.0) == color(-9998.0)` — dead weight, the real check is the `assert_ne` beneath it (a near-miss value is NOT the overlay colour). It's harmless but the reviewer is right it catches nothing. Per the standing review gate I'm not pushing a test-only nit on its own (would trigger another review round on an already-"ready to merge" PR). Two options: I can tighten it to assert the near-miss maps to the inner colormap's output (1-line), or it can ride along with the next PR touching ds-render. Say the word and I'll fix; otherwise it's safe to merge as-is.

mrauhala:
Re-run review receipts (f8f3b0d):
- **Orphan-trail fix now has automated tests.** Extracted the paint loop into a fixture-free `paint_cell_product(canvas, product, project)` and added two tests: a connected track (last point == rendered scan) paints a sentinel trail while an orphan track (ended earlier) paints none; a one-point track draws no trail. Locks the user-reported fix against refactors.
- **Trivial assertion tightened**: the near-miss value now asserts delegation to the inner colormap, not self-equality.
- **Generic-helper / ODIM-overlay-constants** architecture point → filed as #410 (tech-debt, `OverlaySpec` on StyleInfo); out of scope here, pure refactor with no behaviour change.

mrauhala:
Re-run round 2 receipts (HEAD):
- **Orphan-trail test hardened**: both tracks now span real multi-point lines (distinct endpoints), not a single coincident coordinate. The connected trail is asserted multi-pixel, and the orphan's absence is checked in a 5×5 neighbourhood around *every* one of its points — so a broken filter draws an unmistakable line the test catches (no longer relies on degenerate-polyline behaviour).
- **NaN sentinel guard**: `OverlayColorMap::new` `debug_assert!`s `sentinel.is_finite()`.

mrauhala:
Re-run round 3 receipts (12bd6b3):
- **Cross-engine sentinel contamination: fixed** (real correctness, not a nit). The wrap was keyed only on `short_name == "CELLS"` but `register_parameter_layer_styles` runs for every engine — a non-ODIM band named `CELLS` would have had `-9999.0` hijacked to grey. Now gated on `collection.engine_type == "odim-volume"`, the only engine that produces a CELLS overlay. #410 will remove the engine-specific branch entirely (OverlaySpec on StyleInfo).

mrauhala:
Triage (significance gate — not fixing): the orphan-absence check covers ±2px around each track vertex, not the ~25px midline. As the review itself notes, the test **does** catch the regression today — the trail filter is all-or-nothing, so a break draws the endpoint pixels that *are* checked. The gap only matters for a hypothetical future 'partial/truncated trail' painter that doesn't exist. Not worth another review round; if partial trails ever land, a bounding-box scan of the orphan's path closes it. Settling the loop here — the substantive findings (orphan test was degenerate, cross-engine sentinel contamination) are fixed.

Issue #411, opened 2026-06-13
feat(server): auto-collection mode (--auto-collections <dir>) — zero-config serve a directory tree


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** Auto-collection phase 1 shipped; GeoTIFF/ODIM inference and reload discovery remain separate unfinished scope.

**Next step / remaining scope:** Finish explicitly unmet GeoTIFF/ODIM inference and reload-discovery phases, using the existing auto-collection mode.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Summary

Make the server launchable **without hand-writing `config.toml`**: point it at a directory tree with `--auto-collections <DIR>` and have it **synthesize `CollectionConfig`s in-memory** by classifying what's on disk — zero TOML. With `--auto-collections` and no `config.toml`, the server boots from built-in defaults.

Target UX (pairs with the CLI startup flags in #271):

```bash
# Zero-config: serve a whole tree of products on port 9000, no config.toml needed
server --auto-collections ./data --port 9000

# Or augment an existing config.toml
server --config /etc/mc/config.toml --auto-collections ./extra-data
```

This is the manual-launch / quick-look story (drop a folder, get OGC endpoints). It does **not** replace `config.toml` or `collections_dir` for production — those stay the source of truth for tuned collections (colormaps, S3 sources, poll intervals, licenses).

> The startup CLI overrides (`--port`/`--host`/`--config`) used in the examples are tracked separately in **#271**. This issue is the auto-collection mechanism only.

---

## Flag

`--auto-collections <DIR>` — repeatable (scan multiple roots). Short alias `--auto` acceptable. Added to the hand-rolled argv parser in `crates/server/src/main.rs` (no `clap` dep).

## Directory → collection mapping (chosen model: per-subdirectory + loose files)

For each root `<DIR>`:

- **Each immediate subdirectory** is classified and (if recognized) becomes one or more collections. Non-recursive beyond one level for *grouping* (the engines themselves still walk their own data dir).
- **Loose data files directly in the root** form collection(s) too (directory-native loose files → one collection named after the root; single-file formats → one collection per file — see below).

```
data/
  radar-fmi/      -> collection "radar-fmi"   (geotiff, phase 2)
[...]

PR #412, opened 2026-06-13
feat(server): CLI startup overrides (--host/--port/--config) + no-config auto-port boot

### LINKED ISSUE #271: server: CLI startup overrides (--port / --host / --config)
## Motivation

The server's listen address and config path are only settable via `config.toml` / the `CONFIG_PATH` env var. Running a throwaway or second instance (smoke-test a fresh build while another holds 8000, CI, a quick local look) currently means copying the whole config to change one line. A small set of CLI startup overrides makes throwaway/parallel runs friction-free — and pairs with the zero-config auto-collection mode in #411 (`server --auto-collections ./data --port 9000`).

## Proposal

Extend the hand-rolled argv parser in `crates/server/src/main.rs` (`parse_cli_args` / `CliArgs`, alongside the existing `--collections`). Keep it **dependency-free** (the parser is intentionally hand-rolled — no `clap`). Support both `--flag value` and `--flag=value` forms, matching the existing `--collections` parsing.

| Flag | Overrides | Precedence |
|---|---|---|
| `--port <u16>` | `[server].port` | CLI > config > default |
| `--host <str>` | `[server].host` | CLI > config > default |
| `--config <path>` | the `CONFIG_PATH` env var | CLI flag > `CONFIG_PATH` env > `./config.toml` |

- `--port`/`--host` are applied in `main()` **after** `ServerConfig::from_file()` returns, **before** the socket bind at `main.rs:205` (`format!("{}:{}", host, port)`).
- `--config` is resolved **first** (it picks the file to load): CLI flag > `CONFIG_PATH` env > `./config.toml`.
- Reject a non-numeric / out-of-range `--port` with a clear error + `exit(2)`, consistent with the existing unknown-arg / unknown-`--collections`-id handling.
- Update the `-h/--help` text and the env-var list.

## Notes

- `BASE_URL` env / `[server].base_url` still win for link generation (`ServerSettings::base_url()`). A CLI `--host`/`--port` change does **not** retro-fit `base_url` — document that (set `BASE_URL` for that). Out of scope: a `--base-url` flag.
- Out of scope: env-var port override; auto-collection (tracked separately in #411).

## Acceptance criteria

- [ ] `server --port 9100` binds 9100 regardless of `[server].port`; `--port=9100` works too.
- [ ] `--host 127.0.0.1` and `--config /path/to/config.toml` behave as specified, with the documented precedence.
- [ ] Non-numeric/out-of-range `--port` exits non-zero with a clear message.
- [ ] `--help` lists all three flags and the precedence vs env vars.

mrauhala:
Closes #271.

## What

Adds the CLI startup overrides from #271 and a **no-config-file boot path** so the binary can be launched without a `config.toml` (the foundation for auto-collections, #411).

### CLI flags (hand-rolled parser, no `clap`)

| Flag | Effect |
|---|---|
| `--port <u16>` | Override `[server].port` (CLI > config). Rejects non-`1..=65535` with exit 2. |
| `--host <str>` | Override `[server].host` (CLI > config). |
| `--config <path>` | Config file path (CLI flag > `CONFIG_PATH` env > `./config.toml`). A missing **explicit** path is a hard error. |

Each also accepts `--flag=value`. `BASE_URL` still wins for link generation.

### No-config boot + auto-port

When the default config path is absent **and** no `--config` is given, the server boots from built-in defaults — host `127.0.0.1`, **auto-scanning for the first free port at/above 8000** (up to 100 ports) — with no collections. It serves `/health` and an empty `/collections`, and can be populated via `POST /admin/collections/reload`.

A port **pinned** by config or `--port` is *not* auto-scanned — a bind conflict stays fatal (unchanged behavior). The "refuse to start empty" guard is relaxed only when nothing is configured; configured-but-all-failed is still fatal. The actually-bound port is written back into config so `base_url()` follows the auto-selected port.

> Per @-request, this is split from the auto-collection feature itself (#411). This PR is the flags + boot/auto-port plumbing; the directory→collections classifier is #411 and builds on `ServerConfig::default_for_auto()` added here.

## Tests

Unit tests: `parse_port`, `bind_auto_port` (picks free start / skips a held port), `parse_collection_list`, and `ServerConfig::default_for_auto`. `cargo test -p server -p ds-core` green; fmt + clippy `--tests` clean.

Manually verified end-to-end:
- `--help` lists the new flags + auto-port note
- `--port abc|0|99999` → error, exit 2
- `--config <missing>` → hard error, exit 1
- no config + no flags → WARN, bound `127.0.0.1:8000`, empty server answers `/health` + `/collections`
- auto-port: 8000 busy → bound `8001`, `base_url` follows
- `--port 8042` with config present → exact bind
- pinned conflict (config or `--port`) → "Failed to bind", **no** scan

## Docs

`README.md` Command-Line Options + `CLAUDE.md` Build & Run updated with the flags and the no-config/auto-port behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,6 +13,24 @@ cargo run -p server          # Start server (reads config.toml)
 cargo check -p <crate>       # Type-check a single crate
 ```
 
+### Server CLI flags
+
+The `server` binary takes a small hand-rolled set of flags (no `clap`), each
+also accepting `--flag=value` (see `parse_cli_args` in `server/src/main.rs`):
+
+- `--collections <id1,id2,…>` — only load these collection IDs.
+- `--host <HOST>` / `--port <PORT>` — override `[server].host`/`port` (CLI wins
+  over config). `BASE_URL` still wins for link generation.
+- `--config <PATH>` — config file path (wins over `CONFIG_PATH` env, then
+  `./config.toml`). A missing `--config` path is a hard error.
+
+**No-config boot:** if the default config path is absent **and** no `--config`
+is given, the server starts from built-in defaults — host `127.0.0.1`, and it
+**auto-scans for the first free port at/above 8000** (up to 100 ports) — with no
+collections. A port pinned by config or `--port` does **not** auto-scan: a bind
+conflict is fatal. This is the base for pointing the server at a directory with
+no `config.toml` (auto-collections, #411).
+
 ### Fuzz testing
 
 Fuzz targets live in `fuzz/` (separate from workspace, uses `cargo-fuzz` + `libfuzzer`). Requires nightly.
--- a/README.md
+++ b/README.md
@@ -73,14 +73,39 @@ Options:
   --collections <id1,id2,...>   Only load collections with these IDs (comma-separated).
                                 All others are silently skipped. Useful for smoke-testing
                                 a single collection without editing config.toml.
+  --host <HOST>                 Bind host. Overrides [server].host (CLI wins over config).
+  --port <PORT>                 Bind port. Overrides [server].port (CLI wins over config).
+                                Must be 1..=65535.
+  --config <PATH>               Config file path. Overrides the CONFIG_PATH env var and the
+                                ./config.toml default. A path given here that does not exist
+                                is a hard error.
   -h, --help                    Print this help and exit.
+```
+
+Each flag also accepts the `--flag=value` spelling (e.g. `--port=8011`). The
+parser is hand-rolled (no `clap`); `BASE_URL` still wins over `--host`/`--port`
+for generated links.
+
+**No-config boot.** If the default config path (`./config.toml`) is absent **and**
+no `--config` is given, the server starts from built-in defaults — host
+`127.0.0.1`, **auto-selecting the first free port at or above 8000** (scanning up
+to 100 ports) — with no collections (it answers `/health` and an empty
+`/collections`, and can be populated via `POST /admin/collections/reload`). A port
+pinned by config or `--port` is **not** auto-scanned: a bind conflict is fatal.
+This is the base for pointing the server at a directory of data with no
+`config.toml` (auto-collections — see issue #411).
+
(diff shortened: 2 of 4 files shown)


mrauhala:
Thanks — all three addressed in 34090cf:

1. **Silent write-back** → now `config.server.port = listener.local_addr().expect(...).port()`, so an (impossible but load-bearing) failure surfaces instead of leaving the wrong port in `base_url()`/logs/reload.
2. **Flaky positive test** → `bind_auto_port_picks_start_when_free` no longer pins an exact port across the `drop`/re-bind gap; it asserts the chosen port is in `[free, free + AUTO_PORT_SCAN)`. The skips-used-port test still covers the advance-on-conflict path.
3. **Security note** → added a startup WARN when the admin endpoint is unauthenticated **and** bound to a non-loopback host (e.g. `--host 0.0.0.0` with no `config.toml`/`ADMIN_TOKEN`), matching the PostGIS pattern.

fmt + clippy `--tests` clean; `cargo test -p server` green.

mrauhala:
Re-review (on 34090cf) raised two follow-up nits — both low-impact; deferring rather than churning a green PR:

- **`Path::exists()` TOCTOU (main.rs:262).** Real but theoretical for a startup config file (it doesn't appear/vanish in the sub-ms window between the check and `from_file()`). The idiomatic fix — drop the existence pre-check and match `io::ErrorKind::NotFound` off `from_file()` — is a rework of this exact no-config boot path, which **#411 (auto-collections) rebuilds** for directory loading. Folding it in there so the try-and-handle shape lands with the code that exercises it.
- **`tried {AUTO_PORT_SCAN}` count (main.rs:357).** Only inaccurate when `start > 65435`; auto-scan always starts at 8000 (`default_for_auto`), so the message is correct in every reachable path. Will tidy the wording ("up to …") alongside #411.

CI is green across the board (Check/Lint/Test, Docker Build, claude-review). Considering this ready to merge.

PR #415, opened 2026-06-13
feat(server,ds-core): reverse-proxy base URL detection (trust_proxy_headers) (#12)

### LINKED ISSUE #12: Proxy header-based URL detection
Auto-detect base URL from X-Forwarded-*/Forwarded headers when behind a reverse proxy.

## Design
- Replace static `base_url` with per-request resolution from proxy headers, gated by `trust_proxy_headers` config flag (default: false)
- **Header priority:** (1) `Forwarded` RFC 7239, (2) `X-Forwarded-Host` + `X-Forwarded-Proto` + `X-Forwarded-Port`, (3) fallback to BASE_URL env > config > `http://{host}:{port}`
- Add `trust_proxy_headers` to `core/src/config.rs`, new `core/src/proxy.rs` with `resolve_base_url()`
- **Security:** Sanitize host values (reject whitespace/slashes/non-ASCII); use first value only
- **Backward compatible:** No behavior change without the flag enabled

mrauhala:
Closes #12.

## What

Absolute self-links across every API (EDR/Features/Maps/Tiles/WMS/3D Tiles landing pages, collection docs, GeoJSON `links`, WMS `GetCapabilities` OnlineResource + legend URLs, the server root page) were built from a base URL resolved **once at startup** (`BASE_URL` env > `[server] base_url` > `http://{host}:{port}`). Behind a reverse proxy that base is wrong (`http://0.0.0.0:8000/edr/...`), which previously required a hardcoded `BASE_URL`.

This adds opt-in **per-request** base-URL resolution from standard proxy forwarding headers.

## How

- New `ds_core::proxy::resolve_base_url(fallback, trust, header_fn)` — the pure resolver. Precedence:
  1. RFC 7239 `Forwarded` (`proto=`/`host=` of the first element)
  2. `X-Forwarded-Proto` + `X-Forwarded-Host` (+ optional `X-Forwarded-Port`)
  3. the static startup fallback
  It is **framework-free**: the axum handlers pass header values in via a closure, so `ds-core` keeps its no-HTTP-deps rule.
- New `[server] trust_proxy_headers` config flag (`#[serde(default)]` → **false**), threaded through `load_collections` into all six API state structs.
- Each API handler that builds self-links resolves the base per request through a tiny per-crate `request_base_url(state, headers)` helper (most handlers already had a `HeaderMap` for `?f=` negotiation; the rest gained one). The server root page resolves inline.

Per-request (not startup) resolution is required: forwarding headers only exist at request time, and one backend behind a proxy can serve multiple external hostnames.

## Security

- **Off by default** — forwarding headers are client-controllable, so resolution only happens when the operator opts in (the proxy is trusted to set/overwrite them).
- When enabled: host values are **sanitised** (whitespace, `/`, `\`, `@`, `%`, control bytes, non-ASCII rejected), scheme restricted to `http`/`https`, and only the **first** value of a comma list is used. A malformed/suspicious header falls through to the next source rather than producing a spoofed URL.
- **No behaviour change without the flag** — `trust=false` returns the static fallback unchanged.

## Tests

- 18 unit tests in `ds-core` for `resolve_base_url` (X-Forwarded, Forwarded incl. quoted host + multi-element, port handling, IPv6 literal, host/proto/port validation + rejection, precedence, trust-off).
- 4 api-edr integration tests: X-Forwarded rewrite, Forwarded rewrite, spoofing-ignored-when-untrusted, fallback-without-headers.
- `cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, and `cargo test --workspace` all green.

## Docs

`trust_proxy_headers` documented in CLAUDE.md (Config Format + a dedicated note) and `config.toml`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -765,7 +765,13 @@ are client-controllable; even when enabled, host values are sanitised
 (whitespace/slashes/`@`/non-ASCII rejected), the scheme is restricted to
 `http`/`https`, and only the first value of a comma list is used — a malformed
 header falls through to the next source rather than producing a spoofed URL.
-Enable it only when a trusted proxy sets/overwrites these headers.
+Enable it only when a trusted proxy sets/overwrites these headers. **Ensure the
+proxy strips or overwrites `Forwarded`/`X-Forwarded-*` on untrusted client
+requests, and that clients cannot reach the backend directly** — otherwise a
+client bypassing the proxy could spoof these headers and control the emitted
+self-links (an open-redirect risk for downstream consumers of those links). The
+`Forwarded` parser uses the **last** element specifically because a client can
+pre-inject the first one (RFC 7239 §4 append semantics).
 
 ### Collection Keywords & License
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -676,7 +676,8 @@ tracked in #125. **The crate is phased; Phases 1-3 ship today.**
 [server]
 host = "0.0.0.0"
 port = 8000
-# base_url = "https://api.example.com"  # optional, for absolute links behind a proxy
+# base_url = "https://api.example.com"  # optional, static fallback for absolute links behind a proxy
+# trust_proxy_headers = true            # optional, derive per-request link base from proxy headers (default false)
 # collections_dir = "collections.d"     # optional, directory of per-collection .toml files
 # watch_collections_dir = true          # optional, auto-reload on collections_dir changes (default false)
 # watch_debounce_ms = 500               # optional, coalesce-window for the watcher (default 500)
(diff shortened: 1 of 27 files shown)


mrauhala:
Thanks — addressed all three review notes:

**Issue 1 (security: `Forwarded` first vs last element) — fixed in c7873bb.**
You're right per RFC 7239 §4 (proxies append, so the first element is the client-injectable oldest hop). `from_forwarded` now uses the **last** element (closest trusted proxy). Added a test asserting a client-injected `host=evil.example.com` first element is ignored in favour of the trusted proxy's last element. `X-Forwarded-*` left as-is (proxies overwrite those; authoritative under the documented trust model).

**Issue 2 (bare IPv6 → broken authority) — fixed in 0d91410** (landed just before this review). `sanitize_host` now validates authority *structure*: a bare unbracketed IPv6 (`2001:db8::1`), unbalanced brackets (`[2001:db8::1`), and stray brackets are rejected and fall through to the static fallback. Tests added for bare/unbalanced IPv6, bracketed-IPv6 + separate port, and leading/trailing-colon hosts.

**Gap (`api_definition` / OpenAPI JSON at `/api`) — verified, no change needed.** The OpenAPI document has no `servers` array and builds no absolute URLs — all paths are relative, so they resolve correctly against the proxied spec URL. Only the Swagger UI page (`/api/docs`), which embeds an absolute `spec_url`, needed per-request resolution, and that was already updated. So `api_definition` correctly takes no `headers` param.

Verify: `cargo test -p ds-core` (22 proxy tests) + api-edr proxy integration tests green; `clippy -D warnings` + `fmt` clean.

mrauhala:
Addressed the two new findings (commit fb3f439):

**`/preview/manifest.json` not proxy-aware — fixed.** Good catch, this endpoint was genuinely missed — it builds its own manifest in `server/src/preview.rs` outside the OGC handler sweep. `manifest_handler` now resolves the base per request from the forwarding headers and passes it into a new `build_manifest_with_base`; the tile-URL templates (`vector_tile_descriptor`/`raster_tile_descriptor`) now follow `Forwarded`/`X-Forwarded-*`. `build_manifest` is now a `#[cfg(test)]` wrapper (existing unit tests unchanged). Added a handler test asserting the manifest's tile URLs use the proxy host.

**Operator-config warning — added** to CLAUDE.md: the proxy must strip/overwrite `Forwarded`/`X-Forwarded-*` on untrusted requests and clients must not reach the backend directly, else the headers can be spoofed (open-redirect risk in downstream link consumers). Also noted the `Forwarded` last-element rationale.

Verify: `cargo test -p server` preview suite (30 tests incl. the new proxy test) green; `clippy -D warnings` + `fmt` clean.

Issue #212, opened 2026-05-19
perf(engine-odim): single-slot composite cache thrashes under concurrent multi-location/timestep queries


mrauhala:
## Problem

`OdimEngine` caches the decoded radar composite in a **single-slot** cache:

```rust
// crates/engine-odim/src/engine.rs:175
cached: Mutex<Option<(String, Arc<OdimComposite>)>>,
```

It holds exactly one `(location_id, composite)` pair. `load_composite()` (`engine.rs:498-558`) checks the slot, and on a key mismatch re-parses the ODIM_H5 file and overwrites the slot. Concurrent requests for **different** locations or timesteps therefore evict each other on every call — each request that doesn't match the currently-cached key pays a full HDF5 parse.

The code already acknowledges this — comment at `engine.rs:519`:
> "...should be a multi-entry cache rather than this single-slot..."

## Evidence

- Cache field: `crates/engine-odim/src/engine.rs:175`
- Lookup + single-slot overwrite: `crates/engine-odim/src/engine.rs:528-558`
- Self-documented limitation: `engine.rs:519-524`
- Seeded with one entry at construction: `engine.rs:317`

## Impact

For any ODIM collection where clients hit more than one location/timestep concurrently (PVOL collections expose per-station locations; COMP collections expose multiple timesteps), the cache hit rate collapses and every miss re-parses the HDF5 file. This is a throughput/latency limiter under concurrent load, independent of the GeoTIFF WMS performance work in #201.

## Suggested fix

Replace the single-slot `Mutex<Option<...>>` with a small bounded multi-entry cache (e.g. an LRU keyed by `location.id()` / timestep, sized to a few entries — composites are large, so a small cap is fine). Recover from mutex poison as the current code already does.

## Labels
`priority: medium`, `performance` — milestone v0.3

---
_Surfaced during the #201 GeoTIFF radar WMS performance audit (hunter-concurrency, finding C4). Correctly excluded from that epic because it affects the ODIM engine, not the local-GeoTIFF WMS path — filed separately here._

mrauhala:
## Update from the local-h5 ODIM WMS full-viewport hot-path perf audit (2026-06-14)

This is the **#1 fix to land before directing animation traffic** at the local-h5 ODIM composite layers. Confirming and sharpening:

[...]

PR #420, opened 2026-06-14
perf(render,api): shared empty-tile cache (#171) + uniform-region assembly fast path (#416)

### LINKED ISSUE #171: Unify empty-tile PNG: WMS/Maps allocate per-request, Tiles uses pre-generated global
## Cleanup

The three raster handlers all serve a transparent PNG on the all-nodata fast path, but the implementation diverges:

- **\`api-tiles\`** uses a \`LazyLock\` global \`EMPTY_TILE_PNG\` ([\`handlers.rs:27-33\`](../blob/main/crates/api-tiles/src/handlers.rs#L27-L33)) and clones the \`Bytes\` handle per request.
- **\`api-wms\`** allocates a zero-filled \`Vec<u8>\` of \`width × height × 4\` and calls \`ds_render::encode_png\` on every empty-tile hit ([\`handlers.rs:245-249\`](../blob/main/crates/api-wms/src/handlers.rs#L245-L249)).
- **\`api-maps\`** same shape as WMS ([\`handlers.rs:833-836\`](../blob/main/crates/api-maps/src/handlers.rs#L833-L836)).

WMS/Maps's per-request allocation is wasteful — empty tiles are uniform, so the encoded PNG is identical across all requests of the same dimensions. The Tiles handler hard-codes \`TILE_SIZE × TILE_SIZE\` (always 256×256), which is why the global works there.

## Constraints

WMS and Maps can't quite reuse the same \`EMPTY_TILE_PNG\` global because the requested \`WIDTH × HEIGHT\` is variable. Options:

1. **Per-dimension cache** — \`DashMap<(u32, u32), bytes::Bytes>\` keyed by output dimensions. First request at a given size pays the encode, every subsequent request clones the \`Bytes\`. Need a bounded LRU to prevent abuse (\`?WIDTH=999&HEIGHT=999\`-style fan-out).
2. **Just encode 256×256 and rely on the client** — Cheapest, but breaks the (rare) case where a caller asks for an unusual aspect ratio.
3. **Encode lazily into a single shared cache shared across all three crates** — Lives in \`ds_render\`. Most invasive.

(1) is the right answer; the rendered-image cache already exists and could potentially serve this role if extended to insert empty tiles under a dimension-only key.

## Priority

Low. Empty-tile PNG encoding for a 256×256 RGBA buffer is fast (~µs scale) and the rendered cache eats the populated-tile cost. This is a "while we're here" cleanup.

## Related

- Pre-existing inconsistency surfaced by review on #169 (the #162 Content-Type fix).

### LINKED ISSUE #416: perf(render): WMS meta-tile assembly runs full premul bilinear + per-pixel divide on every GetMap — add exact 4-equal-texels fast path
## Problem

`render_metatiled` runs its assemble loop (`crates/render/src/metatile.rs:416-428`) + encode on **every** successful full-viewport WMS GetMap — regardless of how many 256² meta-tiles were cache hits — because the arbitrary outer viewport bbox is never itself the cached unit (the exact-bbox `RenderedCache` gets ~3% on pan/zoom; only the aligned 256² `TilePixelCache` absorbs panning). So during animation replay/pan, the assemble pass is the irreducible steady-state per-frame floor: ~2.09M px (1900×1100) × premultiplied bilinear including a **per-pixel unpremultiply divide** ≈ ~10–25 ms/frame.

## Fix

Add an **exact** fast path to the assemble loop: when the 4 source texels for an output pixel are equal (compared as `u32`), return that value directly and skip the premultiply / bilinear / unpremultiply. Bilinear of 4 identical values *is* that value — mathematically exact, so it **cannot** misfire at antialiased edges (edges have differing corners and take the full path). Radar composites are mostly transparent (`[0,0,0,0]` over clear air / ocean), so this fires on the majority of pixels and skips the divide for them.

Do **not** add a "zoom-aligned → crop/nearest" sub-path — it rarely fires for arbitrary full-viewport WMS zoom levels (resolution almost never lands exactly on a TMS ladder level), so it's low value for this client.

## Evidence

- Assemble loop + per-pixel `sample_bilinear` (premul/unpremul, the unpremultiply divide): `crates/render/src/metatile.rs:416-428` (and `sample_bilinear`/`premul`/`unpremul` ~498-545).
- Runs on every GetMap because `TilePixelCache` is keyed on aligned 256² col/row/level/time, but the final viewport image is not cached at the exact bbox.
- PR #377 already trimmed the assembly tax (meta mean ~46 ms); this removes most of the remaining per-frame floor for sparse/transparent radar.

## Scope / severity

MEDIUM. Steady-state per-frame cost on the full-viewport WMS GetMap animation path (EPSG:3857), independent of meta-tile cache hits. Bundles naturally with #171 (uncached empty-tile encode — same `render_metatiled` exit path).

_Surfaced by the local-h5 ODIM WMS full-viewport hot-path performance audit (2026-06-14, 5 auditors + devil's advocate). Related: #241 (parallelize the meta-tile loop), #377 (assembly tax), #171, #212, #268._

mrauhala:
Two render-exit-path efficiency fixes from the local-h5 ODIM WMS hot-path audit, bundled (both touch the `render_metatiled`/empty-tile output path).

## #171 — shared empty-tile PNG cache
WMS and Maps re-allocated `width·height·4` zero bytes and **re-encoded + re-hashed** a transparent PNG on **every** all-nodata response (panning over off-coverage or hitting an empty timestep repeats it per frame — ~8 MB alloc + encode at 1900×1100). 

New `ds_render::empty_tile(w, h)` → a process-global, dimension-keyed, count-bounded cache returning a ready-to-serve `CachedRendered` (bytes + stable ETag), encoded once per distinct size and cloned thereafter. WMS, Maps, **and Tiles** now all source their empty response from it (Tiles keeps its 256×256 `LazyLock`, which delegates to the shared cache) — one mechanism instead of three. Bounded so a `?WIDTH=…&HEIGHT=…` fan-out can't pin unbounded memory.

## #416 — uniform-region assembly fast path
`render_metatiled`'s assemble loop runs a full premultiplied bilinear blend + a per-pixel **alpha-reciprocal unpremultiply divide** on every output pixel of every GetMap — the irreducible per-frame floor, run even when all meta-tiles are cache hits (the arbitrary outer viewport is never itself cached).

Exact fast path in `sample_bilinear`: when the four texels are identical, **bilinear of equal values IS that value** — mathematically exact, so it can't soften antialiased edges (those have differing corners and take the full path). Radar composites are mostly transparent (`[0,0,0,0]`) with large flat regions, so it fires on the majority of pixels and skips the divide for them.

## Testing
- `empty_tile_is_deterministic_and_memoized` — same `(w,h)` → identical bytes + ETag (cache hit); different `(w,h)` differ.
- `sample_bilinear_uniform_region_is_exact` — opaque, translucent (the alpha-reciprocal case), and fully-transparent uniform regions return the exact texel.
- `cargo test -p ds-render` (79 pass) + `cargo clippy -p ds-render -p api-wms -p api-maps -p api-tiles --all-targets -- -D warnings` clean + `cargo fmt` clean.

Closes #171. Closes #416.

### DIFF
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -1184,11 +1184,10 @@ async fn render_map(
             (cached, "MISS", content_type)
         }
         Ok(None) => {
-            // Empty tile: transparent PNG, never cached.
-            let rgba = vec![0u8; (width * height * 4) as usize];
-            let png = ds_render::encode_png(&rgba, width, height)
+            // Empty tile: a transparent PNG, encoded once per (w,h) and shared
+            // across WMS/Maps/Tiles (#171). Not inserted into the rendered cache.
+            let cached = ds_render::empty_tile(width, height)
                 .map_err(|e| MapsError::Internal(format!("Failed to encode empty tile: {e}")))?;
-            let cached = ds_render::CachedRendered::new(bytes::Bytes::from(png));
             (cached, "EMPTY", "image/png")
         }
         Err(e) => {
--- a/crates/api-tiles/src/handlers.rs
+++ b/crates/api-tiles/src/handlers.rs
@@ -22,26 +22,19 @@ use crate::error::TilesError;
 use crate::params::{self, TileQueryParams};
 use crate::tilematrixset::{self, SUPPORTED_TILE_MATRIX_SETS};
 
-/// Pre-generated 256x256 fully transparent PNG for empty (all-nodata) tiles.
-/// Avoids running the colorization + encoding pipeline when a tile has no data.
-static EMPTY_TILE_PNG: LazyLock<bytes::Bytes> = LazyLock::new(|| {
-    let size = params::TILE_SIZE;
-    let rgba = vec![0u8; (size * size * 4) as usize];
-    bytes::Bytes::from(
-        ds_render::encode_png(&rgba, size, size).expect("encoding empty tile PNG must not fail"),
-    )
+/// Pre-generated 256×256 fully transparent `CachedRendered` for empty
+/// (all-nodata) tiles — the bytes and their FNV-1a ETag are computed once per
+/// process instead of on every empty-tile response, and the colorization +
+/// encoding pipeline is skipped when a tile has no data.
+///
+/// Sourced from the shared [`ds_render::empty_tile`] cache, which WMS and Maps
+/// now use too (#171) — so all three raster APIs share one mechanism rather than
+/// each rolling their own. Tiles is always 256×256, so a process-once `LazyLock`
+/// keeps cloning per request essentially free (`Bytes` is `Arc`-backed).
+static EMPTY_TILE_CACHED: LazyLock<ds_render::CachedRendered> = LazyLock::new(|| {
+    ds_render::empty_tile(params::TILE_SIZE, params::TILE_SIZE)
+        .expect("encoding empty tile PNG must not fail")
 });
-
-/// Companion to [`EMPTY_TILE_PNG`]: the same bytes wrapped in
-/// `CachedRendered`, so the FNV-1a hash that produces the empty-tile
-/// ETag is computed **once per process** instead of on every empty-tile
-/// response. Both fields are `Clone`-cheap (`Bytes` is `Arc`-backed,
-/// `String` is a 20-byte allocation done once), so cloning per request
-/// is essentially free. `api-maps` and `api-wms` cannot use this trick
-/// — they allocate fresh empty bytes per request to match the requested
-/// dimensions — but Tiles is always 256×256.
-static EMPTY_TILE_CACHED: LazyLock<ds_render::CachedRendered> =
-    LazyLock::new(|| ds_render::CachedRendered::new(EMPTY_TILE_PNG.clone()));
 
 /// Shared state for the OGC API Tiles service.
 #[derive(Clone)]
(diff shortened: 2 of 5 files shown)


mrauhala:
Re: the `get_or_insert_with` single-flight note on `lib.rs` — I verified against the quick_cache **0.6.23** source (the version we pin), and `get_or_insert_with` **does** deduplicate concurrent first-access, so the "encoded once per distinct size" wording holds:

```rust
// quick_cache-0.6.23/src/sync.rs
pub fn get_or_insert_with<Q, E>(&self, key: &Q, with: impl FnOnce() -> Result<Val, E>) -> Result<Val, E> {
    match self.get_value_or_guard(key, None) {   // None timeout
        GuardResult::Value(v) => Ok(v),
        GuardResult::Guard(g) => { let v = with()?; let _ = g.insert(v.clone()); Ok(v) }
        GuardResult::Timeout => unsafe { unreachable_unchecked() },
    }
}
```

`get_value_or_guard(key, None)` → `PlaceholderGuard::join`, where a second concurrent caller for the same key joins a **waiters list and parks** (`sync_placeholder.rs`: `waiters`, `notify` → `thread.unpark()`); when the guard holder inserts, the waiters wake and return `GuardResult::Value` — they never run the closure. `Timeout` is `unreachable_unchecked()` precisely because `None` means "wait indefinitely". So exactly one caller encodes; the rest coalesce.

The single exception is if the computing closure **errors** (the guard drops without `insert`, so a waiter is promoted to retry) — irrelevant here since encoding a transparent PNG can't fail. Leaving the comment as-is.

(Note: this also confirms the same `get_or_insert_with` single-flight property relied on by the merged #212 composite-cache.)

Issue #46, opened 2026-04-06
Batched byte-range reads for COG


mrauhala:
Batch multiple tile range reads into fewer HTTP requests. Deferred from Phase 3.5.

mrauhala:
Measured the real-world impact against the **live OPERA pan-European COG** (CloudFerro S3) on the nexus production instance (2026-06-14). OPERA is a remote-S3 GeoTIFF in the WMS full-viewport hot path, so this is squarely on the path traffic is being directed at.

## Cold full-viewport GetMap (EPSG:3857, 1900×1100)

| | total | tile_loop | assemble | encode |
|---|---|---|---|---|
| **OPERA cold** (remote S3) | **1.845 s** | **1778 ms** | 56 ms | 6 ms |
| OPERA warm (RenderedCache hit) | 0.0014 s | — | — | — |
| local FMI composite (identical 70-tile geometry) | 0.573 s | 530 ms | — | — |

Both renders fetch **70 meta-tiles, 70 misses** — identical tile geometry. The only variable is where the bytes come from.

## Breakdown

- **96% of the cold render (1778 of 1840 ms) is the tile loop** — remote byte-range fetch + DEFLATE decode of 70 tiles.
- Remote round-trips add **~1.25 s** over the same 70-tile render on local data (530 ms loop).
- At the old fetch-pool cap of 5 (`MAX_TILE_CONCURRENCY`, global shared `TILE_FETCH_POOL`), 70 tiles serialized into **~14 waves ≈ 127 ms/wave** (CloudFerro RTT + per-tile decode).

## Estimated effect of two levers

1. **Raise the concurrency cap (cheap, this is a separate PR).** 5 → 16 turns 14 waves into ~5 → tile_loop roughly **1778 → ~650–750 ms**, i.e. **~1 s off** cold OPERA. One constant; the env override `MC_COG_TILE_CONCURRENCY` lets prod tune without a rebuild.
2. **#46 proper — coalesce adjacent tiles into a few contiguous ranged reads.** Collapses the remaining waves toward 1–2 fat reads + cuts per-request overhead → tile_loop approaches the decode-bound floor (a few hundred ms). Best case **~1–1.3 s off** a cold render.

## Caveat (scope of the win)

[...]

PR #423, opened 2026-06-14
perf(engine-geotiff): raise remote-COG tile fetch concurrency 5→16 (#46)

mrauhala:
Cheap partial mitigation for #46 — the dominant cost of cold full-viewport WMS renders over remote COGs (e.g. the OPERA pan-European composite on CloudFerro S3, which is in the production hot path).

## Why

A cold full-viewport GetMap fetches every covering tile through the shared `TILE_FETCH_POOL`; the pool's thread count is the number of remote byte-range reads in flight at once. The work is **I/O-bound** — each rayon worker blocks on a network range read (driven on the captured runtime handle, #222), then spends a short CPU burst decoding the tile — so the ceiling should track remote RTT, not CPU cores.

## Measured (live OPERA COG, nexus, 2026-06-14)

A 1900×1100 EPSG:3857 cold render fetches **70 tiles, 70 misses**:

| | total | tile_loop | assemble | encode |
|---|---|---|---|---|
| OPERA cold (remote S3) | 1.845 s | **1778 ms** (96%) | 56 ms | 6 ms |
| local FMI (identical 70-tile geometry) | 0.573 s | 530 ms | — | — |

At the old cap of 5, 70 tiles serialized into **~14 waves ≈ 127 ms/wave**. Raising the default to **16** turns that into ~5 waves → tile_loop roughly **1778 → ~650–750 ms**, i.e. **~1 s off** a cold OPERA full-viewport render.

The threads are mostly parked on the network, so a count above the core count is fine.

## Scope of the win

Only **cold / cache-miss** renders — new timesteps, first pan into a new area, fresh data every ~5 min. Warm renders are 1.4 ms (RenderedCache); meta-tile pan reuse is ~85% hit. So this targets **first paint of each animation frame**, which over remote OPERA is the part that feels slow.

## Change

- `MAX_TILE_CONCURRENCY = 5` → `DEFAULT_TILE_CONCURRENCY = 16`.
- Tunable in prod without a rebuild via `MC_COG_TILE_CONCURRENCY` (value ≥ 1 wins, else default), mirroring the `MC_*` env-knob convention used by the ODIM composite / PVOL caches.
- Parse logic factored into a pure `parse_tile_concurrency` + unit test.

#46 (coalescing adjacent tiles into a few contiguous ranged reads) remains the full fix for the remaining few hundred ms; this is the low-risk first lever.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ not re-scan the auto roots.
 | `MC_3DTILES_CONTENT_CACHE_MB` | `512` | 3D Tiles encoded-content cache size in MB. `0` disables. |
 | `MC_PVOL_VOXEL_GRID_CACHE_MB` | `512` | PVOL polar-resampled voxel-grid cache size in MB. `0` disables. |
 | `MC_PVOL_PIXEL_CACHE_MB` | `1024` | PVOL per-moment decoded-pixel cache size in MB. `0` disables. |
+| `MC_ODIM_COMPOSITE_CACHE_MB` | `1024` | ODIM decoded-composite (COMP) cache size in MB. `0` disables. |
+| `MC_COG_TILE_CONCURRENCY` | `16` | Max concurrent remote-COG tile (byte-range) fetches in the shared fetch pool. Raise for high-latency object stores; value must be ≥ 1. |
 | `MC_ALLOW_INLINE_DB_URL` | _(unset)_ | Set to `1` to allow a literal `postgres://` URL in TOML instead of `dsn_env` (development only). |
 
 ### Fuzz Testing
--- a/crates/engine-geotiff/src/reader.rs
+++ b/crates/engine-geotiff/src/reader.rs
@@ -1352,24 +1352,109 @@ pub fn read_pixel(
     Ok(values[local_idx])
 }
 
-/// Maximum number of concurrent tile fetches for remote sources.
-const MAX_TILE_CONCURRENCY: usize = 5;
+/// Default ceiling on concurrent remote-tile fetches when
+/// `MC_COG_TILE_CONCURRENCY` is unset.
+///
+/// A cold full-viewport WMS GetMap over a remote COG fetches all covering tiles
+/// through [`TILE_FETCH_POOL`], so the pool's thread count is the number of
+/// byte-range reads in flight at once. The work is **I/O-bound** — each worker
+/// blocks on a network range read (driven on the captured runtime handle), then
+/// spends a short CPU burst decoding the tile — so the right ceiling is set by
+/// remote round-trip latency, not CPU cores. Measured against the live OPERA
+/// pan-European COG on CloudFerro S3 (2026-06-14): a 1900×1100 EPSG:3857 cold
+/// render fetches 70 tiles; at the old cap of 5 that serialized into ~14 waves
+/// (~127 ms/wave ≈ 1.78 s in the tile loop, 96% of a 1.84 s render). 16 cuts
+/// that to ~5 waves. The threads are mostly parked on the network, so a count
+/// above the core count is fine; raise it further for high-latency stores.
+const DEFAULT_TILE_CONCURRENCY: usize = 16;
+
+/// Hard safety ceiling on the resolved tile-fetch concurrency.
+///
+/// `MC_COG_TILE_CONCURRENCY` feeds `rayon::ThreadPoolBuilder::num_threads`
+/// inside the [`TILE_FETCH_POOL`] `LazyLock`. A fat-fingered value (e.g. a
+/// `100000` typo) would make the OS refuse to spawn that many threads, the
+/// `build().expect(...)` would panic *inside the static initialiser*, poison
+/// the `LazyLock`, and then **every** later request that touches the pool would
+/// panic too — an unrecoverable crash from one bad env var. Clamping to this
+/// ceiling keeps the process alive; 1024 is far above any useful I/O fan-out
+/// yet trivially creatable.
+const MAX_TILE_CONCURRENCY: usize = 1024;
+
+/// Pure parse of `MC_COG_TILE_CONCURRENCY`: `Some(n)` is an applied override
+/// clamped into `[1, MAX_TILE_CONCURRENCY]`; `None` means fall back to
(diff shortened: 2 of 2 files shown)


mrauhala:
Both review findings addressed in e3df595:

1. **Crash-on-misconfig (real reliability bug)** — `parse_tile_concurrency` now clamps an explicit value into `[1, MAX_TILE_CONCURRENCY=1024]` before it reaches `rayon::ThreadPoolBuilder::num_threads`. A `100000` typo no longer panics the `TILE_FETCH_POOL` `LazyLock` initialiser (which would poison the lock and crash every subsequent pool-touching request). Unset / non-numeric / zero still fall back to the default 16. Added clamp assertions to the unit test.

2. **Truncated doc comment** — it was an orphaned doc block (describing a removed `read_bbox` function) wrongly glued onto `read_bbox_overview`, which already has its own doc. Removed the orphan instead of just closing the parenthesis.

fmt + clippy clean, `tile_concurrency_parse` test passes.

mrauhala:
Latest round addressed (force-pushed `181bf61`, history rewritten to a single clean commit):

1. **Unrelated files removed.** You were right — `git add -A` accidentally swept ~40 untracked testdata fixtures + `collections.d/radar-dk-volume-local-h5.toml` (all uncommitted local fixtures, unrelated to this change) into the PR. The branch is now reset to just the two intended files: `crates/engine-geotiff/src/reader.rs` + `README.md`. The fixtures are untouched on disk, back to untracked.

2. **Test clamp boundary.** Correct — `1024 == MAX_TILE_CONCURRENCY` is a passthrough (`1024.min(1024)`), not a clamp. Test now asserts `1024` passes through unchanged *and* `1025`/`100000` clamp down, so the clamp path is actually exercised.

CI green on the prior commit; re-running on the rewritten one.

mrauhala:
Added in e2a6217: the resolved thread count is now logged at pool init (`tracing::info!(threads, "initializing remote-COG tile fetch pool (MC_COG_TILE_CONCURRENCY)")`), so an operator can confirm an override took effect.

One correction for the record: the suggestion cited the `MC_PVOL_*` / `MC_3DTILES_*` caches as logging their resolved capacity at startup — they don't (those are bare `LazyLock` statics with no init log). So this isn't matching an existing pattern, but the underlying point stands: logging the value is genuinely useful for verifying the env override, so I added it. (It fires lazily on the first remote-COG render, not at boot, which is fine — that's exactly when the pool matters.)

mrauhala:
Good escalation — addressed in f1d60e6. You're right that the clamp only handles the absurd-value case; OS thread limits (`ulimit -u`, cgroup `pids.max`, container caps) can fail `build()` for a perfectly reasonable value like 32 in a constrained pod, and the bare `.expect()` would still poison the `LazyLock`.

`build_tile_pool` now degrades gracefully: on a `build()` error it logs a WARN and retries with a **single thread**, so remote-COG renders fall back to serial fetching (the pre-parallel behaviour) instead of crashing every render forever. A 1-thread pool needs exactly one spawnable thread; if even that fails the process can't serve any request at all, so the final `.expect()` is a genuinely unreachable last resort.

One note on the suggested "fall back to `DEFAULT_TILE_CONCURRENCY`": that's the wrong direction — if the requested count already failed to spawn, retrying at a *larger* default (when the operator set something below 16) wouldn't help. The fallback has to go *down*, hence 1.

mrauhala:
Folded both polish items into 37b7573:

1. **Step-down fallback.** `build_tile_pool` now **halves and retries** (requested → /2 → … → 1) instead of dropping straight to 1, so a constrained container recovers whatever concurrency the OS *does* allow rather than going fully serial. It logs a WARN with `requested` vs the actual `threads` when it built below the request. The final `panic!` only fires if even 1 thread can't spawn — at which point the process can't serve any request anyway.

2. **Log source disambiguation.** Refactored `parse_tile_concurrency` to return `Option<usize>` (`Some` = applied override, `None` = fall back to default), so the init log now carries `source = "MC_COG_TILE_CONCURRENCY" | "default"`. An operator grepping the logs can now tell an override of 16 from the compiled default 16 unambiguously. Test updated for the `Option` return.

That fully hardens the path: clamp (absurd value) + halving step-down (OS limits) + unambiguous source logging. I think this is the convergence point — the remaining surface is a rarely-hit init path that now degrades gracefully and logs what it did. Ready to merge from my side.

mrauhala:
Addressed in f558c71:

- **(reader.rs) Error reason on halving** — the halving arm now logs `error = %e` at debug so the cgroup/`ulimit` failure reason is recoverable when diagnosing a constrained-container shrink.
- **(test) Boundary fragility** — the ceiling/over-ceiling assertions now derive their input from `MAX_TILE_CONCURRENCY.to_string()` / `(MAX_TILE_CONCURRENCY + 1).to_string()`, so the test tracks the constant instead of silently desyncing from a hardcoded `"1024"`.
- **(README) `MC_ODIM_COMPOSITE_CACHE_MB`** — confirming: it documents **already-merged** code (the COMP composite LRU from #419, `engine.rs:96` reads the env var). It was just missing from the README table; not a forward declaration. The commit message notes it as a backfill.

That clears this round. The change is fully converged from my side — clamp + halving step-down + source/error logging + self-consistent tests. Ready to squash-merge.

mrauhala:
Considered the latest two and consciously holding both — declining rationale for the record:

1. **README "read once / restart required" note.** Accurate, but the read-once-at-init (`LazyLock`) semantics are identical for all six `MC_*` knobs in that table and none carry the note. Adding it to only this row would be inconsistent, and a deploy here is a container recreate (= restart) anyway, so the value takes effect through the normal flow. Would rather not special-case one row.

2. **INFO logs `requested` before the build.** The message is "**initializing** … pool" (present tense, before the attempt) and the `WARN requested=16 threads=8 …` fires immediately after if the OS forces a reduction, so both lines are visible together. Marginal phrasing preference, not a correctness issue.

Neither is a correctness/security/perf finding. The change is converged: clamp + halving step-down + source/error logging + self-consistent tests, CI green across all rounds. Recommending merge as-is; happy to fold these two into a later doc/log-polish pass if desired.

Issue #421, opened 2026-06-14
perf(server): collections reload rebuilds render caches, dumping the warm meta-tile cache


mrauhala:
## Problem

A collections reload — via `POST /admin/collections/reload` **or** the `collections_dir` filesystem watcher — rebuilds the render caches from scratch in `load_collections` (`crates/server/src/admin.rs`):

```rust
let rendered_cache = Arc::new(ds_render::RenderedCache::new(rendered_cache_mb));
let tile_cache     = Arc::new(ds_render::TilePixelCache::new(metatile_cache_mb));
let vector_tile_cache = Arc::new(ds_mvt::VectorTileCache::new(128));
```

So **every reload dumps the fully-warmed meta-tile cache** (multiple GB in production) and forces a cold re-warm. The meta-tile cache is the primary thing that makes full-viewport WMS `GetMap` fast (the exact-bbox `RenderedCache` only gets ~3% on arbitrary viewports; the meta-tile cache carries ~80%), so a dump spikes render latency until it re-warms — very visible for high-resolution browser clients.

Made worse by the `collections_dir` watcher: a **spurious** inotify event (an `ATTRIB`/access event that doesn't change file content — observed in production from a periodic file reader) triggers a reload with no actual config change, repeatedly dumping the cache on a timer. A live incident showed the meta-tile cache resetting to 0 every ~2.5 min, keeping it permanently cold.

## Fix

Thread the live caches into `load_collections` via a `ReusableCaches` struct and **reuse** each one when its configured byte capacity is unchanged; a size change (or startup) builds fresh. Stale entries survive harmlessly:
- a removed collection's layer 404s before its cached tiles can be served (then they age out via LRU);
- a changed colormap re-colors as new timesteps render (the cache key carries `time`); a hard guarantee for an in-place colormap swap still needs a restart.

## Follow-up (separate)

Have the watcher **skip the reload entirely** when the watched files' content is unchanged (fingerprint compare), so spurious metadata/access events don't churn engines + poll-loops at all — not just the cache.

_Surfaced during a production browser-latency investigation (see the local-h5 ODIM WMS hot-path perf audit)._

PR #425, opened 2026-06-14
fix(server): collections_dir watcher ignores read events — stops self-reload loop (#424)

### LINKED ISSUE #424: collections_dir watcher self-triggers an infinite reload loop (reacts to IN_OPEN read events)
tracked from a prod incident on meteocore.app.meteo.fi (2026-06-14).

## Symptom

With `watch_collections_dir = true`, the server reloads collections continuously — back-to-back, cadence tracking the reload duration (~2.5 min on prod, dominated by a remote-S3 `odim-volume` scan). Each reload rebuilds engines (and, pre-#422, dumps the warm render caches), so high-res browser WMS is permanently cold/slow. A container restart stops it temporarily.

## Root cause (confirmed)

The watcher reacts to **its own reload reading the config files**:

1. `notify` 8.2.0s inotify backend registers `IN_OPEN` in its watch mask **unconditionally** — `notify-8.2.0/src/inotify.rs` `add_single_watch()` uses `ATTRIB | CREATE | OPEN | DELETE | CLOSE_WRITE | MODIFY | MOVED_FROM | MOVED_TO`. `IN_OPEN` surfaces as `EventKind::Access(AccessKind::Open)`.
2. The watcher callback filters on the file **path only** (`is_toml_event`, `crates/server/src/watcher.rs`) and never inspects `event.kind`.
3. `do_reload` → `ServerConfig::from_file` re-opens and reads **every** `collections.d/*.toml` (`fs::read_to_string`).
4. Those opens emit `IN_OPEN` → `Access(Open)` events whose path contains `.toml` → the filter forwards them → another reload → more opens → **infinite self-sustaining loop**.

The watcher + notify 8.2.0 were both introduced in #319 (`fd946aa`, 2026-06-04); the loop went live when that build was deployed. It is invisible to `stat` (reads dont change timestamps) and was only fully provable because the prod `collections_dir` bind mount is **read-only** — so `IN_OPEN` is the *only* inotify event physically possible there, yet the watcher still fires every cycle.

Verified: nothing in the codebase writes to `collections_dir` (all file-mutating calls are test/example-only), so it is not a write side-effect.

## Fix

Filter the notify callback on `event.kind`: ignore read-type accesses (`Access(Open/Read/Close(Read))`) and metadata-only touches (`Modify(Metadata)`); keep every genuine change (create/remove/rename, data writes, `IN_CLOSE_WRITE`). PR incoming.

## Related
- #422 (reload-preserves-cache, merged, not yet deployed) hides the *user-visible* symptom (cache dump) but does not stop the wasteful reload loop — this issue is the actual root cause.

mrauhala:
Fixes #424 — a production incident where the `collections_dir` auto-reload watcher reloaded continuously (every ~2.5 min on prod), keeping the render caches permanently cold and high-res browser WMS slow.

## Root cause (triple-confirmed, with verbatim source)

The watcher reacted to **its own reload reading the config files**:

1. **`notify` 8.2.0 registers `IN_OPEN` in its inotify watch mask unconditionally** — `notify-8.2.0/src/inotify.rs` `add_single_watch()`: `ATTRIB | CREATE | OPEN | DELETE | CLOSE_WRITE | MODIFY | MOVED_FROM | MOVED_TO`. `IN_OPEN` → `EventKind::Access(AccessKind::Open)`.
2. **`is_toml_event` filtered on the file path only**, never `event.kind`.
3. **`do_reload` → `ServerConfig::from_file` re-opens and reads every `collections.d/*.toml`** (`read_to_string`).
4. Each open emits `IN_OPEN` → `Access(Open)` with a `.toml` path → the filter forwards it → another reload → more opens → **infinite self-sustaining loop**.

Invisible to `stat` (reads dont change timestamps); only fully provable on the prod **read-only** `collections_dir` mount, where `IN_OPEN` is the *only* inotify event physically possible — yet the watcher still fired every cycle.

The loop is **latent**: a fresh boot reads the configs *before* the watcher is spawned (`main.rs:425` load vs `main.rs:602` spawn), so it stays quiet until one real reload (a collection edit, a deploy, or an external reader of the dir) arms it — then it self-sustains. That matches the field timeline (watcher enabled for days, loop began when the new radar collections were added).

Confirmed: nothing in the codebase writes to `collections_dir` (all file-mutating calls are test/example-only), so it is not a write side-effect.

## Fix

The notify callback now forwards only **real content/structural changes**:

```rust
Ok(event) if is_toml_event(&event) && !is_read_only_event(&event.kind) => { ... }
```

`is_read_only_event` drops `Access(Open/Read/Close(Read))` and `Modify(Metadata)` (reads + atime/chmod touches) and keeps create/remove/rename, data writes, and `IN_CLOSE_WRITE` (`Access(Close(Write))`) — which `do_reload`s read-only opens never produce, so keeping it is loop-safe. Real edits always emit a kept kind (in-place save → `Modify(Data)`; atomic save/rename → `Create`/`Modify(Name)`/`MovedTo`; rm/add → `Remove`/`Create`), so auto-reload still works. This also immunises the watcher against any external periodic reader of the directory (the originally-suspected #421 kick).

## Tests

`read_only_events_are_dropped` (Open/Read/Close(Read)/Metadata → dropped) and `real_changes_are_kept` (Create/Remove/Modify(Data/Name/Any)/CLOSE_WRITE/Any/Other → kept); existing path-filter tests unchanged. fmt + clippy clean.

## Note on deploy
The deployed `af6bcc1` also lacks #422 (reload-preserves-cache, merged but not deployed), which is why each loop iteration dumped the warm cache. Deploying current main + this fix gets both: #424 stops the loop, #422 keeps any future reload cache-safe.

### DIFF
--- a/crates/server/src/watcher.rs
+++ b/crates/server/src/watcher.rs
@@ -15,6 +15,7 @@
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
+use notify::event::{AccessKind, AccessMode, EventKind, ModifyKind};
 use notify::{RecursiveMode, Watcher};
 use tracing::{info, warn};
 
@@ -36,7 +37,13 @@ pub fn spawn_collections_watcher(
 
     let mut watcher =
         notify::recommended_watcher(move |res: notify::Result<notify::Event>| match res {
-            Ok(event) if is_toml_event(&event) => {
+            // A reload must fire only on a real content/structural change — NOT
+            // on a read. `do_reload` re-opens every `.toml` to parse it, and
+            // notify's inotify mask always includes `IN_OPEN`, so reacting to
+            // read-access would let a reload's own file reads re-trigger the
+            // next reload — an infinite self-sustaining loop (see
+            // [`is_read_only_event`]).
+            Ok(event) if is_toml_event(&event) && !is_read_only_event(&event.kind) => {
                 // Ignore send errors — a closed receiver just means the task ended.
                 let _ = tx.send(());
             }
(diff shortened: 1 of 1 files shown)


mrauhala:
Good catch — added in d175642. The `Open(_)` wildcard dropping `Open(Write)` is deliberate (a write-mode open isn't a commit; the real signals are `Close(Write)`/`Modify(Data)`), and the test now asserts it so a future narrowing to `Open(Any | Read)` can't silently reintroduce premature reloads on write-opens.

PR #426, opened 2026-06-14
feat(wms,render): labelled GetLegendGraphic — tick values + title/unit (#371)

### LINKED ISSUE #371: WMS GetLegendGraphic: draw value tick labels + unit on the legend (match the 3D Tiles point-cloud legend)
## Problem

The WMS `GetLegendGraphic` legend is just a bare colour gradient with **no value scale**. `render_legend` (`crates/render/src/lib.rs:385`) fills only the left ~40% with the colormap gradient and leaves the right ~60% blank white — there are **no tick labels, no min/max values, no unit, no title**. A client gets a coloured strip with no way to read which colour means which value.

```rust
// crates/render/src/lib.rs:385
let gradient_width = (width * 2 / 5).max(10); // left 40% is gradient
// … fills gradient only; the reserved right 60% is never drawn into
```

## What we want

Match the quality of the new **3D Tiles point-cloud legend** added in #370 (commit 2ee8585): the collection JSON carries a `legend` (the colormap sampled into `#rrggbb` stops over a value range via `legend_stops`), and the viewer renders a labelled gradient bar — "Reflectivity (dBZ)" title with ticks at 0/10/…/70. The WMS `GetLegendGraphic` PNG should be comparably informative:

- **Tick labels** at round values alongside the gradient (min, max, and a few intermediate), drawn in the currently-blank right portion.
- A **title / unit** (e.g. `Reflectivity (dBZ)`) — the colormap min/max come from the WMS layer config; the unit from the parameter metadata.
- Works for **every** colormap and the layer's configured value range.

## Design constraint (the crux)

`ds-render` must stay **framework-free — only `ds-core` + `png`** (see CLAUDE.md). So text/labels can't pull in a heavy font crate. The label drawing needs a **minimal embedded bitmap font** (a small 5×7 digit/glyph table is enough for numbers, `-`, `.`, and a short unit string) or simple vector glyph drawing. This font/glyph helper is the bulk of the effort.

## Where

- `crates/render/src/lib.rs` `render_legend` — add tick-label + title drawing (and a small glyph helper module).
- The plumbing already exists: `crates/api-wms/src/handlers.rs:543` (`GetLegendGraphic` → `render_legend`) and `crates/api-wms/src/capabilities.rs` (`LegendURL`). The handler will need to pass the unit/title through.

## Acceptance

- `GetLegendGraphic` returns a PNG with the gradient **and** labelled ticks + unit; legible at the default size.
- Unit tests in `ds-render` cover the labelled output (dimensions, that labels are drawn, deterministic bytes).

## Reference

3D Tiles legend (the model to match): #370 / `crates/api-3dtiles/src/handlers.rs` `legend_stops` + `crates/api-3dtiles/viewer/index.html` `buildLegend`.

mrauhala:
Closes #371.

## Problem

The WMS `GetLegendGraphic` legend was a bare colour gradient: `render_legend`
filled only the left ~40% with the colormap and left the rest blank white — no
tick labels, no min/max, no unit, no title. A client got a coloured strip with
no way to read which colour means which value.

## What changed

Matches the quality of the 3D Tiles point-cloud legend (#370): a labelled
gradient bar with a value scale and a "Reflectivity (dBZ)"-style title.

- **`ds-render::font`** — a new framework-free 5×7 bitmap font module
  (`ds-render` must stay `ds-core` + `png` only, so no font crate). A
  hand-authored glyph table (digits, punctuation, A–Z, a–z, degree sign) written
  as `0bXXXXX` row literals so each glyph is a reviewable little picture, plus
  `draw_text` / `text_width` blitting into an RGBA buffer.
- **`render_legend`** — now draws a bordered gradient swatch (colours composited
  over white so a colormap with transparent low values reads as a solid faded
  swatch, not a see-through strip), a nice-number tick scale (`nice_ticks`,
  classic 1-2-5 algorithm) with stripped numeric labels (`format_tick`), and an
  optional title line. The layout adapts to the requested size and degrades to a
  bare swatch for very small thumbnails. New `title: Option<&str>` arg.
- **WMS handler** — resolves the title (`<parameter> (<unit>)`) from the
  engine's `raster_info()` (parameter short-name + collection-level unit), and
  bumps the default legend size to 180×300 so the labels fit. Clients can still
  request any size (capped 512×1024); the renderer degrades gracefully.

### Known limitation

The unit comes from the collection-level `RasterInfo.unit`. Multi-unit sources
(radar polar volumes, where quantities span dBZ / m·s⁻¹ / dB / …) deliberately
report no single collection unit, so their legends show the quantity name +
tick scale but no unit suffix. Threading per-parameter units through to the
legend is a possible follow-up.

## Verification

Render-verified the output via the `gen_legend` example (radar, temperature,
viridis, fraction, and a 40px thumbnail) — title, ticks, labels, and the
composite-over-white swatch all legible:

- radar `DBZH (dBZ)`, −32…95, ticks −20/0/20/…/80
- temperature `Temperature (degC)`, −30…40
- 40px thumbnail still fits a slim labelled scale (graceful degrade)

## Tests

- **ds-render**: font glyphs / `text_width` / clipping; `nice_ticks` (round
  values, degenerate ranges, reversed bounds); `format_tick`; deterministic
  labelled-legend bytes + title-affects-output.
- **api-wms**: `GetLegendGraphic` end-to-end (200, `image/png`, immutable
  cache-control) and the default-size bump (PNG IHDR parsed to assert 180×300),
  plus explicit-size round-trip.

### DIFF
--- a/crates/api-wms/src/handlers.rs
+++ b/crates/api-wms/src/handlers.rs
@@ -636,8 +636,20 @@ pub async fn wms_handler(
                 .as_ref()
                 .map(|i| i.unit.clone())
                 .filter(|u| !u.is_empty());
-            let title = match (param, unit) {
+            let param_unit = match (param, unit) {
                 (Some(p), Some(u)) => Some(format!("{p} ({u})")),
+                (Some(s), None) | (None, Some(s)) => Some(s),
+                (None, None) => None,
+            };
+            // For a non-default style, show its name on a second line so the
+            // legend identifies which of a layer's styles it depicts (the colours
+            // already come from this style's colormap). "default"/"Default" carry
+            // no information, so they're omitted.
+            let style_line = (style_name != "default")
+                .then(|| style_info.title.trim().to_string())
+                .filter(|t| !t.is_empty() && !t.eq_ignore_ascii_case("default"));
+            let title = match (param_unit, style_line) {
+                (Some(pu), Some(sl)) => Some(format!("{pu}\n{sl}")),
                 (Some(s), None) | (None, Some(s)) => Some(s),
                 (None, None) => None,
             };
--- a/crates/api-wms/src/handlers.rs
+++ b/crates/api-wms/src/handlers.rs
@@ -591,18 +591,21 @@ pub async fn wms_handler(
                 ))
             })?;
 
+            // Default to a size that fits the value-tick labels + title (#371);
+            // a client can still request a smaller thumbnail, where the renderer
+            // degrades to a bare swatch.
             let width: u32 = query
                 .width
                 .as_deref()
                 .and_then(|s| s.parse().ok())
-                .unwrap_or(40);
+                .unwrap_or(180);
             let height: u32 = query
                 .height
                 .as_deref()
                 .and_then(|s| s.parse().ok())
-                .unwrap_or(200);
-            let width = width.min(256);
-            let height = height.min(1024);
+                .unwrap_or(300);
+            let width = width.clamp(1, 512);
+            let height = height.clamp(1, 1024);
 
             let format = crate::params::parse_image_format(query.format.as_deref())?;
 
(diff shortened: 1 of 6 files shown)


mrauhala:
Addressed the `render_legend` swatch finding in 763f317:

- **Bare-swatch bounds bug**: capped `swatch_w` at `width - grad_x0` (not `width`) so the gradient no longer spills `PAD` columns past the row end, and routed every swatch/tick pixel write through a bounds-checked `set_px` so tiny heights (swatch row off-canvas) draw nothing instead of panicking. Added `render_legend_tiny_sizes_do_not_panic` sweeping w×h down to 1×1 — it caught a second OOB in the vertical geometry, also fixed.
- While here, dropped the legend's duplicate 1-2-5 nice-number tick code and reused `plot::nice_ticks` (now `pub(crate)`), clipped to the gradient range. Radar/temp/fraction legends render byte-identically.

mrauhala:
Addressed the second review round in 07592e5, plus your question about styles:

- **`format_tick` precision loss** (lib.rs:442): now picks the fewest decimals (0–6) that reproduce the value, so a `[0, 0.001]`-style range keeps its tick digits instead of collapsing to `0`; a round-trip epsilon snaps float noise (`0.1+0.2` → `0.3`). Test: `format_tick_keeps_small_magnitude_digits`.
- **`pub mod font`** (lib.rs:3): narrowed to `pub(crate) mod font` — it's a legend implementation detail, nothing outside ds-render uses it.
- **Selected style on the legend**: the gradient already came from the requested STYLE; now the style's name is drawn on a second title line (non-default styles only), so a layer's styles are distinguishable. `render_legend` takes a newline-separated title; single-line output stays byte-identical. Tests: `render_legend_draws_multi_line_title` (ds-render) + `legend_graphic_reflects_selected_style` (api-wms).

mrauhala:
Addressed the third review round in 00910a6:

- **Label-width fit** (lib.rs:534): the "room for labels?" check now sizes from the *actual widest* label (e.g. "-32" = 17 px) instead of a fixed "00" (11 px), so a narrow legend no longer clips a label's right edge — it falls back to a bare swatch below the fit threshold. Ticks are computed once and reused.
- **Sub-µ tick precision** (lib.rs:462): `format_tick` falls back to scientific notation ("2e-8") when a value is too small to show at 6 dp, so a `[0, 1e-7]` range's ticks stay distinct instead of all collapsing to "0". Test: `format_tick_uses_scientific_for_sub_micro`.
- **Brittle assertion** (lib.rs:772): relaxed the 0..70 exact-equality to intent-level checks (≥4 uniformly-spaced ascending ticks incl. both extremes), so a future tick-density tweak won't break it — while still guarding the extreme-inclusion property this PR fixed.

Radar legend stays byte-identical; ds-render 91 + api-wms 29 tests pass, fmt + clippy clean.

mrauhala:
Per-parameter legend units (the radar PVOL no-unit limitation noted above) is now tracked as #428.

PR #427, opened 2026-06-14
fix(wms): GetLegendGraphic accepts singular STYLE alias (#165)

### LINKED ISSUE #165: WMS GetLegendGraphic only accepts STYLES, not STYLE — asymmetric with GetMap
## Bug

The two WMS sub-handlers treat the \`STYLE\` / \`STYLES\` aliases inconsistently:

- **GetMap** ([\`params.rs:185-195\`](../blob/main/crates/api-wms/src/params.rs#L185-L195)) accepts either:
  \`\`\`rust
  let style = self.styles.as_deref().or(self.style.as_deref()).unwrap_or(\"\");
  \`\`\`
- **GetLegendGraphic** ([\`handlers.rs:281\`](../blob/main/crates/api-wms/src/handlers.rs#L281)) only reads the plural:
  \`\`\`rust
  let style_name = query.styles.as_deref().unwrap_or(\"default\");
  \`\`\`

A client sending \`STYLE=my-style\` to \`GetLegendGraphic\` (without also sending \`STYLES=my-style\`) silently receives the \`default\` legend instead of the one they asked for. No error, just wrong output.

Note that \`LAYER\` / \`LAYERS\` IS handled symmetrically in GetLegendGraphic ([\`handlers.rs:275-279\`](../blob/main/crates/api-wms/src/handlers.rs#L275-L279)) — only the STYLE alias is the odd one out.

## Fix

Mirror the LAYER/LAYERS pattern:

\`\`\`rust
let style_name = query
    .styles
    .as_deref()
    .or(query.style.as_deref())
    .unwrap_or(\"default\");
\`\`\`

One-line change.

## Related

Spotted by the Claude review on #159 (README PR documenting WMS behaviour).

mrauhala:
Closes #165.

## Bug

`GetMap` accepts both `STYLE` and `STYLES`, but `GetLegendGraphic` only read the
plural `STYLES`. A client sending `STYLE=my-style` (the SLD-standard singular
parameter for a legend) silently received the `default` legend instead of the
requested one — no error, just wrong output. `LAYER`/`LAYERS` was already
handled symmetrically; the `STYLE` alias was the odd one out.

## Fix

Mirror the existing LAYER/LAYERS pattern — fall back to `query.style` when
`query.styles` is absent:

```rust
let style_name = query
    .styles
    .as_deref()
    .or(query.style.as_deref())
    .unwrap_or("default");
```

`WmsQuery` already parses both (`#[serde(alias = "STYLE")]`), so this is purely
the handler honouring the field GetMap already uses.

## Test

`legend_graphic_accepts_singular_style_alias` (api-wms integration): `STYLE=radar_fmi`
returns bytes identical to `STYLES=radar_fmi` and distinct from the default legend.

### DIFF
--- a/crates/api-wms/src/handlers.rs
+++ b/crates/api-wms/src/handlers.rs
@@ -571,10 +571,7 @@ pub async fn wms_handler(
                 .or(query.layer.as_deref())
                 .ok_or(WmsError::missing_parameter("LAYER"))?;
 
-            // Accept singular STYLE as well as plural STYLES — GetLegendGraphic's
-            // SLD parameter is `STYLE`, and GetMap already honours both. Without
-            // this a client sending only STYLE=… silently gets the default
-            // legend instead of the one it asked for (#165).
+            // Also accept singular STYLE= — SLD standard for GetLegendGraphic (#165).
             let style_name = query
                 .styles
                 .as_deref()
--- a/crates/api-wms/src/handlers.rs
+++ b/crates/api-wms/src/handlers.rs
@@ -571,7 +571,15 @@ pub async fn wms_handler(
                 .or(query.layer.as_deref())
                 .ok_or(WmsError::missing_parameter("LAYER"))?;
 
-            let style_name = query.styles.as_deref().unwrap_or("default");
+            // Accept singular STYLE as well as plural STYLES — GetLegendGraphic's
+            // SLD parameter is `STYLE`, and GetMap already honours both. Without
+            // this a client sending only STYLE=… silently gets the default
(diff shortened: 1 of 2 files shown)


mrauhala:
Re the comment-length nit: the cited rule ("CLAUDE.md: Never write multi-line comment blocks — one short line max") isn't actually in this repo's CLAUDE.md (`grep` finds no such rule), and multi-line explanatory comments are used throughout the codebase. The comment documents the #165 "why" (silent default-fallback), so I'm leaving it as-is rather than pushing a nit-only commit. CI is green.

Issue #399, opened 2026-06-12
feat: IWXXM aviation-weather engine — hazard areas (SIGMET/AIRMET/VA/TC) as Features + WMS/Maps layers


mrauhala:
## Summary

New **`engine-iwxxm`** crate ingesting **ICAO/WMO IWXXM** (ICAO Meteorological Information Exchange Model — GML-based aviation MET XML) and exposing the **hazard-area products** as:

- **OGC API – Features** — one GeoJSON Feature per hazard area (SIGMET / AIRMET / Volcanic Ash / Tropical Cyclone), geometry from the GML polygon, aviation metadata as properties.
- **WMS / OGC API – Maps / Tiles** — hazard zones rendered as colored map layers (by phenomenon / intensity), with TIME selecting active hazards.

**Depends on #397** (shared `ds_render::rasterize::fill_polygon` + `ds_core::geo::geometry_to_pixels`). Third consumer of that primitive, after CAP #396 and GeoJSON-Maps. Architecturally this is **CAP's twin**: XML hazard polygons valid over a time window → Features + Maps. The differences are the *parser* (GML/OM, not CAP) and the *domain vocabulary*.

IWXXM spec/schemas: WMO No. 49 / ICAO Annex 3; schemas at https://schemas.wmo.int/iwxxm/ (current operational release 3.0 / 2021-2).

## Why

IWXXM is the mandated international format for aviation hazard exchange (replacing legacy TAC SIGMET/AIRMET text over WMO WIS 2.0 / GTS). Rendering SIGMET/AIRMET/volcanic-ash/tropical-cyclone areas on the same map as radar and NWP gives a complete aviation-weather picture through standard OGC APIs.

## Scope

**v1 = hazard polygons (Features + Maps):**
- **SIGMET** (incl. Volcanic Ash SIGMET, Tropical Cyclone SIGMET), **AIRMET**, **Volcanic Ash Advisory**, **Tropical Cyclone Advisory** — each carries a GML geometry (`gml:Polygon`/`gml:Surface` with `gml:posList`).

**Deferred (separate surface / follow-ups):**
- **METAR / SPECI / TAF** — station *point* observations/forecasts. These are EDR/Features **point** data, not polygons; a different mapping (likely `EdrEngine`) — file separately.
- **Space Weather Advisory** (global/band regions, not local polygons).

## IWXXM parsing (quick-xml; GML geometry)

[...]

Issue #398, opened 2026-06-12
feat(engine-geojson): add MapEngine — render polygon features as WMS/Maps/Tiles layers


mrauhala:
## Summary

Add a `MapEngine` impl to **`engine-geojson`** (today Features-only — `crates/engine-geojson/src/loader.rs`, R-tree in `spatial.rs`) so a GeoJSON collection's **polygon features** can also be served as WMS / OGC API Maps / Tiles **map layers**, not just as Features. The fill is driven by a feature **property** (categorical or numeric) and colorized with a `ds-render` colormap.

**Depends on #397** (shared `ds_render::rasterize::fill_polygon` + `ds_core::geo::geometry_to_pixels`). This is the second consumer of that primitive (after CAP #396).

## Why

GeoJSON is the natural source for static/reference vector layers — administrative boundaries, flood/hazard zones, catchments, categorical masks, choropleths. MeteoCore already serves them as Features; rendering the same polygons as a colored basemap-style layer lets them sit on the WMS/Maps stack alongside radar/NWP without a second tool or a pre-rasterized GeoTIFF.

## Scope (v1 = polygons)

- **Polygon / MultiPolygon** features fill into the pixel grid. (`Point`/line geometries are skipped in v1 — marker/stroke styling is a separate follow-up; the shared primitive is a polygon fill.)
- One layer per collection. Optional: one sub-layer per category value as a follow-up.
- **No time dimension** — GeoJSON is treated as static reference data (no `TIME`); `data_version()` (file mtime, already present) drives ETags. A datetime-property time axis is a possible follow-up.

## Design

`MapEngine::get_raster_tile(bbox, w, h, time, output_crs, parameter, z, reference_time)`:

1. Query the existing R-tree for features intersecting `bbox` (reuse `spatial.rs`).
2. For each polygon feature, derive a numeric **value** from the configured property (see config), project its rings via `ds_core::geo::geometry_to_pixels` (#397), and `fill_polygon` it into the `w×h` grid (`Combine::Replace` or a configured precedence).
3. Return `RasterTile { values }`; the WMS/Maps layer colorizes via the configured `ds-render` colormap.

`raster_info()` is **O(1) from a cached `ArcSwap<RasterInfo>` snapshot** rebuilt on catalog swap (per the hot-path rule — don't recompute per request).

## Attribute-driven styling

Two modes, picked in config:

[...]

Issue #396, opened 2026-06-11
feat: CAP (Common Alerting Protocol) engine — alerts as Features + WMS/Maps layers


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Completed in current code; close.

**Current assessment and challenge:** CAP Features and raster engine, source ingestion and server wiring exist. Close completed engine scope or split explicit unimplemented extras.

**Closure boundary:** The CAP engine, local/feed acquisition, Features/raster exposure and server wiring are implemented. This closes the engine-delivery request; the original explicitly out-of-scope product extensions are not being claimed as delivered.

**Source:** [crates/engine-cap/src/engine.rs:55](https://github.com/mrauhala/meteocore/blob/524a1df934d8da97c34af943f43df892f97fb704/crates/engine-cap/src/engine.rs#L55) · [crates/engine-cap/src/source.rs:1](https://github.com/mrauhala/meteocore/blob/524a1df934d8da97c34af943f43df892f97fb704/crates/engine-cap/src/source.rs#L1) · [crates/engine-cap/src/engine.rs:694](https://github.com/mrauhala/meteocore/blob/524a1df934d8da97c34af943f43df892f97fb704/crates/engine-cap/src/engine.rs#L694)

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Summary

Add a new **`engine-cap`** crate that ingests **OASIS Common Alerting Protocol (CAP) v1.2** emergency/weather alerts and exposes each alert area as:

- **OGC API – Features** — one GeoJSON Feature per alert area (polygon/circle), with the full alert/info metadata as properties; filterable by bbox, datetime (active-at-time), and limit/offset.
- **WMS 1.3.0 / OGC API – Maps / Tiles** — alert zones rendered as a colored map layer (severity-shaded polygon fill), with the standard CRS/TIME machinery.

Spec: https://docs.oasis-open.org/emergency/cap/v1.2/CAP-v1.2-os.html

Two mutually-exclusive data sources (both required):
1. **Local directory** of CAP `.xml` files — polled/scanned like the other file-backed engines.
2. **CAP feed from a web address** — an Atom/RSS index whose entries link to individual CAP documents (the MeteoAlarm / US-NWS pattern), fetched over HTTP via `ds-storage`.

[...]

PR #429, opened 2026-06-14
feat(ds-render): shared vector polygon → raster fill primitive (#397)

mrauhala:
Implements #397 — a **shared vector polygon → raster fill primitive in `ds-render`** so any vector source (alert/hazard areas) can be rendered as a colored WMS/Maps/Tiles layer, not just exposed as Features. Built once, reusable, rather than re-deriving scan-conversion in each engine. Foundation/dependency for CAP (#396), GeoJSON Maps (#398), and IWXXM (#399).

## What's in this PR

Clean split that keeps the framework-free invariants (`ds-render` = `ds-core` + `png`; CRS math stays in `ds-core::geo`):

**1. Projection — `ds_core::geo::geometry_to_pixels` (vertices only)**
- Projects a `Geometry`'s polygon rings into output **pixel** space via `OutputCrs::world_to_fraction` — projecting *vertices*, never per output pixel (per #203).
- Returns `PixelGeometry { polygons: Vec<PixelPolygon { exterior, holes }> }` — `Polygon` → one, `MultiPolygon` → one per member, `Point`/`Null` → none.

**2. Fill — `ds_render::rasterize::fill_polygon` (pure 2-D scanline)**
- Even-odd scanline conversion over the combined exterior + hole edge set, so holes punch out **regardless of winding** (robust to inconsistently-wound source data).
- Spans clip to `[0,width) × [0,height)` — polygons partly off-tile fill only their visible span (load-bearing for tiled WMS/Maps).
- `Combine::{Replace, Max, First}` overlap policy. `Combine::Max` resolves overlapping alerts to the highest severity **deterministically, regardless of draw order** (CAP's policy).
- No `ds-render` colorization changes: it fills a `Vec<Option<f64>>` *before* colorize, so the existing `LutColorMap`/`IntegerLutColorMap` pipeline styles the result.

## Acceptance criteria

- [x] `fill_polygon` unit tests: convex, concave, holes (+ winding-independence), per-tile-edge clipping, degenerate/collinear rings, `Combine::Max` overlap (order-independent), buffer-mismatch no-op, non-finite vertices.
- [x] `geometry_to_pixels` tests for `Wgs84`, `WebMercator`, and a `Projected` (EPSG:3067) output, pinned against expected pixel coords (central-meridian `fx=0.5` is an independent pin).
- [x] Antimeridian-crossing polygon behavior documented as a v1 limitation.
- [ ] **#396 (CAP) adoption** — intentionally **out of scope** here; CAP is not yet implemented. This PR is the dependency it will adopt.

## Out of scope / follow-ups (per the issue)

Edge densification for large projected polygons, antimeridian splitting, line/point styling, anti-aliased edges.

## Testing

- `cargo test -p ds-core` (5 new `geometry_to_pixels` tests) — green
- `cargo test -p ds-render` (10 new `rasterize` tests) — green
- `cargo fmt` + `cargo clippy -p ds-core -p ds-render --all-targets` — clean
- `cargo check --workspace` — green

### DIFF
--- a/crates/core/src/geo.rs
+++ b/crates/core/src/geo.rs
@@ -1,5 +1,8 @@
 use std::f64::consts::PI;
 
+use crate::feature::Geometry;
+use crate::map_engine::OutputCrs;
+
 /// WGS84 ellipsoid parameters
 const WGS84_A: f64 = 6_378_137.0; // semi-major axis (meters)
 const WGS84_F: f64 = 1.0 / 298.257223563; // flattening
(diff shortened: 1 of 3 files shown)


mrauhala:
Addressed the `Combine::Max` NaN asymmetry note (1530547): added `debug_assert!(value.is_finite())` at the top of `fill_polygon`, documenting the finite-value precondition and catching misuse early without supporting NaN in the hot path. No caller passes non-finite values (severity/category codes are finite). Tests + clippy green.

mrauhala:
Addressed all four round-2 nits in 5c22608 (batched, none blocking):

1. **Combine::Max NaN comment** — reworded: the risk is the first-fill case (`None => value` writes `Some(NaN)`; `f64::max` maxNum keeps `existing` after), not a populated slot.
2. **x_start `.min(w)`** — added for explicit symmetry with `x_end`.
3. **non_finite_vertices test comment** — corrected: this ring leaves one valid edge → odd crossings → zero fill; the contract is no-panic robustness.
4. **`geometry_to_pixels` param** — renamed `bbox` → `wgs84_bbox` to match `world_to_fraction` and flag WGS84-only semantics at the call site.

Tests (15) + clippy green.

PR #430, opened 2026-06-14
feat(engine-cap): CAP v1.2 alert engine — Features + vector→raster WMS/Maps/Tiles (#396)

mrauhala:
Implements **#396** — a new `engine-cap` crate ingesting OASIS **Common Alerting Protocol (CAP) v1.2** alerts and serving each alert area as **OGC API – Features** *and* a severity-shaded **WMS / Maps / Tiles** layer. Builds on the shared polygon→raster fill primitive **#397** (already merged).

## What's in this PR

**First vector→raster `MapEngine`.** Every existing map engine samples a grid; `engine-cap` fills alert *polygons* into the output pixel grid with `ds_render::rasterize::fill_polygon`, fed by `ds_core::geo::geometry_to_pixels` — projecting polygon **vertices** (never per pixel, #203). One engine implements both `FeatureEngine` and `MapEngine` over a single poll-and-swap `Catalog`.

**Sources** (config-validated, mutually exclusive, both via `ds-storage` on the background poll runtime):
- local directory of CAP `.xml` files;
- Atom/RSS `feed_url` whose entries link to individual CAP docs — index fetched, links extracted, docs fetched with bounded **`get_many`** concurrency (origin-grouped), not a sequential blocking loop.

**Parsing (quick-xml):**
- **`lat,lon` → `[lon, lat]` swap on ingest** (CAP §3.3.4) — pinned by an absolute-position test (a Helsinki alert must land at lon≈25/lat≈60, not the wrong hemisphere).
- rings closed defensively; `<circle>` → geodesic N-gon (`circle_segments`, default 64) carrying `radius_km`;
- **one Feature per `(alert, info, area)`**, id `{identifier}.{infoIdx}.{areaIdx}`; multiple `<info>`/`<area>` fan out;
- geocode-only areas → **`Geometry::Null` Features** (RFC 7946, listed but off-map) + counted;
- `language` / `status_filter` (default `["Actual"]`) / `default_ttl` honored.

**Time:** active window `[onset∨effective∨sent, expires∨+ttl∨open]`. Features `datetime=` selects overlapping areas; Map/WMS `TIME` selects active-at-instant; **no time ⇒ active *now*** (snapshot `as_of`, advanced each poll). WMS advertises a `TIME` dimension whose default is "now".

**Rendering:** value = CAP severity code (Unknown=0 … Extreme=4); overlaps resolve via **`Combine::Max`** (highest severity wins, order-independent). New **`cap_severity` builtin colormap** (grey→green→yellow→orange→red, semi-transparent). Web Mercator + projected output CRS supported via the shared `OutputCrs` machinery.

## Acceptance criteria (all met)

- [x] Local-dir **and** feed-url collections load and appear in `/collections` (config + wiring + validation).
- [x] `GET /features/.../items` returns valid GeoJSON, correctly positioned (lat/lon swap verified), with bbox/datetime/limit filtering; `/items/{id}` returns a single area.
- [x] WMS `GetCapabilities` advertises the layer + `TIME`; `GetMap` renders severity zones in **CRS:84 and EPSG:4326** (byte-identical → axis order correct) **and EPSG:3857**; Maps/Tiles render the same.
- [x] `datetime`/`TIME` select active alerts; default = now.
- [x] Poll picks up added/removed/changed files without restart.
- [x] `cargo test -p engine-cap` (18 unit + 15 integration) and `cargo test --workspace` pass; `cargo fmt` + clippy clean.

## Verification

End-to-end smoke test against the bundled demo collection (`collections.d/cap-alerts.toml` over `testdata/cap/`): Features list/single/bbox/datetime, WMS GetCapabilities/GetMap (CRS:84 ≡ EPSG:4326 ≡ EPSG:3857), Maps, Tiles, and GetLegendGraphic all return correctly. The rendered map shows the flood polygon (orange/Severe) under the heat circle (red/Extreme) with the overlap resolving to red — confirming `Combine::Max`.

## Out of scope for v1 (follow-ups, noted in CLAUDE.md)

geocode→geometry lookup, reference-chain supersedes/cancel beyond latest-wins, XML-DSig verification, per-`event` sub-layers, conditional-GET feed caching, antimeridian splitting.


mrauhala:
Thanks for the review — addressed in `03d281d`:

**✅ Bug (admin.rs health) — fixed.** A reachable CAP source with zero active alerts was permanently `Degraded`. Added `CapEngine::is_loaded()` (set `true` on the first successful `refresh()`, including a zero-record load) and gated `Ready` on that instead of `feature_count > 0`. New integration test `empty_source_is_loaded_but_has_no_records` covers it.

**↩️ Bug (catalog.rs:148 feature-ID) — respectfully, this is not a collision; keeping `.` with a proof comment + test.** The format `{identifier}.{info}.{area}` is injective because `info`/`area` render as **digits only** — so peeling `.{digits}` twice from the right always recovers exactly `(identifier, info, area)`, regardless of dots in the identifier. The example in the comment doesn't actually collide: `"X.0.0"` + `.0.0` → `"X.0.0.0.0"`, but `"X.0.0.0"` + `.0.0` → `"X.0.0.0.0.0"` (different lengths). Added `dotted_identifiers_produce_distinct_retrievable_ids` to lock this in.

Also, the suggested `{}:{}/{}` would introduce a real regression: the feature id is a **single URL path segment** in `/items/{featureId}`, so an embedded `/` would split the route. Left a code comment noting both points.

**✅ Minor (catalog.rs DefaultHasher) — fixed.** Reworded the `data_version` doc to not overclaim cross-Rust-version stability: identical content hashes equal within a build/toolchain (keeping ETags valid across polls), and a hasher change across Rust versions is a one-time, consequence-free ETag invalidation — never staleness.

mrauhala:
Round-2 review addressed in `cc5d5d7`:

**✅ SSRF (Medium) — fixed.** Feed entry links are now fetched only if they share the feed's **exact** origin (scheme+host+port — compared as origins, never a prefix, so `https://feed` rejects `https://feed.evil.com`) or match a new optional `feed_allowlist` of URL prefixes; everything else is dropped with a WARN. So a compromised feed can't pivot to `http://169.254.169.254/…` or internal hosts. Config-validated (entries must be http(s); require a feed). Unit test `ssrf_filter_allows_feed_origin_and_allowlist_only` + config tests cover it. Mirrors `stac_asset_allowlist`.

**✅ ETag misses text fields (Low) — fixed.** `compute_version` now also hashes `event`/`headline`/`description`/`instruction`/`areaDesc`, so an in-place text correction invalidates the Feature ETag (no stale 304).

**📝 build_store per poll (Perf Low) — annotated, not warned.** After the SSRF filter the common case is a single origin (the feed's own) → one HTTP client per poll cycle (every `poll_interval_secs`, default 300s), negligible. I added a code comment rather than a per-poll `WARN`, which would be log spam for any legitimate multi-origin feed. `DataStore` has no cross-call client reuse; pooling is a noted future optimisation.

**↩️ Duplicate fixtures (Minor) — intentional, kept separate.** They aren't identical dirs: `crates/engine-cap/tests/fixtures/` has **5** files including `test-status.xml` (to exercise status filtering) and is referenced cwd-independently via `CARGO_MANIFEST_DIR`; `testdata/cap/` is a **4-file demo subset** for the runnable example config + smoke test, referenced from the workspace root. A symlink would couple the demo to test internals and git symlinks are platform-fragile.

**✅ `local_name` misnamed (Minor) — renamed** to `decode_name` with an accurate doc comment (quick-xml's `.local_name()` already strips the prefix; the fn only decodes bytes → String).

mrauhala:
Round-3 review addressed in `0980391` — all three real:

**✅ Duplicate `<identifier>` collision — fixed.** `Catalog::build` now dedups records by id (records are sorted by id, so duplicates are adjacent → keep the first) and logs a `WARN`. `records` and `id_index` are consistent again — no listed-but-unreachable features. Test `duplicate_identifiers_are_deduped_and_reachable`.

**✅ Identifier with `/` → unreachable — fixed (lossless).** Confirmed api-features builds the self-link href **verbatim** (`format!(… /items/{}, feature.id)`, no encoding) and routes via `Path<(String,String)>` (which percent-decodes). So the fix is: emit a **percent-encoded `Feature.id`** (URL-safe single segment — `%`/`/`/`?`/`#`/space) while keeping the **decoded** id as the catalog/lookup key. On `GET`, axum decodes the href segment back to the lookup key → match. The reviewer's `replace('/', "%2F")` on the *stored key* would actually 404 (stored encoded, but `Path` yields the decoded form), so I split display-vs-key instead. Test `identifier_with_slash_is_url_safe_and_reachable` pins the round-trip.

**✅ `feed_allowlist` prefix widening — fixed.** Config validation now **rejects** any allowlist entry that doesn't end with `/`, so `https://cdn/cap` (which would also admit `https://cdn/cap-staging/`) is a load error. Doc comment + config test updated. (The exact-origin default needs no slash — it's an origin equality check, not a prefix.)

Thanks — the "not flagged" notes (Relaxed ordering, bounded degenerate-bbox scan, DefaultHasher) match my reasoning.

mrauhala:
Round-4 review addressed in `6983290`:

**✅ Degenerate bbox over-render — fixed.** `get_raster_tile` now returns an empty tile when `Bbox::new` fails, instead of falling back to rendering every record. Test `degenerate_request_bbox_renders_empty_tile`.

**✅ Silent no-status drop — fixed.** A missing `<status>` under a non-empty filter is now counted and surfaced with **one WARN per build** (`dropped N alert(s) with no <status>…`). I used a per-build count rather than the suggested per-alert WARN so a systematically-malformed feed doesn't spam a warning every poll for every alert. Test `missing_status_dropped_under_filter_but_kept_when_unfiltered`.

**📝 Index-fetch redirect SSRF — verified + documented (honest, no false guarantee).** I checked object_store 0.11's source: its reqwest client is built with the **default** redirect policy (follows up to 10 redirects) and `ClientOptions` exposes **no** redirect knob. So redirect-following is a property of *every* `ds-storage` HTTP fetch (zarr/grib HTTP too), not CAP-specific — and it affects the entry fetches as much as the index fetch (a redirect response isn't re-checked by `is_allowed_entry`). The correct fix is to disable redirects at the `ds-storage` layer (a custom reqwest client, since object_store can't), which hardens all HTTP engines and is a cross-engine change out of scope for this PR. I documented it as a known limitation in the index-fetch comment and the CAP CLAUDE.md notes (feed host is operator-trusted) rather than adding a comment claiming a guarantee that doesn't exist. Happy to file the ds-storage follow-up.

**✅ Missing Features `info!` log — fixed** (parity with the other arms).

The "not flagged" notes (feature-id encoding, allowlist trailing slash, `MAX_RENDER_RECORDS`) match.

mrauhala:
Round-5 review addressed in `eae2416`:

**✅ Post-hoc index size check — fixed.** The feed index now goes through `get_many(&[index_path], 1, Some(MAX_DOC_BYTES))`, which HEAD-checks the body size before pulling it into memory — same download-layer cap as the entry fetches. No more loading an oversized body before a length check.

**↩️ `origin_of` default-port — non-issue (verified + pinned).** The `url` crate **normalises default ports away**: `Url::parse("https://host:443/x").port()` returns `None`, identical to `https://host/x` (and `:80` for http). So `origin_of` already treats `https://host:443` and `https://host` as the same origin — an entry link that spells out the default port is allowed. Added an assertion in `ssrf_filter_allows_feed_origin_and_allowlist_only` to lock it in. (A non-default port like `:8443` is correctly still a distinct origin.)

**✅ Dead `collection_id` param — removed** from `base_raster_info` and `Catalog::empty` (and the call sites). `Catalog::build` still takes it — it's used by the dedup / missing-status WARNs.

**✅ Redirect SSRF tracking — opened #431.** Filed the cross-engine ds-storage redirect-disable work as a security/enhancement issue, referenced from the index-fetch comment and CLAUDE.md.

The "verified correct" list (lat/lon swap, exact-origin allowlist, concurrency, ds-render dependency scope, id uniqueness, DefaultHasher) all match. I believe the substantive findings are now resolved.

mrauhala:
Round-6 addressed in `41a4470`. Fixed the real items; the rest I've reviewed and am rebutting with evidence (several contradict round-5's own "verified correct" list).

### Fixed
- **[Must] `[`/`]` (and any path-unsafe byte) encoding.** Replaced the hand-listed `replace` chain with a proper **RFC 3986 pchar** encoder: every byte outside unreserved/sub-delims/`:`/`@` is percent-encoded — so `[`, `]`, space, `%`, `?`, `#`, `/`, and non-ASCII all round-trip. Unit test `encode_feature_id_handles_path_unsafe_chars` + integration `identifier_with_brackets_is_url_safe_and_reachable`.
- **[HIGH] query-URL post-hoc size check.** The query-bearing entry fetch now uses the size-guarded `get_many(.., Some(MAX_DOC_BYTES))` (HEAD-checks before buffering), like the index + main paths.
- **[MEDIUM] silent `MAX_ALERTS_PER_DOC` truncation** → now WARNs.
- **[MEDIUM] vertex truncation → wrong polygon.** `parse_polygon` now **rejects** (drops + WARN) an over-`MAX_POLYGON_VERTICES` ring instead of truncating it.
- **[MEDIUM] `circle_ring` close** → now via `first()` (defensive; `n>=3` already guarantees non-empty, and inputs are range/finite-validated so `destination_point` can't produce NaN).

### Reviewed — intentionally unchanged (with evidence)
- **[CRITICAL] redirect SSRF** — already documented + tracked in **#431**. object_store 0.11 follows redirects with no disable knob *and* doesn't expose the final URL, so a CAP-local block is infeasible; the correct fix is at the `ds-storage` layer (#431). It's the existing behaviour of every HTTP-backed engine, not a regression this PR introduces; feed mode trusts the operator-configured host (matches the issue's scope).
- **[MEDIUM] `ds-render` dependency "violation"** — this is exactly #397's design: the shared `fill_polygon` lives in `ds-render` for engines to use. CLAUDE.md documents engine-cap as the first engine to depend on it, and **round-5's own review listed this under "verified correct"** ("Architecture compliance … explicitly documented in CLAUDE.md"). The rule restricts *API* crates and *ds-render's own* deps, not engines.
- **[HIGH] sync HTTP fetch in `load_collections`** — `do_reload` is offloaded to `poll_runtime` (admin.rs:2577), not a request worker, and the live engines keep serving until the atomic state swap, so request serving is never blocked. It's identical to the initial scans geotiff/querydata/odim already do in `new()`.
- **[MEDIUM] `DefaultHasher` across restarts** — `DefaultHasher::new()` is fixed-key SipHash (deterministic), *not* `RandomState`, so a 304 **does** fire across restarts of the same binary; only a Rust-version recompile changes it (one-time, consequence-free). Round-5 verified this too.
- **[LOW] `origin_of` default ports** — the `url` crate normalises `:443`/`:80` away (verified + test added in `eae2416`), so `https://host:443` ≡ `https://host`.
- **[LOW] CDATA not unescaped** — correct by spec: CDATA is *literal* character data, never entity-encoded. Unescaping it would corrupt a literal `&`/`<`.
- **[LOW] language fallback** — intentional (don't silently drop an alert that lacks the requested language); the feature's `language` property reports the language actually served, so it isn't hidden.
- **[LOW] `build_times` ttl boundary** — the `default_ttl`-derived end **is** the effective expiry, a real active-set-change instant, so advertising it as a TIME value is correct.
- **[LOW] `list` depth guard** — the local-path trust model shared by every local-dir engine (csv/geojson/geotiff `data_path`, `collections_dir`): whoever can set the path controls what's served; `MAX_LOCAL_FILES` bounds the retained set. `data_path=/` is operator misconfiguration. Recursive listing is intentional (date-organised subdirs).

mrauhala:
Addressed in `87d4c36`:

**✅ [Medium] one HTTP client per query-bearing entry URL.** Capped `with_query` at **64** with a WARN. Origin-grouping (as the query-less path does) isn't possible here because an object-store `ObjectPath` can't carry a query string — that's exactly why these are split out — so a tight cap is the right bound. (MeteoAlarm's `application/cap+xml` links are query-less, so this path is empty for it; the `?index_*` query strings are only on the `<id>` elements, which aren't followed.)

**✅ [Low] RFC 7946 exterior ring winding.** Now normalized in `build_geometry` — exterior → CCW, holes → CW — covering CAP polygons, the CW-generated circle N-gons, and geocode-resolved geometry. Heads-up: the suggested snippet was **sign-flipped** (`if signed_area > 0 { reverse }` makes a CCW ring CW — the opposite of §3.1.6). Positive shoelace area = CCW in lon/lat, so the fix reverses when the area is **negative**. Pinned with `exterior_ring_is_normalized_ccw` (feeds a CW polygon, asserts positive output area).

**✅ [Low] zero-width temporal extent.** `temporal_extent()` now returns `None` (unknown ⇒ unbounded per OGC) when every window is fully open, instead of `(as_of, as_of)`. Test `temporal_extent_none_for_fully_open_windows`.

**✅ [Low] dedup readability** → `records.dedup_by_key(|r| r.id.clone())` with `before - len()` count.

**✅ [Trivial] double scan** → `parse_polygon` is single-pass; the cap now counts **valid** vertices (post-parse), matching its description.

**📝 Acknowledged:** redirect-SSRF tracked in **#431** (filed); CLAUDE.md now records the `ds-render` dep as an approved exception for vector→raster `MapEngine` impls; `DefaultHasher` cross-instance ETag note stands (no change for v1).

mrauhala:
Addressed in `87816cc`:

**✅ Bug — `parse_polygon` silently dropping invalid vertices.** Now an invalid vertex (out-of-range lat/lon, NaN, malformed token) **rejects the whole ring** with a WARN, instead of skipping it and serving a different-shaped polygon. Same stance as the existing vertex-count guard (reject, don't reshape). Test feeds `91,24 60,25 …` (one bad vertex + 4 valid) and a NaN token → both `None`; a fully-valid ring still parses.

**✅ Note — `raster_info()` O(1) comment.** Corrected to "cheap clone, O(times)" (bounded by the 256-entry TIME cap).

**✅ Note — `Feature.id` percent-encoding.** Added to CLAUDE.md: `Feature.id` is the percent-encoded single-segment id (so the verbatim self-link routes); clients building a URL from it must use it **as-is** (not re-encode), and the raw CAP `<identifier>` is in `properties.identifier`.

**📝 SSRF redirect (confirmed known)** — no change; tracked in **#431** for the cross-engine `ds-storage` fix.

27 unit + 22 integration tests pass, clippy clean.

mrauhala:
Addressed in `be40ce9`:

**✅ [Must] shutdown race.** Real bug. Retained the initial `watch::Receiver` and `poll_loop` now **clones** it (pinned at version 0) instead of `subscribe()`-ing fresh — so a `shutdown()` that fires before the loop runs is still observed. Heads-up: the suggested "retain the receiver as a field" alone **doesn't fix it** — a fresh `subscribe()` after the `send()` starts at the post-send version and still misses the signal; the receiver must be *cloned* from the unpolled retained one. New `#[tokio::test] shutdown_before_poll_loop_is_not_lost` (would hang → time out under the old code). Same `channel(_, _drop)` + subscribe-in-loop pattern is in querydata/geotiff/odim — same latent race, worth a cross-engine follow-up.

**✅ [Must] `temporal_extent()` O(N).** Precomputed in `Catalog::build` alongside `spatial_extent`; the accessor is now an O(1) field read (§211).

**✅ [Should] `DefaultHasher` ETag stability.** Replaced with **FNV-1a** — a fixed, documented algorithm, so `data_version` is stable across toolchain upgrades *and* across instances (so a conditional GET keeps hitting `304` behind a load balancer, addressing the earlier round's note too). Existing stability test still passes.

**✅ [Should] non-UTF-8 docs.** Now `WARN`s (no longer silent) on invalid UTF-8 — but I kept the lossy parse rather than skipping: dropping a real emergency alert over an encoding issue seems worse than serving it with the load-bearing fields (geometry/severity/times — all ASCII/numeric) intact and only free text possibly carrying a U+FFFD. The WARN removes the "silent" part. Open to switching to skip-entirely if you'd prefer.

**✅ [Nit] `cap_xml`** now XML-escapes the interpolated identifier.

27 unit + 23 integration tests pass, clippy clean.

mrauhala:
Thanks — this reads as a convergence review (all prior fixes verified). Both remaining findings are non-blocking and need no change here:

**1. `raster_info()` O(times) clone.** Agreed it's the shared `MapEngine::raster_info(&self) -> RasterInfo` by-value shape — **every** raster engine (geotiff/grib/querydata/zarr/odim/cap) clones, so the fix belongs cross-engine, not in this PR. Filed **#432** to track it (favoring the lower-risk option: intern the heavy `RasterInfo` fields as `Arc<[…]>` so a by-value clone becomes pointer copies). engine-cap already serves from a prebuilt `Arc<RasterInfo>` snapshot, so it'll benefit for free once #432 lands.

**2. Redirect SSRF** — tracked in **#431** (ds-storage-wide redirect-disable); documentation is accurate, trust model matches every other HTTP-backed engine. No change.

No outstanding blocking items.

PR #434, opened 2026-06-15
feat(engine-postgis): optional stations table — orphan / observations-derived locations (#433)

### LINKED ISSUE #433: engine-postgis: make the stations table optional (orphan-station / observations-derived locations)
## Problem

The PostGIS engine requires every observed station to have a row in the configured `stations` table. Observations whose `station_fk` is absent there are invisible (locations list, Features, position/area all go through the stations table).

Live on nexus (`fmi-obs`, TimescaleDB `per_parameter`), this is a large gap: `airtemperature` has **8655 distinct `wigos_id`, of which 7206 are orphans** (no `public.stations` row) — genuine unregistered stations (e.g. Chad `0-148-0-CHD…`) with valid `the_geom`. ~83% of observed stations are undisplayable.

## Solution

Make `[postgis.stations]` optional and derive locations from the observations table's own geometry (`observations.geom_col`). Two modes, one mechanism:

- **A — observations-only:** no stations block; all locations from the obs geometry.
- **B — orphan fallback:** stations supply label/properties for the registered subset; orphans surface from the obs geometry (`label = id`, empty properties).

Mode is derived (no flag) from stations-presence × obs-geom; no stations + no geom = hard config error.

### Design
- New `LocationSource` enum gates the metadata refresh and position/area; the per-request observation fetch (`WHERE station_fk = $1`) is unchanged in every mode → **request latency unchanged**.
- Location derivation: `SELECT DISTINCT ON (station_fk)` (per_parameter `UNION`s tables → outer `DISTINCT ON (id)`), capped at `MAX_LOCATIONS`. On TimescaleDB the unique `(station_fk, time)` index drives a SkipScan (~2 s over 12.7M rows), in the background refresh only.
- Modes A/B answer position (haversine nearest) and area (bbox) **in-memory** from the cached location set — area is a bbox superset, not exact `ST_Within` (documented v1 simplification). Stations-only keeps the live `ST_DWithin`/`ST_Within` SQL path.

### Scaling
For very large tables, pre-materialize a distinct `(id, geom)` view wired as the `stations` block (back onto the indexed fast path), and/or raise `metadata_refresh_secs`.

### Out of scope (v1)
- Exact in-memory polygon containment for area (bbox only).
- lat/lon-column geometry (single PostGIS geometry column only; use a VIEW with `ST_MakePoint`).
- Live-DB integration test harness (relies on SQL-string + pure-function unit tests, the crate's existing convention).

mrauhala:
## What

Makes the PostGIS `[postgis.stations]` table **optional**. When the observation tables carry their own point geometry (`observations.geom_col`), locations are derived from the observations themselves. Two modes, one mechanism (derived from config, no flag):

| `[postgis.stations]` | `observations.geom_col` | Mode | Behavior |
|---|---|---|---|
| present | absent | stations-only | Original behavior (unchanged). |
| present | present | **orphan fallback** | Registered stations keep label/properties; observations whose `station_fk` has no stations row still surface from the obs geometry (`label = id`, empty props). |
| absent | present | **observations-only** | Every location derived from the obs geometry. |
| absent | absent | — | Hard config error. |

Closes #433.

## Why

Live on nexus (`fmi-obs`, TimescaleDB `per_parameter`): `airtemperature` has **8655 distinct `wigos_id`, of which 7206 are orphans** — genuine unregistered stations (e.g. Chad `0-148-0-CHD…`) with valid `the_geom` but no `public.stations` row. ~83% of observed stations were undisplayable. Enabling mode B for fmi-obs is a one-line config change (`geom_col = "the_geom"`).

## Design / performance

- New `LocationSource` enum (`Stations` / `Observations` / `StationsWithOrphans`) gates the **metadata refresh** and **position/area** only. The per-request observation fetch (`WHERE station_fk = $1`) is identical in every mode → **request latency unchanged**.
- Location derivation: `SELECT DISTINCT ON (station_fk)` (per_parameter `UNION`s the tables → outer `DISTINCT ON (id)`), capped at `MAX_LOCATIONS`, `WHERE geom IS NOT NULL`. On TimescaleDB the unique `(station_fk, time)` index drives a **SkipScan (~2 s over a 12.7M-row hypertable)** — measured live; runs in the background refresh only.
- Modes A/B answer **position (haversine nearest)** and **area (bbox)** in-memory from the cached location set. ⚠️ area is a **bounding-box superset**, not exact `ST_Within` (documented v1 simplification); stations-only mode keeps the live `ST_DWithin`/`ST_Within` SQL path.
- Mode B merges **stations-first** so a registered station always wins over an orphan with the same id.

## Security

New identifiers go through the existing whitelist + `quote_ident`; new SQL keeps `SELECT` in plain literals (no verb in `format!`/`push_str(var)`). `scripts/check_sql_safety.sh` clean.

## Tests

- query: SQL-string asserts for `build_locations_from_observations` (DISTINCT ON, UNION, null-geom filter, cap, no props) + `NoStations`/`NoObservationGeom` guards.
- config (ds-core + engine): mode resolution for all four combos.
- metadata: orphan-merge precedence (registered wins, orphan gets id-as-label + empty props).
- engine: in-memory nearest within/outside radius; `area_to_bbox` bbox + polygon→bbox + garbage.
- feature: orphan → Point Feature with empty properties.

Full workspace `cargo test` green; clippy + fmt + SQL-safety clean.

## Verification against nexus (post-merge)

Add `geom_col = "the_geom"` under `[postgis.observations]` in `collections.d/fmi-obs.toml` (mode B), deploy, reload:
- `/edr/collections/fmi-obs/locations` count ~6230 → ~8655.
- `query_location` for an orphan `0-148-0-CHD64901}` returns air_temperature; a registered id still shows `name`/`territory`.
- position near Chad + area bbox over Africa return orphan stations.

## Out of scope (v1)

Exact in-memory polygon containment for area; lat/lon-column geometry (single PostGIS geometry column only — use a VIEW); live-DB integration harness.


mrauhala:
Addressed the review findings in ce526af:

1. **`merge_orphans` cap (metadata.rs)** — re-cap the merged stations+orphans set at `MAX_LOCATIONS` with a WARN on truncation, so mode B honours the same documented protective limit as the single-source paths.
2. **`fetch_station_rows` (metadata.rs)** — dropped the redundant `stations` param and the misleading `_cfg` name; the mapping is derived once from `cfg.location_source.stations()`, removing the divergence risk.
3. **`area_to_bbox` case-sensitivity (engine.rs)** — accepts a lowercase `polygon(...)` keyword now (case-insensitive prefix check + token skip); added a regression assertion.

`cargo test -p engine-postgis` (96), clippy, `cargo fmt`, and `scripts/check_sql_safety.sh` all clean.

mrauhala:
Force-pushed a cleaned branch. Two things:

**Removed accidentally-committed files.** An earlier `git add -A` swept in ~280 MB of pre-existing untracked `testdata/` radar fixtures plus `collections.d/cap-alerts-feed.toml` and `collections.d/radar-dk-volume-local-h5.toml` — none related to this feature. The branch now contains **only the 9 feature files** (the 7 engine/core sources + CLAUDE.md + README). This also resolves review finding #3 (the live MeteoAlarm feed URL in `cap-alerts-feed.toml` — that file is no longer in the PR).

**Review round-2 fixes** (commit 1b03da3):
1. **In-memory `query_area` (engine.rs)** — added `.take(MAX_STATIONS_IN_POLYGON)` so a large bbox over a dense dataset stops materialising at the cap instead of allocating the full match set before the error check, mirroring the SQL path's `LIMIT`.
2. **`UNION ALL` (query.rs)** — the per-parameter obs-locations branches now use `UNION ALL`; the outer `DISTINCT ON (id)` already dedups, so per-branch dedup was wasted work. Test updated to assert `UNION ALL`.

Round-1 fixes (merge cap, `fetch_station_rows` tidy, case-insensitive POLYGON) remain folded into the feature commit.

`cargo test -p engine-postgis` (96), clippy, fmt, and `scripts/check_sql_safety.sh` all clean.

PR #435, opened 2026-06-16
fix(engine-postgis): derive obs-locations one query per table — avoid statement_timeout (#433)

mrauhala:
## Problem (prod incident)

After deploying #434 and enabling mode B on nexus `fmi-obs` (`geom_col = "the_geom"`), the collection went **degraded with an empty EDR extent**. Root cause: the observations-derived location query for the `per_parameter` shape was a **single 6-table `UNION ALL`**. Each per-table `DISTINCT ON (station_fk)` scan is ~1.5 s over a 12.7 M-row TimescaleDB hypertable, so the union ran **~12 s in one statement** — which the read-only role's `statement_timeout` cancels (`meteocore_ro` = **5 s**). The metadata refresh failed → cache left at its empty boot state → empty extent.

Confirmed live as `meteocore_ro`: the 6-table union cancels at exactly **5001 ms**; a single table is **~1.5 s**. My original benchmark ran as the `meteo` superuser (no `statement_timeout`), which is why this slipped through.

## Fix

`build_locations_from_observations` now returns **one `BuiltQuery` per observation table** instead of a single union. `fetch_observation_rows` runs them on one connection and **deduplicates by id across tables in Rust** (first geometry wins — the cross-table dedup the union's outer `DISTINCT ON (id)` previously did), capped at `MAX_LOCATIONS`. Each statement is now ~1.5 s, comfortably under a 5 s role cap. `long`/`wide` are unaffected (already single-table). No SQL `UNION` remains.

## Tests

Updated query-builder tests to assert one query per table (no `UNION`); `long`/inherited-`geom_col` tests adjusted for the `Vec<BuiltQuery>` return. `cargo test -p engine-postgis` (97) green; clippy, fmt, `scripts/check_sql_safety.sh` clean.

## Prod state

Service already **restored** on nexus by reverting the `geom_col` edit (stations-only mode; 6230 locations, extent populated). After this merges + deploys, re-add `geom_col = "the_geom"` to enable orphan fallback — the per-table derivation (6 × ~1.5 s) stays under the 5 s cap.

## Follow-up (not in this PR)

Two adjacent robustness ideas worth filing: (a) `fetch_*` discards the real Postgres error via `map_err(|_| Db)` — surfacing it (at least in logs) would have made this a 1-minute diagnosis; (b) mode B could degrade gracefully (serve registered stations if the orphan derivation fails) instead of failing the whole refresh.


mrauhala:
Folded in the review notes (311f8d9):

- **Stale CLAUDE.md** — updated the PostGIS notes: per_parameter derivation is now one-query-per-table (no `UNION`), with the `statement_timeout` rationale.
- **Truncation WARN** — added `tables = queries.len()` context.
- **Buffering invariant** — documented on `fetch_observation_rows` that total rows fetched can reach `tables × MAX_LOCATIONS` (peak buffered ≈ one table; accumulated set bounded by the early break), vs. the old single-UNION cap.
- **Nit** — use the imported `HashSet` instead of the fully-qualified path.

Left as filed follow-ups (not in this PR): the `map_err(|_| Db)` error-swallowing → **#436**, and graceful mode-B degradation → **#437**.

`cargo test -p engine-postgis` (97), clippy, fmt, SQL-safety clean.

PR #438, opened 2026-06-16
feat(engine-postgis): time-windowed obs-location derivation — configurable, default 24h (#433)

mrauhala:
## Problem

Even after splitting the obs-location derivation per-table (#435), a **single** observation table's full-history `DISTINCT ON` can exceed a read-only role's `statement_timeout`. On nexus the largest table (`relative_humidity`, ~13 M rows / 2.8 GB) takes **>5 s** as `meteocore_ro` (5 s cap) → metadata refresh fails → `fmi-obs` degraded with an empty extent. Per-table fixed the *sum* but not the biggest table.

## Fix

Derive the location list from a **recent time window** instead of full history (your call — also more correct for a live obs collection). `observations.locations_window` (ISO 8601 duration, **default `"PT24H"`**; `"all"` = full history) adds `AND time_col >= now() - window` to each per-table `DISTINCT ON`, so the scan only touches recent hypertable chunks.

**Measured on the table that was timing out (`relative_humidity`):** full history >5 s (canceled) → **24h window ≈ 0.13 s** (7707 recent stations). Chunk exclusion does the work.

A future **climate collection** sets `locations_window = "all"` (and then needs a role `statement_timeout` big enough to scan all history, or a pre-materialized locations table — limits come from the role, by design).

## Changes

- **ds-core**: `PostgisObservationsConfig.locations_window`; validated at config load (ISO 8601 or `"all"`).
- **engine**: resolve to `Option<chrono::Duration>` (absent → 24h default; `"all"` → `None` = full history). `build_locations_from_observations` takes a `since: Option<DateTime<Utc>>` and emits `AND time_col >= $1`, wrapping the **bind** (not the column) via the existing `time_filter_rhs` so the column index / chunk-pruning stays usable. `fetch_observation_rows` computes `since = now - window`.
- The recent window also means the location list = **currently-reporting** stations (a station silent for >window drops off), which is the desired semantic for live obs.

## Tests

Window resolution (default 24h / `"all"` → None / `"PT12H"` / invalid → error); config-load validation rejects a bad duration; SQL builder includes the filter + bind when `since` is set, omits both when `None`. `cargo test -p engine-postgis -p ds-core` (367) green; clippy, fmt, `scripts/check_sql_safety.sh` clean.

## Rollout

After merge + deploy, set on nexus `fmi-obs` `[postgis.observations]`:
```toml
geom_col = "the_geom"
# locations_window defaults to 24h; no need to set it explicitly
```
Then locations = stations reporting in the last 24h (orphans included), refresh well under the 5 s cap. I'll run the smoke test on your signal.

Builds on #433/#434 (optional stations) and #435 (per-table split).


mrauhala:
Folded in the review notes:
- **Panic-safe window** (metadata.rs): `Utc::now().checked_sub_signed(window)` — an absurd misconfigured `locations_window` now degrades to full history instead of panicking the background refresh.
- **Long/Wide `since` coverage** (query.rs): added recent-window tests for the Long (naive-UTC → `AT TIME ZONE` bind) and Wide (timestamptz → bare bind) shapes; previously only per_parameter was tested.

The upgrade-behavior note (existing `geom_col` collections silently switch full-history → 24h) is intentional and documented in CLAUDE.md + the crate README. `cargo test -p engine-postgis` (103), clippy, fmt, SQL-safety clean.

Issue #110, opened 2026-04-20
engine-postgis: health + metrics integration


mrauhala:
Part of #99.

## \`/health\`

- \`SELECT 1\` at construction with 2s deadline decides initial status
- Background ping every 30s flips between \`ready\` and \`degraded\`
- \`failed\` means couldn't load (config, privilege, or first ping all failed)
- Never returns DSN details; error-class enum only

## \`/metrics\` (Prometheus)

New series:

- \`postgis_pool_size{pool_key}\`, \`postgis_pool_idle{pool_key}\`, \`postgis_pool_waiting{pool_key}\` — labelled by pool, not collection
- \`postgis_query_duration_seconds{collection, query_type}\` — histogram; \`query_type ∈ {position, locations, area, features, ping}\`
- \`postgis_query_errors_total{collection, kind}\`
- \`postgis_rows_returned{collection, query_type}\` — histogram with buckets around the 10001 cap
- \`postgis_metadata_refresh_duration_seconds{collection}\`

## pool_key cardinality (fresh-critique item 7)

\`pool_key = <user>@<host>:<port>/<db>\` by default. If hostname is ephemeral (k8s pods, etc.) operators override via a \`pool_label\` config field to avoid cardinality explosion.

PR #441, opened 2026-06-16
feat(engine-postgis): background metadata refresh loop (#110)

mrauhala:
## Problem

The PostGIS metadata cache — location list, spatial/temporal extents, and (now) the `locations_window` "currently reporting" set — was built **only at engine construction** (startup or `collections/reload`). `metadata_refresh_secs` was parsed but wired to nothing, so the location set went **stale until a manual reload**:

- a station that started or stopped reporting never appeared / dropped off,
- the 24h `locations_window` never actually *slid* (frozen at the last reload's `now()`),
- DB changes (e.g. the recent nexus brace-cleanup migration) were invisible until someone reloaded.

This is most acute for **mode B** (#440), whose membership is inherently time-sensitive. (Found while diagnosing exactly that on nexus — the brace cleanup didn't show up until a manual reload.)

## Fix

`PostgisEngine::poll_loop` — a watch-channel-shutdown loop mirroring the querydata/cap pattern:
- re-runs `refresh_metadata()` every `metadata_refresh_secs` (default 300 s) on the **dedicated background poll runtime** (never a request worker, #221),
- skips the first tick (construction already refreshed),
- a failed refresh logs a WARN and **keeps the previous snapshot** (`MetadataCache::refresh` only swaps on success) — a transient DB blip never empties the cache,
- `shutdown()` stops it on reload.

Wired into **both** the boot path (`main.rs`) and the reload path (`admin.rs`), with the matching `shutdown()` of old engines on reload — identical to every other poll-loop engine.

`/health` status remains boot-time only; live health-flip on sustained DB failure is the remaining part of #110 (noted in the construction WARN).

## Notes / out of scope

- The async refresh runs on the postgis deadpool pool from the background runtime — no `ds-storage`/`block_in_place` involved (that's only the sync request path), so it's safe per the engine concurrency rules.
- Drive-by observation (not fixed here): `cap_engines` is **also** missing from the boot-path poll-loop spawn in `main.rs` (it's only spawned on reload) — looks like a pre-existing oversight; happy to fix in a separate PR.

## Tests

`cargo test -p engine-postgis` (103) green; clippy (engine-postgis + server), fmt, SQL-safety clean. The loop is a timing construct (no live-DB unit test, matching the crate's convention); it'll be verified live on nexus.

Builds on #433/#434/#435/#438/#440.


mrauhala:
Good catch — fixed the real bug in 51d9500:

**shutdown-receiver race:** `poll_loop` now clones a retained version-0 `watch::Receiver` instead of calling `shutdown_tx.subscribe()`, so a `shutdown()` that fires before the spawned loop runs is no longer missed (the stale loop would otherwise survive the reload and keep hitting the DB). Mirrors engine-cap's documented pattern.

On the **`biased;`** nit: left it out to stay consistent with cap/querydata (neither uses it) — it's a harmless one-orphaned-refresh case and, as you noted, better as a cross-engine sweep than a one-engine divergence.

The `cap_engines` boot-spawn gap is filed as #442. `cargo test -p engine-postgis` (103), clippy, fmt clean.

mrauhala:
On the two remaining nits — `biased;` and `MissedTickBehavior::Skip`: both are valid but apply to **every** poll-loop engine equally (none currently sets either), so I'm not diverging postgis alone here. Tracked as a cross-engine sweep in **#443**. The real finding (the `subscribe()` rapid-reload race) is fixed in 51d9500. Merging.

Issue #444, opened 2026-06-16
engine-postgis: per-query metrics histograms (query_duration / rows_returned / query_errors) — needs API-layer recording


mrauhala:
Follow-up from the #110 health/metrics work (the PR adds live health + pool/refresh gauges). The remaining #110 metrics are the **per-query** series:

- `postgis_query_duration_seconds{collection, query_type}` — histogram (`query_type ∈ position/locations/area/features`)
- `postgis_rows_returned{collection, query_type}` — histogram (buckets around the 10001 cap)
- `postgis_query_errors_total{collection, kind}` — counter

## Why deferred

These need to time/observe **each request** and can't be pulled from engine state at scrape time like the gauges:

- The **engine** doesn't depend on `prometheus`, and the metric objects are server-global statics the engine can't reach (server depends on engine, not vice versa). Counters could be snapshot-exposed (the odim pattern), but **histograms can't be snapshot-and-replayed**.
- The **API layer** (`api-edr` / `api-features`) calls engines **generically** through the `EdrEngine`/`FeatureEngine` traits — it doesn't know which collections are postgis, so `postgis_*` recorded there would mislabel every engine.

## Options

1. **Generic EDR/Features query metrics** (`edr_query_duration_seconds{collection, query_type, engine_type}`) recorded in the api handlers — covers all engines, arguably more useful; would supersede the `postgis_`-prefixed naming in #110. (Recommended.)
2. Give `engine-postgis` a `prometheus` dep + its own registry the server merges (breaks the "engines don't do metrics" convention).
3. Pass metric handles into the engine from the server (awkward layering).

Needs a small design note before implementing. Already shipped for #110: live `SELECT 1` health ping (ready↔degraded), `postgis_up`, `postgis_pool_*`, `postgis_metadata_refresh_seconds`, refresh/ping failure gauges.

PR #445, opened 2026-06-16
feat(engine-postgis): live health monitoring + ops metrics (#110)

mrauhala:
## What

Implements live health monitoring + the gauge-based `/metrics` surface for PostGIS collections — the part of #110 the construction WARN flagged as missing ("a DB that goes down after startup will still show ready").

### `/health` — live `SELECT 1` ping
`PostgisEngine::poll_loop` now also runs a `SELECT 1` ping every **30 s** (2 s deadline), flipping a loaded collection `ready`↔`degraded` on DB reachability. The `/health` handler overrides the boot snapshot with each postgis engine's **live** status (`engine.health_status()`). A `failed` collection has no engine (couldn't construct) and keeps its boot status. The ping fires immediately at loop start, so `/health` is accurate within ~2 s of boot.

**The ping is the `/health` authority** — a *metadata-refresh* failure keeps the last good snapshot and does **not** degrade health (it's surfaced via metrics + logs instead). This avoids flapping when a heavy refresh times out but the DB is reachable.

### `/metrics` — live gauges (no delta-state)
- `postgis_up{collection}` — 1 ready / 0 degraded (the alerting metric)
- `postgis_pool_{size,idle,waiting}{pool_key}` — deadpool status
- `postgis_metadata_refresh_seconds{collection}` — last refresh duration (would have made the earlier statement-timeout incident obvious)
- `postgis_{metadata_refreshes,metadata_refresh_failures,ping_failures}{collection}` — counts since engine start (reset on reload)

All true gauges, set per-scrape from live engine state — no counter delta-tracking needed.

## Design notes
- `engine-postgis` keeps its own 2-state `HealthStatus` (it can't depend on the server crate for `CollectionStatus`); the handler maps at the boundary.
- The async ping runs on the existing background poll runtime (never a request worker, #221); no `ds-storage`/`block_in_place` involved.
- `pool_key` label = `pool_label` else the derived `<user>@<host>:<port>/<db>` (no password).

## Deferred (tracked in #444)
The per-query histograms (`postgis_query_duration_seconds` / `rows_returned` / `query_errors_total`) need API-layer recording — the API calls engines **generically** through the trait, and histograms can't be snapshot-scraped like the gauges. Likely better as generic cross-engine EDR metrics; needs a design note.

## Tests
`health.rs` unit tests (starts ready; ping flips status + counts failures; refresh records metrics but not status). `cargo test -p engine-postgis` (106) green; clippy `--all-targets` (engine-postgis + server), fmt, SQL-safety clean.

Builds on the #441 background refresh loop.


mrauhala:
Both review findings fixed in e7e62ab:
1. **Metric types** — `postgis_metadata_refreshes` / `metadata_refresh_failures` / `ping_failures` are now `IntCounterVec` with the `_total` suffix, scraped via rebaseline-on-reset delta-tracking (the codebase's existing replaceable-cache pattern), so they stay monotonic across reloads and `rate()`/`increase()` behave. `postgis_up`, `postgis_pool_*`, and `postgis_metadata_refresh_seconds` remain gauges (correctly).
2. **Stale CLAUDE.md clause** — dropped "/health status is still boot-time only".

`cargo test -p engine-postgis` (106), clippy `--all-targets`, fmt, SQL-safety clean.

mrauhala:
Fixed all three round-2 findings in 3bc0a26:
1. **False-Degraded under pool exhaustion** — `ping()` now uses a **dedicated** `tokio_postgres::connect` (NoTls), not `pool.get()`, so a pool fully checked out by request handlers no longer masquerades as DB-unreachable. The probe is pure reachability; pool saturation shows up in `postgis_pool_waiting`.
2. **`postgis_pool_size`** now reports `st.size` (open connections), and a new **`postgis_pool_max_size`** carries the capacity.
3. **`postgis_pool_idle` → `postgis_pool_available`** (deadpool's `available` = open-idle + unallocated slots; "idle" was misleading at startup).

`cargo test -p engine-postgis` (106), clippy `--all-targets`, fmt, SQL-safety clean.

mrauhala:
Round-3 fix in 1cc85d5: **ping driver-leak**. `ping()` now drives the dedicated connection inline (`tokio::pin!` + `select!`) instead of `tokio::spawn` + `abort()` — on the outer timeout the query and the connection driver drop together, with no detached/orphaned task. Also saturated the `as_millis()`→`u64` cast (rode along). `cargo test -p engine-postgis` (106), clippy `--all-targets`, fmt clean.

mrauhala:
Round-4 fix in bb16986: **boot-degraded status no longer cleared before the first ping**. Added a `probed` flag — `live_health()` returns `None` until the first `SELECT 1` completes, so the `/health` handler keeps the boot snapshot (which reflects the boot refresh outcome) during the load→first-ping window instead of the optimistic `ready` seed. New unit test covers it. `cargo test -p engine-postgis` (107), clippy `--all-targets`, fmt clean.

mrauhala:
Round-5 fixes in 69af052:
1. **Transition logging** — `record_ping` returns the transition; `poll_loop` logs `warn!` on Ready→Degraded and `info!` on recovery, once per transition (not per tick), so log pipelines get a DB-down/up signal. New test covers the transition reporting.
2. **`postgis_up` consistency** — `HealthSnapshot` gains `probed`; the metric is emitted only after the first ping, so it no longer reads `1` for a boot-degraded collection while `/health` shows `degraded` during the load→first-ping window.

`cargo test -p engine-postgis` (108), clippy `--all-targets`, fmt, SQL-safety clean.

mrauhala:
Round-6 in 94c4cc5:
1. **`postgis_pings_total`** — added the total alongside `ping_failures_total` so a failure *ratio* (`increase(failures)/increase(total)`) is computable; symmetric with the refresh counters. Threaded through the snapshot + the rebaseline delta-tracking.
2. **TLS divergence** — added a `TODO(#110)` on the ping's `NoTls` so the future TLS work keeps the ping connector in sync with the pool (else a TLS-required server falsely degrades a healthy collection).

`cargo test -p engine-postgis` (108), `cargo build -p server`, clippy `--all-targets`, fmt, SQL-safety clean.

mrauhala:
Round-7 review **passed**. The three findings are all pre-existing/latent/systemic, not new actionable bugs in this PR's shipping code, so I'm not fixing them here — tracked in **#446**:
1. **Stale `postgis_up` after collection removal** — the *same* per-collection-label-eviction pattern the round-5 review explicitly called "pre-existing… not new here"; it affects every engine's per-collection metrics (tile/grid/…), so a postgis-only fix would diverge it. → #446.
2. **`pool_key_label` omits `sslmode`** — explicitly latent (zero effect under `NoTls`); belongs with the TLS work in #110 where it's testable. → noted in #446.
3. **`counter_state.postgis` unbounded growth** — minor, same pre-existing pattern as `.tile`/`.grid`. → #446.

The loop converged: 6 substantive rounds (metric types, ping connection + pool semantics, task-leak, boot-degraded window, transition logging + metric consistency, ratio counter + TLS note) all fixed, and round 7 surfaced only pre-existing/latent items. Merging.

PR #447, opened 2026-06-17
feat(engine-odim PVOL): configurable render resampling, default nearest (#186)

mrauhala:
## Summary

The per-site PVOL Cartesian render (`polar_sample` — WMS/Maps/Tiles for `engine_type = "odim-volume"`) has sampled **bilinearly** since #186, which blends the four surrounding `(ray, bin)` cells to close the radial bin/wedge structure far from the radar. That smoothing also **softens peaks and erases the discrete radar bins** — detail that *is* visible in the COMP composite (nearest) and GeoTIFF (nearest) renders, so per-site `DBZH` layers look blurry by comparison (user-reported).

This adds a per-collection toggle and **restores nearest-neighbour as the default**:

```toml
[odim]
# resampling = "nearest"   # default — discrete bins visible, full detail
resampling = "bilinear"    # opt back into the #186 smoothing
```

`nearest` was the pre-#186 default: each output pixel takes its single enclosing polar cell, so cells tile the output plane and widen with range — the bins stay visible and every gate's value is preserved, with no gaps (there never were empty gaps; #186's "radial spokes" were this discrete-bin structure itself). `bilinear` keeps the #186 look available per collection.

## Changes

- New `ds_core::config::ResamplingMethod` enum (serde lowercase; `Default = Nearest`); `#[serde(default)]` field `resampling` on `OdimConfig`.
- Threaded engine → every `PolarVolumeSiteView` → `polar_sample`, which now branches per pixel: **nearest** = `floor` of the fractional `(ray, bin)` (ray wraps at the 0°/360° seam); **bilinear** = the existing mask-aware `bilinear_cell`.
- **Unaffected** (verified against the code): COMP composite render (always nearest), EDR position/area queries (always nearest — a point query wants the measurement, not a cosmetic blend), and the 3D Tiles voxel / isosurface / echo-top products (those sample nearest then apply a *separate* `smooth_grid` 3-D blur in `ds-3dtiles`, #381; the point cloud is raw). So this toggle is map-render-only.
- Docs: `OdimConfig`/`ResamplingMethod` rustdoc + a CLAUDE.md PVOL note.

## Test plan

- [x] `cargo test -p engine-odim` — 32 integration + lib tests pass.
- [x] New `polar_sample_nearest_does_not_blend_rays` locks in the default (ray-only field → every sampled value is an exact integer ray index, no blending); existing bilinear tests now pin `ResamplingMethod::Bilinear` explicitly (incl. the per-pixel-reference and blend-across-rays tests).
- [x] `cargo clippy -p engine-odim --all-targets` clean; `cargo fmt --check` clean.
- [x] `cargo check --workspace --tests --examples` clean.


mrauhala:
Thanks — addressed all three findings (force-pushed `bf84200`):

1. **`collections.d/cap-alerts-feed.toml` (+ ~45 other files) — removed.** These were untracked local files (binary radar fixtures, demo configs) accidentally swept into the commit by a `git add -A`; they were never part of this change. Force-pushed a commit containing only the 8 intended files. Good catch — that would have shipped hundreds of MB of binaries and an external network dependency.

2. **`volume_engine.rs` nearest arm — `debug_assert!` added.** The `bin_f >= nbins` guard earlier in the loop does hold the invariant (so the path was safe), but the assert now documents and enforces it in debug builds, per your suggestion.

3. **COMP ignores `resampling` — startup WARN added.** `OdimEngine::new` now logs a `tracing::warn!` when `config.resampling != ResamplingMethod::default()`, mirroring the existing `time_window`-ignored warning, so a misconfigured `[odim] resampling` on a COMP collection surfaces instead of silently no-op'ing.

`cargo test -p engine-odim` (incl. the 6 `polar_sample` tests), clippy `--all-targets`, and `fmt --check` all clean.

PR #448, opened 2026-06-17
fix(engine-geotiff): low-zoom reprojection artifacts (ghosts + curtains)

mrauhala:
## Problem

Projected-CRS radar composites served via the GeoTIFF engine (FMI EPSG:3067 TM35FIN, Norway LCC — **different projections, same bug**) showed two distinct artifacts at low zoom. Reproduced server-side (production WMS + tiles), **not** a frontend issue.

| | symptom |
|---|---|
| **Curtains** | dotty / "curtain" misregistration along curved edges (worst on the east), zoom-dependent — correct when zoomed in |
| **Ghosts** | echoes painted far from the data (Arctic / Antarctic corners) at extreme zoom-out; "echoes lower south into place as you zoom in" |

Both are rooted in the shared `ds_core::resample::ProjectionGrid`.

## Root cause & fix

**1. Curtains — error estimator underestimated → premature refinement stop.**
`estimate_error` probed only 5 near-centre points per cell, so it under-reported the true bilinear error by up to ~20× (0.08px reported vs **1.62px** real at production resolution). The adaptive refinement converged far too early, leaving ~1.6 source-px misregistration — invisible on low-res test fixtures, visible in production. Replaced with a regular, edge-reaching **4×4 probe grid** that brackets the residual peak. Projection-agnostic → fixes every projected source.

**2. Ghosts — out-of-domain aliasing at extreme zoom.**
At low zoom the coarse grid can't converge (hits the cell cap) and out-of-world output pixels (longitudes wrapping past ±180°, near-pole latitudes) alias onto valid raster pixels. Added a cheap **domain guard** in `get_raster_tile` (`footprint_pixel_window`): bounds — in output-pixel space, from the source's WGS84 envelope — the window the footprint can occupy, dropping everything outside to nodata. No per-pixel projection (~130 perimeter points once + integer compares).

## Performance

| View | OLD (broken) | NEW (4×4) | Δ |
|---|---|---|---|
| Tight tile 256×256 (hot path) | 4470µs | 3756µs | ~0% (noise) |
| Med viewport 1024×768 | 19.6ms | 16.0ms | **−18%** (guard skips out-of-footprint pixels) |
| Wide zoomed-out | 15.7ms | 28.3ms | +80% |
| Extreme whole-world | 87ms | 191ms | +119% |

Hot path unchanged; extra cost lands only on zoomed-out views that were previously *broken*. 4×4 (vs 5×5) probes halve the zoomed-out cost with no accuracy loss.

## Tests

- `ds-core`: `grid_holds_budget_at_production_resolution` — pins the estimator at production resolution (verified to **fail** with the old 5-probe scheme, 1.68px > budget; passes now).
- `engine-geotiff`: `low_zoom_domain_guard` — end-to-end ghost guard on the committed TM35FIN fixture (extreme + wide + hemispheric viewports → 0 ghosts, footprint still renders).
- All existing `render_projection` / `resample` tests pass; clippy clean.

## Follow-up

The domain guard is currently GeoTIFF-only. ODIM composite (stereographic) and Zarr (projected) use the same `ProjectionGrid` and benefit from the estimator fix, but not the ghost guard — tracked in a separate issue.

### DIFF
--- a/crates/core/src/resample.rs
+++ b/crates/core/src/resample.rs
@@ -47,19 +47,39 @@ const MAX_CELLS: u32 = 256;
 /// the 0.5 px resolution of nearest-neighbour resampling.
 const MAX_INTERP_ERROR_PX: f64 = 0.2;
 
-/// Interior probe points `(tx, ty)` within a cell, used by `estimate_error`.
-/// The centre is the dominant (quadratic) bilinear-residual peak; the quarter
-/// points catch the asymmetric higher-order part that a real projection adds
-/// and the centre alone would miss. Sampling several points per cell also
-/// means a cell counts toward the estimate if *any* probe lands on the raster,
-/// not just its centre.
-const ERROR_PROBES: [(f64, f64); 5] = [
-    (0.5, 0.5),
-    (0.25, 0.25),
-    (0.75, 0.25),
-    (0.25, 0.75),
-    (0.75, 0.75),
-];
+/// Per-axis fractions of the 2-D grid of interior probe points `(tx, ty)` that
+/// [`ProjectionGrid::estimate_error`] samples within each cell (the cartesian
+/// product gives `PROBE_FRACS.len()²` points).
+///
+/// A **dense, regular** grid is load-bearing for correctness, not just nicety.
+/// The bilinear residual is zero at the cell corners and peaks somewhere in the
+/// interior, but for a real projection over a wide cell that peak is *not* at
+/// the centre — its location depends on the projection's higher-order terms and
+/// shifts toward the edges. The original 5-point pattern (centre + four quarter
+/// points, all clustered near the middle) systematically *missed* the peak and
+/// under-reported the true error by up to ~20× for low-zoom viewports, so
+/// [`ProjectionGrid::build`] stopped refining far too early and left several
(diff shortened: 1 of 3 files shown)


mrauhala:
Addressed review: added `debug_assert!(dim > 0)` in `footprint_pixel_window`'s `to_px` closure (fb2989f) so the latent `clamp(0.0, -1.0)` debug panic for `dim == 0` is made explicit. Unreachable in practice (callers pass validated positive dims), but the contract is now asserted.

mrauhala:
Fixed in 901de8e: corrected the comment — `f64::clamp` panics on `min > max` in **both** debug and release; the `debug_assert!` just fires first with a clearer message. Comment-only change.

mrauhala:
Re the two latest nits — leaving both as-is:

- **`debug_assert!` vs `assert!`**: in release, `dim == 0` panics via `clamp(0.0, -1.0)` regardless; `assert!` would only improve the message for an already-unreachable path (validated API dims). `debug_assert!` is the idiomatic choice for a can't-happen internal invariant, so keeping it.
- **Test bounds coupled to the fixture**: intentional generous slack around the committed `radar-tm35fin` footprint, already documented in the test's header comment. Low risk; leaving as-is.

All substantive review feedback addressed; CI green.

Issue #449, opened 2026-06-17
Generalize low-zoom domain guard into ds-core (ODIM composite + Zarr)


mrauhala:
Follow-up to #448.

## Background

#448 fixed two low-zoom reprojection artifacts in the **GeoTIFF** engine:

1. **Curtains / under-refinement** — fixed in the *shared* `ds_core::resample::ProjectionGrid` (denser error-estimator probe grid), so **every** engine that uses `ProjectionGrid` already benefits (ODIM composite, Zarr, GeoTIFF).
2. **Ghosts** (data aliased far from the footprint at extreme zoom-out — wrapped longitudes / near-pole latitudes mapping onto valid raster pixels) — fixed only in `engine-geotiff` via a local `footprint_pixel_window` domain guard applied in `get_raster_tile`.

## Problem

The ghost guard (#2) is **GeoTIFF-only**. The same latent bug exists in the other projected raster engines that render through `ProjectionGrid`:

- **ODIM composite** (`engine-odim`, oblique-stereographic composite render path)
- **Zarr** (`engine-zarr`, projected output via `ProjectionGrid`)

At a whole-world / extreme zoom-out, these can still paint ghost echoes outside the source footprint. It wasn't reported for them (only FMI EPSG:3067 + Norway LCC GeoTIFF were), so it was left out of #448 to keep the fix scoped.

## Proposal

Lift the domain guard into a shared, tested ds-core helper and adopt it in every projected raster engine:

- Move the `footprint_pixel_window` logic (output-fraction bounds derived from the source's WGS84 envelope, via `OutputCrs::world_to_fraction`) into `ds-core` (e.g. alongside `ProjectionGrid` in `resample.rs`, or `map_engine`/`geo`). Keep it framework-free.
- Cover it with a unit test (the GeoTIFF integration test `low_zoom_domain_guard` can stay as an e2e check).
- Adopt it in `engine-geotiff` (replace the local copy), `engine-odim` (composite render), and `engine-zarr`.

### Notes / known limitation to carry over
- The guard is a single output-space box, so a viewport showing >1 world copy (Web Mercator spanning >360° longitude) keeps only the primary copy of the footprint; additional wrapped copies render as nodata. Acceptable (vs ghost aliasing) — document it in the shared helper.
- No per-pixel projection (perimeter sampling once + integer compares per pixel), per the #203 "never project per output pixel" rule.

## Acceptance
[...]

PR #452, opened 2026-06-17
fix(render): don't clamp meta-tile viewport bounds to ±85° (north displacement)

mrauhala:
## Problem

The projection fix (#448) corrected the **direct** `get_raster_tile` path, but WMS GetMap in **EPSG:3857** goes through the **meta-tile** path — and it still displaced the radar composite ~10° **north** on zoomed-out views. Confirmed on production after deploy: a tight view is correct (and now PNG8, #451), but an extreme/wide view places Finland at lat 66–77°N instead of 56–73°N.

## Cause

`render_metatiled` computes the viewport's Mercator-metre bounds with `lat_to_y`, which **clamps latitude to ±85.05°** (the tile-grid limit). For a bbox reaching lat 87°N, `north_m` clamps to `ORIGIN` (lat 85.05°). The assembly then maps output pixels over `[south_m, ORIGIN]`, while the **client places the returned image over the FULL requested extent** `[south_m, lat_to_y_unclamped(87°)]`. That vertical-scale mismatch shifts all content poleward.

Isolated by reproducing the meta-tile path locally:
- DIRECT `get_raster_tile` (full bbox): lat **58–70.6** ✓
- META-TILE assembly: lat **66.5–76.7** ✗
- Instrumented mosaic: data correctly at lat **58.7–70.7** — so the per-tile renders and the assembly *resample* are right; only the **bounds** were clamped.

## Fix

Compute `north_m`/`south_m` with **`merc_y_exact`** (the unclamped inverse of `y_to_lat`) so the viewport mapping matches the client's request bbox exactly even past ±85°. The ±85° clamp is kept where it belongs — tile **selection** (`row0`/`row1`) still uses `lat_to_y`, since tiles only exist within the valid world; output pixels past ±85° have no tile and assemble transparent (no phantom rows).

After the fix the meta-tile path renders lat **58.5–70.5**, matching the direct path.

## Test

`extreme_latitude_viewport_does_not_displace_data` renders a viewport reaching lat 87°N with a known data stripe at lat 55–70 and asserts it reads back at ~55–70. Verified to **fail** with the clamped bounds (stripe displaced to lat 62–75) and pass with the fix.

## Scope

Affects all WMS EPSG:3857 GetMap renders at zoomed-out extents (every MapEngine — radar, NWP, etc.). Maps/Tiles APIs don't use meta-tiling. Independent of the #449 footprint-guard work (separate PR).

### DIFF
--- a/crates/render/src/metatile.rs
+++ b/crates/render/src/metatile.rs
@@ -69,6 +69,22 @@ fn lat_to_y(lat_deg: f64) -> f64 {
     let lat = lat_deg.to_radians().clamp(-LAT_LIMIT_RAD, LAT_LIMIT_RAD);
     R * (std::f64::consts::FRAC_PI_4 + lat / 2.0).tan().ln()
 }
+/// Web Mercator northing for a latitude — the inverse of [`y_to_lat`], WITHOUT
+/// [`lat_to_y`]'s ±85° tile-grid clamp.
+///
+/// The meta-tile **viewport bounds** must match the client's request bbox
+/// exactly: a zoomed-out EPSG:3857 GetMap can have a bbox reaching past ±85°
+/// toward a pole, and the client places the returned image over the FULL
+/// requested extent. Using the ±85°-clamped [`lat_to_y`] for the bounds (as this
+/// did) shrinks the assembled image's vertical span relative to the client's, so
+/// all data is shifted toward the pole — the radar composite appeared ~10° too
+/// far north on whole-world views. Clamps just shy of ±90° only to keep `tan`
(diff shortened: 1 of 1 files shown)


mrauhala:
Re the test-panic message nit: `MetaTile` doesn't derive `Debug` (that's why the test uses `matches!`), so `{other:?}` won't compile. The message is only reached on a test *failure*; I've left it to avoid a CI/merge cycle on this priority fix. Will tidy it to name the variant explicitly if you'd like, or fold it into a later touch of this file.

PR #453, opened 2026-06-17
feat(render): generalize low-zoom footprint domain guard into ds-core (#449)

### LINKED ISSUE #449: Generalize low-zoom domain guard into ds-core (ODIM composite + Zarr)

mrauhala:
Closes #449.

## Summary

#448 added a per-engine "ghost guard" to `engine-geotiff`: at low zoom the coarse `ProjectionGrid` (and a projection's out-of-domain forward) can map a far-away output pixel onto valid source data, painting echoes far from the real coverage. This lifts that guard into `ds-core` so every projected raster engine shares it.

## Changes

- **ds-core**: `OutputCrs::footprint_pixel_window(wgs84_bbox, src_env_wgs84, w, h)` — bounds the output-pixel window the source footprint can occupy, mapping the envelope perimeter to output-fraction space via `world_to_fraction` (per-vertex, never per pixel — #203). Moved verbatim from engine-geotiff, now unit-tested (`footprint_window_bounds_a_small_source_in_a_wide_view`, `footprint_window_is_permissive_when_view_is_inside_the_source`).
- **engine-geotiff**: use the shared method; drop the local copy.
- **engine-odim (COMP)**: adopt it — the stereographic source aliases like TM, so this is a *real* fix for whole-world Web Mercator views.
- **engine-zarr (projected path)**: adopt it. Today's geographic Zarr grids use an affine/monotonic `frac_px` map that can't alias (out-of-range → nodata already), so it's belt-and-suspenders now; it becomes **load-bearing for Phase 4 projected source grids** (STAC per-item-CRS, e.g. HRRR/Lambert).

## Not changed: ODIM PVOL

Its exact `bin_f >= nbins` coverage-disk check plus great-circle ENU mapping already reject far/wrapped points (no truncated-series aliasing), so the bbox guard would be redundant. Noted in the issue.

## Notes

- Independent of the meta-tile ±85° displacement fix (separate PR #452).
- Existing `engine-geotiff` `low_zoom_domain_guard` integration test still passes against the shared method.

### DIFF
--- a/crates/core/src/map_engine.rs
+++ b/crates/core/src/map_engine.rs
@@ -132,7 +132,10 @@ impl OutputCrs {
     /// domain guard for projected-raster resampling.
     ///
     /// `src_env_wgs84` is the source raster's WGS84 envelope `[w, s, e, n]`;
-    /// `wgs84_bbox` is the requested tile bbox. The envelope perimeter is mapped
+    /// `wgs84_bbox` is the requested tile bbox (for `Wgs84`/`WebMercator` output;
+    /// it is **ignored for `Projected` output**, whose extents are embedded in the
+    /// `Projected { crs, bbox }` variant — same as [`Self::world_to_fraction`]).
+    /// The envelope perimeter is mapped
     /// to output-fraction space with [`Self::world_to_fraction`] (the per-vertex
     /// inverse of the per-pixel [`Self::project_node`] — only ~130 perimeter
     /// points, never per output pixel, per the #203 rule), the fractional extent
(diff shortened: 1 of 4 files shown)


mrauhala:
Addressed both review points in 7a5b5cc (docs only, no behavior change): the shared `footprint_pixel_window` docstring now states it (1) requires a true WGS84 envelope — a native-CRS extent would silently disable the guard, called out for Phase 4 projected Zarr sources — and (2) doesn't handle an antimeridian-crossing (`w>e`) envelope. Also clarified the Zarr call site's WGS84-invariant assumption.

mrauhala:
Addressed the re-review: added `footprint_window_bounds_source_for_projected_output` (unit-covers the `OutputCrs::Projected` EPSG:3067 path that ODIM COMP relies on for national-grid requests) and documented that `wgs84_bbox` is ignored for `Projected` output. Commit on branch.

PR #457, opened 2026-06-19
fix(engine-geotiff): footprint-MAX downsampling to stop dropping sparse radar echoes (#456)

### LINKED ISSUE #456: GeoTIFF: evaluate sparse-echo downsampling after the edge-stride fix
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep open with narrowed scope.

**Current assessment and challenge:** Original dramatic artefact was the fixed edge-tile stride bug; MAX downsampling was reverted for dilation. Retain only measured sparse-echo sampling tradeoff.

**Next step / remaining scope:** Reproduce any remaining sparse-echo loss and compare sampling methods for both missed echoes and dilation. Do not reinstate MAX based on the old stride artefact.

**Priority after triage:** medium.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Symptom

When a projected GeoTIFF radar composite (FMI EPSG:3067, MET Norway LCC) is rendered **zoomed out**, echoes past a sharp straight-line boundary break up into dotted **dashes / curtains** and large parts of the coverage vanish:

- FMI: east of ~33°E (toward/past the eastern edge of dense Finnish radar coverage).
- MET Norway: north of a parallel-ish line in northern Finland.

Zooming **in** is correct. `gdalwarp` from full resolution (heavy oversample + exact transform) is fully coherent, so the data is fine — the artefact is in our render path. This predates meta-tiling and is **not** displacement/reprojection/meta-tiling (those were #448/#451/#452/#453/#454).

## Root cause

`GeoTiffEngine::get_raster_tile` selects a COG overview, then resamples to the output with **plain nearest-neighbour**: each output pixel takes the single source pixel under its centre (`grid.sample` → `floor`). When the output is coarser than the chosen overview (downsampling — a measured 409×412 view reads a 732×1736 source window, ~7.5× oversample), nearest keeps 1 of every ~8 source pixels and **discards the rest**. For **sparse far-range echoes** (scattered returns at the coverage edge) the dropped pixels are exactly the echoes → regular dotted dashes; dense coverage over Finland survives because every output pixel's centre still lands on a populated cell — hence the sharp boundary.

## Fix

Reduce each output pixel over its full source **footprint** with a coverage-preserving discrete **MAX** (strongest echo wins) instead of nearest-picking one cell:

- **discrete** — result is an actual source sample (output value set ⊆ source value set) → PNG8 palette preserved, no new colours;
- **no smoothing** — no averaging/blending; peaks kept exactly (the operational reflectivity-downsample convention);
- **perf-neutral** — footprint pixels are already decoded in the read window; non-overlapping footprints visit each source pixel ~once, so total work ≈ the pixels already read (no extra I/O / decode / projection);
- **zoom-in unchanged** — when the footprint is ≤1 source pixel (output finer than source) it degenerates to exactly the current nearest behaviour.

Footprint extent comes from the local `ProjectionGrid` derivative (cheap bilerps, never per-pixel projection — #203). Bilinear is explicitly rejected (too many colours + smooths the radar field).

</details>

mrauhala:
Closes #456.

## Problem

At zoom-out, projected GeoTIFF radar composites (FMI EPSG:3067, MET Norway LCC) broke up into dotted **dashes / venetian-blind "curtains"** past the edge of dense coverage, and large parts of the echo field vanished — while **zoom-in was correct** and `gdalwarp` from full resolution was fully coherent. So the data is fine; the artefact was in our render path. (Distinct from the reprojection / meta-tiling fixes #448/#451/#452/#453/#454.)

## Root cause

`GeoTiffEngine::get_raster_tile` selects a COG overview, then resamples to the output with plain **nearest-neighbour** — each output pixel takes the single source pixel under its centre (`grid.sample` → `floor`). When the output is coarser than the chosen source (downsampling — a measured 409×412 view read a 732×1736 window, ~7.5×), nearest keeps **1 of every ~8** source pixels and discards the rest. For **sparse far-range echoes** the dropped pixels *are* the echoes → regular dashes; dense coverage survives because every output pixel's centre still lands on a populated cell — hence the sharp boundary.

## Fix

Reduce each output pixel over its full source **footprint** with a coverage-preserving **MAX** (strongest echo wins) instead of nearest-picking one cell:

- **discrete** — the result is an actual source sample (output value set ⊆ source value set), so the **PNG8 palette is preserved** (no new colours);
- **not smoothed** — no averaging/blending; peaks kept exactly (the operational reflectivity-downsample convention). Bilinear was rejected: it invents intermediate values (defeats PNG8) and smears the radar field;
- **perf-neutral** — the footprint pixels are already decoded in the read window; abutting footprints visit each source pixel ~once, so total work ≈ the pixels already read (no extra I/O / decode / projection);
- **zoom-in unchanged** — when the footprint is ≤1 source px (output finer than source) it collapses to *exactly* the previous nearest behaviour.

The per-axis footprint half-extent comes from the local `ProjectionGrid` derivative (cheap bilerps, never a per-pixel projection — #203). Two small free fns (`footprint_half`, `push_footprint_max`) carry the logic.

## Tests

9 new unit tests pin: collapse-to-nearest below 1px, growth with downsample factor, non-finite-neighbour fallback, byte-identical nearest at `half=0`, **a sparse echo a plain nearest would drop is preserved**, strongest-and-discrete, all-nodata→nodata, off-window→nodata, window clamping. Full `engine-geotiff` suite green (86 unit + integration).

## Render-verified

Before/after on the **FMI east-of-Finland** sparse echoes and the **MET Norway heavy-zoom-out** arc "curtains": both go from broken dashes to continuous coverage, still 8-bit PNG8.

**Known residual (separate, milder regime):** thin diagonal source artefacts (distant-radar spokes) at the mosaic edge can still alias into dashes at **~1:1 zoom**, where the footprint is ≤1px and nothing non-dilating / non-smoothing can bridge a sub-resolution feature. The dramatic heavy-downsample artefact — the actual "data unusable" complaint — is fixed.

### DIFF
--- a/crates/engine-geotiff/src/lib.rs
+++ b/crates/engine-geotiff/src/lib.rs
@@ -1375,6 +1375,74 @@ fn crs_label(crs: &ds_core::geo::Crs) -> String {
     }
 }
 
+/// Half-extent (in source pixels) of an output pixel's footprint on one source
+/// axis, from the forward grid derivative: `center` is the source coordinate at
+/// the output pixel centre, `fwd_x`/`fwd_y` the source coordinate one output
+/// pixel along +x / +y. The bounding span of the (parallelogram) footprint is
+/// `|∂/∂x| + |∂/∂y|`; half of that is the half-extent.
+///
+/// Returns `0.0` (⇒ a single nearest pixel) when the span is ≤ 1 source pixel
+/// (the output is NOT downsampling that axis — zoom-in stays byte-identical to
+/// plain nearest) or when a neighbour sample is non-finite (a projection
+/// domain-edge cell — fall back to nearest rather than a NaN box).
+#[inline]
+fn footprint_half(center: f64, fwd_x: f64, fwd_y: f64) -> f64 {
+    let half = 0.5 * ((fwd_x - center).abs() + (fwd_y - center).abs());
+    if half.is_finite() && half > 0.5 {
+        half
+    } else {
+        0.0
+    }
+}
+
+/// Reduce one output pixel over its source footprint with a coverage-preserving
+/// MAX and push the result. The footprint is the integer source-pixel box
+/// `[col_f ± half_c] × [row_f ± half_r]`, clamped to the window that was read
+/// (`col_start..col_end`, `row_start..row_end`); `src_nx = col_end − col_start`
+/// is the read window's row stride. The pushed value is the largest data sample
+/// in the box (strongest echo), or `None` if the box is empty/off-window — so
+/// the result is always an actual source sample (discrete, PNG8-friendly) and
+/// never a blend. With `half_c == half_r == 0` the box is the single pixel
+/// `floor(col_f), floor(row_f)`, i.e. exactly nearest-neighbour.
+#[allow(clippy::too_many_arguments)] // window bounds + stride are all genuine inputs
+fn push_footprint_max(
+    values: &mut Vec<Option<f64>>,
+    pixels: &[Option<f64>],
+    col_f: f64,
+    row_f: f64,
+    half_c: f64,
+    half_r: f64,
+    col_start: u32,
+    col_end: u32,
+    row_start: u32,
+    row_end: u32,
+    src_nx: usize,
+) {
+    // Source-pixel box, floored to integers and clamped to the read window.
+    // (`col_end`/`row_end` are exclusive, so the inclusive upper bound is −1.)
+    let c0 = (col_f - half_c).floor().max(col_start as f64) as i64;
+    let c1 = (col_f + half_c).floor().min((col_end - 1) as f64) as i64;
+    let r0 = (row_f - half_r).floor().max(row_start as f64) as i64;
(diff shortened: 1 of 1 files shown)


mrauhala:
Addressed all three review findings in f1abed3:
- **0-dim tile**: `width.saturating_sub(1)` so a 0×N tile is a no-op again (was a debug-mode underflow panic).
- **Edge reverts to nearest**: the last column/row now uses a **backward** difference for the derivative (same magnitude on a smooth grid) instead of clamping the forward neighbour back to itself — so edge pixels keep a real footprint instead of silently dropping to nearest at the downsample scale the fix targets.
- Fixed the wrong arithmetic comment in `footprint_half_grows_with_downsample_factor`.

Full `engine-geotiff` suite still green (86 unit + integration).

PR #459, opened 2026-06-20
fix(engine-geotiff): edge-tile decode stride shear — venetian-blind / displaced data at east tile column (#458)

### LINKED ISSUE #458: GeoTIFF: edge-tile decode stride shear — venetian-blind stripes / displaced data at the east tile column (overviews)
## Symptom

At zoom-out, projected GeoTIFF radar composites (FMI EPSG:3067, MET Norway LCC) show a **venetian blind** of regular stripes in a block at the data's **eastern edge**, with the true eastern data displaced/missing. FMI: horizontal stripes east of ~33°E. MET Norway: a large dropped/sheared block in northern Finland. Zoom-**in** is clean; **GeoServer/GDAL render the same files cleanly**.

## Root cause

The local `tiff`-crate decode (`read_chunk` / `chunk_data_dimensions`, tiff 0.11.3) returns the **rightmost tile column clipped of its padding** — its row stride is `min(tile_width, width − col0)` (e.g. 216), not `tile_width` (256). But `copy_tile_to_result` (and the single-pixel `read_pixel` EDR reader) indexed the buffer with `local_row * tile_width + local_col` → a **progressive horizontal shear of `tile_width − clipped_width` px per row** in the rightmost tile column. That scatters a contiguous radar arm into periodic gaps (the venetian blind) and places decoded data at the wrong columns (the displaced/phantom block + the dropped block).

Why it's edge/east only and overview-only:
- The shear is confined to the **last tile column** (a vertical line in source pixels → the sharp east geographic boundary). Interior tile columns are full-width → unaffected.
- It only bites when **real data lands in that clipped edge tile**, which happens for the selected **overview** at zoom-out; at full resolution the same data sits in interior tiles, so full-res renders clean (the bug is latent there).
- GDAL/GeoServer decode edge-tile strides correctly → clean.

Independently confirmed by byte-level comparison (engine returns data where the stored overview is nodata at the same indices) and shown **compression-independent** (re-encoding the COG to PREDICTOR=1 and COMPRESS=NONE reproduces byte-identical stripes). The earlier ProjectionGrid/resampling hypotheses were refuted (coarse-grid `col_f`/`row_f` match exact per-pixel projection to ≤0.05 px). Pre-dates and is independent of #457 (footprint-MAX), which only *amplifies* the visible artefact.

## Fix

In `crates/engine-geotiff/src/reader.rs`: a `local_tile_data_width(metadata, tile_col)` helper returns the true clipped stride; thread it into the **local** decode paths (overview read, full-res read, single-pixel EDR `read_pixel`) and keep `tile_width` for the **remote/HTTP** paths (which decode the full padded raw tile — confirmed by code + TIFF spec: tiled edge tiles are stored padded to full tile size on disk). `copy_tile_to_result` takes the stride as a parameter. Interior tiles byte-identical; discrete values preserved (PNG8-safe — pure indexing fix, no resampling/smoothing change).

Fixes both manifestations (FMI east phantom + Norway dropped block) — single root cause. EDR position queries on the last tile column also corrected (now match `gdallocationinfo`).

mrauhala:
Closes #458.

## Symptom

At zoom-out, projected GeoTIFF radar composites (FMI EPSG:3067, MET Norway LCC) showed a **venetian blind** of regular stripes in a block at the data's **eastern edge**, with the true eastern data displaced/missing (FMI: horizontal stripes east of ~33°E; Norway: a large dropped/sheared block in northern Finland). Zoom-**in** was clean and **GeoServer/GDAL render the same files cleanly** — so the bug was ours.

## Root cause

The local `tiff`-crate decode (`read_chunk` / `chunk_data_dimensions`) returns the **rightmost tile column clipped of its padding** — its row stride is `min(tile_width, width − col0)` (e.g. 216), not `tile_width` (256). But `copy_tile_to_result` and the single-pixel `read_pixel` EDR reader indexed the buffer with `local_row * tile_width + local_col` → a **progressive horizontal shear of `tile_width − clipped_width` px per row** in the rightmost tile column.

Why edge/east-only and overview-only:
- The shear is confined to the **last tile column** — a vertical line in source pixels → the sharp east geographic boundary. Interior columns are full-width → untouched.
- It only bites when real data lands in that clipped edge tile, which happens for the selected **overview** at zoom-out; at full resolution that data sits in interior tiles, so full-res renders clean (the bug is latent there).
- GDAL/GeoServer decode edge-tile strides correctly → clean.

## Fix

`local_tile_data_width(metadata, tile_col)` returns the true clipped stride; threaded into the three **local** decode paths (overview read, full-res read, single-pixel EDR reader) and into `copy_tile_to_result`. The **remote/HTTP** paths keep `tile_width` — they decode the *full padded raw tile* (confirmed by code + TIFF spec: tiled edge tiles are stored padded to full tile size on disk, so the OPERA S3 COG is safe). Interior tiles byte-identical; values are actual source samples → **PNG8-safe** (pure indexing fix, no resampling/smoothing change).

Fixes **both** manifestations (FMI east phantom + Norway dropped block) from one root cause; EDR position queries on the last tile column now match `gdallocationinfo`.

## Verification

GDAL-parity harness (both render paths, both projections, multiple zooms):
- FMI wide: phantom-shear **6453 → 99**, eastern displaced block gone.
- Norway northern-Finland: dropped block **41002 → 681** (98.3%), IoU **0.702 → 0.991**.
- No newly-dropped data anywhere; zoom-in views byte-identical; PNG8 preserved every view.
- EDR `read_pixel` matches `gdallocationinfo` exactly (incl. a point in the former bug region).
- `cargo test -p engine-geotiff -p ds-core`: green (273 + 89 + integration). Clippy clean.

## Tests

Two unit tests pin the indexing math against a synthetic clipped buffer; one decodes a **real committed fixture's clipped edge tile** (`local_tiff_decode_returns_clipped_rightmost_tile`) so a future `tiff`-crate bump that padded edge tiles would fail loudly.

## Notes

- Independent of #457 (footprint-MAX), which only *amplified* the visible artefact. Earlier ProjectionGrid/resampling hypotheses were **refuted** (coarse-grid col/row match exact per-pixel projection to ≤0.05 px); the cause is **compression-independent** (PREDICTOR=1 / COMPRESS=NONE reproduce byte-identical stripes).
- A separate PR will roll back #457 (its dilation is the "echo field changes between zoom levels" behaviour; combined with this fix it recovers full GDAL-nearest parity).

### DIFF
--- a/crates/engine-geotiff/src/reader.rs
+++ b/crates/engine-geotiff/src/reader.rs
@@ -2060,8 +2060,13 @@ fn read_bbox_parallel_http(
 /// interior columns are exactly `tile_width`. (The remote/HTTP path decodes the
 /// full padded raw tile and uses `tile_width` instead.)
 fn local_tile_data_width(metadata: &TiffMetadata, tile_col: u32) -> usize {
-    let col0 = tile_col * metadata.tile_width;
-    metadata.width.saturating_sub(col0).min(metadata.tile_width) as usize
+    // Compute in u64 so a pathological `tile_col * tile_width` can't overflow
+    // u32 and wrap to a too-large stride (which would silently restore the
+    // pre-fix shear). For any real TIFF the product is bounded by `width`.
+    let col0 = tile_col as u64 * metadata.tile_width as u64;
+    (metadata.width as u64)
+        .saturating_sub(col0)
+        .min(metadata.tile_width as u64) as usize
 }
 
 /// Copy pixels from a decoded tile into the output result grid.
--- a/crates/engine-geotiff/src/reader.rs
+++ b/crates/engine-geotiff/src/reader.rs
@@ -1305,47 +1305,61 @@ pub fn read_pixel(
 
     let local_col = col % metadata.tile_width;
     let local_row = row % metadata.tile_height;
-    let local_idx = (local_row * metadata.tile_width + local_col) as usize;
 
-    let values = match source {
+    // The row stride of the decoded tile depends on the source: the local
+    // `tiff`-crate decode clips the rightmost tile column of its padding (stride
+    // = clipped width), while the remote/HTTP path decodes the full padded raw
+    // tile (stride = tile_width). Indexing with the wrong stride reads a sheared
+    // pixel in the last tile column (#458).
+    let (values, tile_data_width) = match source {
         DataSource::Remote {
             store,
             path,
             tile_info,
-        } => read_remote_chunk_f64(
-            store,
-            path,
-            tile_info,
-            metadata,
-            chunk_index,
-            cache,
-            file_path,
-            band_index,
-            0,    // full resolution IFD
-            None, // single-pixel read: caller is already on a runtime worker
-        )?,
+        } => (
+            read_remote_chunk_f64(
+                store,
+                path,
+                tile_info,
+                metadata,
+                chunk_index,
+                cache,
+                file_path,
+                band_index,
+                0,    // full resolution IFD
+                None, // single-pixel read: caller is already on a runtime worker
+            )?,
+            metadata.tile_width as usize,
+        ),
         DataSource::HttpDirect {
             url,
             http,
             tile_info,
-        } => read_http_chunk_f64(
-            http,
-            url,
-            tile_info,
-            metadata,
-            chunk_index,
-            cache,
-            file_path,
-            band_index,
-            0,    // full resolution IFD
-            None, // single-pixel read: caller is already on a runtime worker
-        )?,
+        } => (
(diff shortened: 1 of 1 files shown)


mrauhala:
Good catch — applied in the latest commit. Widened the whole computation to u64 (rather than the literal `col0 as u32` form, which would still truncate if `col0 > u32::MAX`): `(width as u64).saturating_sub(tile_col as u64 * tile_width as u64).min(tile_width as u64) as usize`. Matches the existing `saturating_sub` defensive intent; no behaviour change for any real TIFF.

Issue #436, opened 2026-06-16
engine-postgis: surface the real Postgres error on metadata refresh (don't discard via map_err(|_| Db))


mrauhala:
## Problem

`fetch_station_rows` / `fetch_observation_rows` / `fetch_temporal_extent` map every `tokio_postgres` failure to a detail-less `MetadataError::Db` via `.map_err(|_| MetadataError::Db)`. The HTTP `/health` then shows only `Engine error: metadata refresh failed: database error`, and nothing is logged with the actual SQL error.

This directly cost diagnosis time during the #433/#435 incident: the real cause was `canceling statement due to statement timeout` (the role's `statement_timeout=5s`), but that string never reached the logs — it had to be reproduced by hand against the live DB as the read-only role.

## Proposal

Log the underlying `tokio_postgres::Error` (and ideally its SQLSTATE) at WARN before collapsing to the generic client-facing `MetadataError::Db`, mirroring how `engine.rs::map_pg_error` already does it for query-path failures (logs `sql`/`error`, returns a generic message). Keep the client-facing message generic (no internal detail leak — per CLAUDE.md), but make the operator-facing log actionable.

Specifically distinguish/log:
- statement timeout (SQLSTATE `57014`) — points at `statement_timeout` vs query cost,
- read-only violation, connection errors, etc.

## Scope

`crates/engine-postgis/src/metadata.rs` — the three `map_err(|_| MetadataError::Db)` sites. Add a `tracing::warn!` with the error before mapping. Pre-existing pattern flagged in the #434 review as well.

Context: surfaced during #433 / PR #434 / PR #435.

PR #478, opened 2026-07-06
docs: restructure CLAUDE.md into scoped per-crate files

mrauhala:
## Summary

Splits the 961-line root `CLAUDE.md` into:

- a **461-line root** holding only workspace-wide content, with a numbered **Critical Rules** section up front (imperative form — the rules whose violation has caused production incidents), and
- **12 per-crate `CLAUDE.md` files** (`crates/{server,api-wms,api-edr,api-3dtiles,ds-3dtiles,engine-odim,engine-geotiff,engine-grib,engine-querydata,engine-zarr,engine-cap,engine-postgis}/`) that agent tooling auto-loads only when working inside that crate.

## Why

- The old file was always-loaded context for every session regardless of task; engine-specific detail (e.g. the CesiumJS voxel gotchas) taxed every unrelated session.
- Load-bearing rules were buried mid-paragraph in long narrative bullets — weaker/older models in particular skim past imperatives that aren't stated first. Rules are now numbered, bolded, one per bullet, with the mechanism/incident as supporting detail underneath.
- Removed `[[...]]` wiki-links that pointed at one agent's private memory files — unresolvable for any other model, tool, or human.
- Folded in verified, previously memory-only gotchas so they survive in the repo: watcher event-kind filter (#424), GeoTIFF edge-tile decode stride (#458), overview-selection cliff, postgis `map_err` detail-swallowing (#436), the `RasterInfo` no-`#[non_exhaustive]` rule.

## Fact preservation

No technical content was dropped. Every section of the old file maps to the root or exactly one crate file; cross-cutting rules (web-mercator, ProjectionGrid, poll-runtime/ds-storage bridge, instances, vertical dimension) stay in the root.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,8 +1,90 @@
 # MeteoCore — Claude Instructions
 
-## What This Is
+MeteoCore is a Rust workspace implementing OGC API - EDR, OGC API - Features,
+OGC API - Maps, OGC API - Tiles, OGC WMS 1.3.0, and OGC 3D Tiles servers for
+weather data (radar, NWP models, observations, alerts).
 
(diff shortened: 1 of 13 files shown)


mrauhala:
Review addressed in the follow-up commit:

1. **Restored** the superuser-DSN / unconfigured-role bypass consequence to `engine-postgis/CLAUDE.md`'s session-limits bullet, and the `/health` failed-vs-degraded boot-status nuance to the live-ping bullet.
2. **Issue numbers verified against the tracker**: #436 (surface real PG error — real, open, fix now up as PR #484), #438/#441/#445 (merged postgis PRs), and #444 is genuinely the per-query-histograms follow-up — the old file referenced it generically as "the #110 follow-up"; the new text is more precise, not drifted.
3. **Restored** the zarr `"zarr" => &["edr","wms","maps","tiles"]` allowlist entry + `register_parameter_layer_styles` reference (new "APIs" section), and grib's multi-parameter-vs-GeoTIFF contrast.

PR #483, opened 2026-07-06
fix(api-edr): 500 instead of request-path panic on registry divergence (#479)

### LINKED ISSUE #479: api-edr: request-path panic if engines/collections registries diverge
`crates/api-edr/src/handlers.rs` (`lookup_engine`, ~line 134): after `state.engines.get(id)` succeeds, the handler does

```rust
let config = state.collections.get(id).unwrap();
```

This assumes the `engines` and `collections` maps always have identical keysets. A registration divergence in `server/src/admin.rs` (an engine wired without its config, or vice versa) would turn every request for that collection into a **panic** on the request path instead of a 500.

Fix: replace the `unwrap()` with `ok_or_else` mapping to a 500 (generic message, per CLAUDE.md error-leak rule). Found during the 2026-07-06 code-quality audit; verified against current main.

mrauhala:
## Summary

`lookup_collection` in `crates/api-edr/src/handlers.rs` did `state.collections.get(id).unwrap()` after the `engines` lookup succeeded — a keyset-parity assumption between two maps built in `admin.rs`. If they ever diverge (engine wired without its config), every request for that collection panics on the request path.

Now: `ok_or_else` → `tracing::error!` + generic 500 (per the CLAUDE.md no-internal-details rule).

Found and verified in the 2026-07-06 code-quality audit. The similar-looking `.expect("searchable params is a JSON array")` sites were checked and deliberately left — they act on a locally-built `json!([...])` literal and are infallible by construction.

## Testing

`cargo test -p api-edr` (all pass), fmt + clippy clean.

Closes #479

### DIFF
--- a/crates/api-edr/src/handlers.rs
+++ b/crates/api-edr/src/handlers.rs
@@ -131,7 +131,18 @@ fn lookup_collection<'a>(
             Json(json!({ "code": "NotFound", "description": format!("Collection '{id}' not found") })),
         )
     })?;
-    let config = state.collections.get(id).unwrap();
+    // engines and collections are built from the same config in admin.rs, but
+    // a registration divergence must surface as a 500, not a request panic.
+    let config = state.collections.get(id).ok_or_else(|| {
+        tracing::error!(
+            collection = id,
+            "engine registered without a matching collection config"
+        );
+        (
+            StatusCode::INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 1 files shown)


mrauhala:
Review addressed: the 500 now reuses `server_error()` — same `"code": "ServerError"` vocabulary as every other 500 in the file; the `tracing::error!` with the collection id stays. Tests re-run green.

PR #487, opened 2026-07-06
chore(claude): committed project hooks — branch/fmt commit guard + per-crate edit checks

mrauhala:
## Summary

First committed `.claude/settings.json` for the repo (project scope; `settings.local.json` remains personal/ignored). Turns three CLAUDE.md prose rules into machine-enforced hooks:

| Hook | Event | Behavior |
|---|---|---|
| `scripts/hooks/pre-commit-guard.sh` | PreToolUse (Bash) | Denies `git commit` on `main`; denies commit when `cargo fmt --check` fails — reason fed back to the model, saving the CI round-trip |
| `scripts/hooks/post-edit-checks.py` | PostToolUse (Edit\|Write) | Re-runs `scripts/check_sql_safety.sh` after any `engine-postgis` `.rs` edit (blocks on failure); injects the `cargo test -p api-edr` CoverageJSON reminder after `api-edr/src/response.rs` edits |

Non-matching tool calls exit silently (~30 ms python3 spawn worst case).

## Verification

Pipe-tested with the exact stdin JSON the harness sends, all 8 paths: non-commit pass-through, commit on feature branch (allow), commit on main (deny JSON), postgis edit on clean tree (silent), postgis edit with a planted `format!("SELECT …")` violation (block JSON, then cleaned up + tripwire re-verified clean), response.rs edit (additionalContext), plus `jq -e` schema validation of settings.json.

Note for reviewers: hooks activate on session start (or `/hooks` reload) — an already-running session won't pick them up mid-flight.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "sh scripts/hooks/pre-commit-guard.sh",
+            "command": "cd \"${CLAUDE_PROJECT_DIR:-.}\" && sh scripts/hooks/pre-commit-guard.sh",
             "timeout": 120,
             "statusMessage": "Pre-commit guard: branch + cargo fmt"
           }
@@ -19,7 +19,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "python3 scripts/hooks/post-edit-checks.py",
+            "command": "cd \"${CLAUDE_PROJECT_DIR:-.}\" && python3 scripts/hooks/post-edit-checks.py",
             "timeout": 60,
             "statusMessage": "Post-edit checks: SQL safety / CovJSON reminder"
           }
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,30 @@
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "sh scripts/hooks/pre-commit-guard.sh",
+            "timeout": 120,
+            "statusMessage": "Pre-commit guard: branch + cargo fmt"
+          }
+        ]
+      }
+    ],
+    "PostToolUse": [
+      {
+        "matcher": "Edit|Write",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3 scripts/hooks/post-edit-checks.py",
+            "timeout": 60,
+            "statusMessage": "Post-edit checks: SQL safety / CovJSON reminder"
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ __pycache__/
 # Docker builder; should never be committed (.dockerignore also excludes
 # it so a stray local copy doesn't override the planner's fresh output).
 recipe.json
+
+# Personal Claude Code overrides — .claude/settings.json is committed
+# (team hooks), but the .local variant is per-machine and must never be.
+.claude/settings.local.json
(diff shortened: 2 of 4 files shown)


mrauhala:
All three findings addressed (branch also rebased onto main):

1. **Substring match → command-position parse.** Detection now requires `git` at a command position (start, or after `;` `&` `|` `(` backtick/newline), followed only by flag-like tokens, then an optionally-quoted `commit`. Re-verified with a proper test matrix — and the review caught more than it knew: my original "deny on main" pipe-test had silently run on the feature branch (the `git checkout main` failed on uncommitted hook edits, stderr swallowed), so the deny path had never truly been exercised. Now verified for real: `git add y && git commit` on main → deny JSON; `git log --grep="git commit"` on main → silent; `git  commit` / `git "commit"` / `-c` flag interludes → detected. Variable-built commands can still evade — noted in the script as guardrail-not-security-boundary.
2. **Rule citations now resolve**: #478 merged the numbered Critical Rules list into CLAUDE.md on main (Rule 1 = never commit to main, Rule 2 = fmt/clippy before commit, Rule 12 = CovJSON tests), so the hooks now enforce documented policy. This PR's branch is rebased on that main.
3. **`$CLAUDE_PROJECT_DIR` anchoring**: both settings.json commands now `cd "${CLAUDE_PROJECT_DIR:-.}"` first, and `post-edit-checks.py` treats `check_sql_safety.sh` exit 2 (missing target dir = config problem) as a non-blocking warning instead of a spurious block — only exit 1 (real violation) blocks.

mrauhala:
Round-2 findings fixed and pushed:

- **Arg-taking flags + env prefixes**: the pre-`commit` group now consumes one optional non-flag argument per flag (`git -C /tmp commit`, `git -c user.email=x commit` → detected) and `NAME=value` prefixes before `git` (`GIT_AUTHOR_NAME=x git commit` → detected).
- **Heredoc false positive**: heredoc bodies (`<<EOF`, `<<-EOF`, quoted delimiters) are stripped before matching, so a data line reading `git commit` (writing an example script) stays silent, while a real commit *after* a heredoc still detects.

Verified with an 11-case matrix run against a scratch copy on actual `main` (deny: plain / `&&` / `-C dir` / `-c k=v` / env-prefix / newline-separated / post-heredoc; silent: `git log --grep="git commit"` / grep-for-string / heredoc data with unquoted and quoted delimiters). Two test-harness gotchas were themselves caught and fixed in the process: a failed `git checkout main` had made earlier deny-rows vacuous, and zsh's `echo` was corrupting `\n` in the JSON payloads (now `printf`).

At this point the detector handles every non-adversarial invocation form named across both rounds; anything beyond (variable-built commands, `sh -c` strings) is the disclaimed guardrail boundary — I'd treat further shell-parsing fidelity as out of scope for a hook.

mrauhala:
Round-3 bug fixed and pushed:

- **`commit\b` → `commit\b(?!-)`**: `git commit-tree` and `git commit-graph` no longer misdetect (verified silent on main); the `(?!-)` form was chosen over `(?:\s|$)` because the latter would have broken quoted-subcommand detection (`git "commit"` — the closing quote is neither whitespace nor end).
- **Bonus from the non-blocking note**: `command git commit` and `\git commit` (the alias-bypass idiom) are now detected — cheap to support. The caveat comment now explicitly names the remaining undetected-by-design forms: interpreter-wrapped commits (`sh -c "git commit …"` — the string is data to the outer shell), absolute-path `​/usr/bin/git`, and variable-built commands.

Nine-case matrix re-run against a scratch copy on actual `main` (branch asserted before running this time): commit-tree/commit-graph/`--grep`/heredoc-data silent; plain, quoted, `-C dir`, `command git`, `\git` denied.

Issue #504, opened 2026-07-11
Lightning WMS/Maps/Tiles raster layer (age-colored symbol splat)


mrauhala:
Optional later phase of the #113 events design — server-rendered lightning for clients that consume WMS/Maps/Tiles rather than EDR/Features.

Render event points as an age-colored raster layer (classic lightning display: color ramp by strike age within the TIME window) via a `MapEngine` impl on the events shape, following the CAP engine's vector→raster precedent: project per-vertex through `ds_core::geo::geometry_to_pixels` / `OutputCrs::world_to_fraction` (never per output pixel), splat through ds-render.

Open questions:
- Symbol size vs zoom (fixed-pixel splat is cheap; icons need more thought).
- TIME dimension semantics: window end = TIME, window length = style or dimension?
- Whether MVT vector tiles (ds-mvt) serve the need better than raster for modern map clients.

Not needed for the EDR-client lightning restoration (#501/#502/#503 cover that).

Issue #503, opened 2026-07-11
engine-postgis: FeatureEngine for events shape (GeoJSON items, keyset pagination)


mrauhala:
Implements the #113 design (sub-issue 3 of 3). Depends on #502 (events shape).

Surface `events` collections via `FeatureEngine` so clients can fetch strikes as GeoJSON for symbol rendering (and future MVT):

- `/features/collections/{id}/items?bbox=…&datetime=…` → one Feature per event: Point geometry, properties = parameter columns + event time (`time` as RFC 3339). `PropertyValue` enum, not `serde_json::Value`.
- Keyset pagination on `(time, id)` (matches the PK): `limit` + `next` link encoding the last `(time, id)` seen. No OFFSET scans.
- Same `default_datetime` window semantics as the EDR side when `datetime` is absent.
- Collection wired to the Features router only when `"features"` is in its `apis` array (enforced config rule).

Acceptance: end-to-end curl smoke test on the example lightning collection from #502; `cargo test -p engine-postgis`.

mrauhala:
Heads-up from PR #506 round-3 review: config validation now rejects any `apis` entry other than `"edr"` on an events-shape collection (ds-core `validate()`). When this issue implements the FeatureEngine surface for events, lift that restriction (allow "features") and add the corresponding acceptance test.

Issue #501, opened 2026-07-11
api-edr + ds-core: CoverageJSON Point domain type


mrauhala:
Prerequisite for the `events` shape EDR output (#113 design, sub-issue 1 of 3).

Add the `Point` CoverageJSON domain type so scattered events (lightning strikes) can be returned as a `CoverageCollection` of Point coverages — one per event, each with a single-value `t` axis.

Per the recipe in `crates/api-edr/CLAUDE.md`:

1. Add a `Point` variant to `DomainDescription` in `ds-core/src/model.rs` (x/y single-value axes, optional single-value `t`).
2. Add the match arm in `build_domain()` in `api-edr/src/response.rs`.
3. The schema's `domainBase.dependencies.domainType` already carries the `Point` constraint block — follow it exactly.
4. Add a validation test in `tests/covjson_validation.rs`, including a CoverageCollection of Point coverages (the actual events response shape).

Remember: `RasterInfo`/domain structs are cross-crate literals — no `#[non_exhaustive]`; run `cargo test -p api-edr` (CoverageJSON schema validation) and `cargo check --workspace --tests`.

Issue #498, opened 2026-07-11
engine-postgis: raise the 500-station area cap once #115 lands (all-country polygons 400 today)


mrauhala:
## Problem

A client integration needs EDR area queries over all of Finland against `fmi-obs` (postgis). The area path caps the stations-in-polygon prefilter at `MAX_STATIONS_IN_POLYGON = 501` (`crates/engine-postgis/src/query.rs`), and live `fmi-obs` advertises **8,276 stations** — so any country-sized polygon breaches the cap. Since PR #497 the breach returns a descriptive `400 "Query too large: area query matched more than the 500 station cap — narrow the polygon"` (previously an opaque 500), but the client still has to split large polygons into ≤500-station pieces client-side.

## Why the cap can't just be raised today

`query_area` fans out **one sequential observation query per station** (each its own pool checkout). Measured on production 2026-07-11: a 126-station area query = 0.31 s end-to-end, ~2.4 ms marginal per station. Extrapolated: ~1,000 stations ≈ 2.5 s per request plus 1,000 serial pool round-trips — unacceptable as a request-path behavior. This is exactly the N+1 tracked in #115 (see production numbers posted there).

## Proposal

Gated on #115 (bounded concurrent fan-out, ~8–16 wide, which puts a 1,000-station query in the few-hundred-ms range):

1. Raise the default cap from 500 to a value sized from measured post-#115 latency (e.g. 2,000–5,000; benchmark against live `fmi-obs` before picking).
2. Consider making it configurable per collection (`[postgis].max_area_stations`) with the raised value as default and a hard upper bound as an invariant, so a dense global dataset can still be capped tighter than a national network.
3. Keep the `QueryTooLarge` → 400 behavior from PR #497 for whatever the effective cap is.
4. Update the row-cap documentation in `crates/engine-postgis/CLAUDE.md` (currently documents the caps as non-configurable invariants) and the crate README.

Note the interaction with the per-station `LIMIT 10_001` row cap: raising the station cap multiplies worst-case response size (5,000 stations × dense series can be very large CoverageJSON) — the response-size dimension needs a sanity check as part of the benchmark, not just latency.

## Acceptance

- All-Finland polygon (`POLYGON((19 59,32 59,32 71,19 71,19 59))`) on live `fmi-obs` returns 200 CoverageJSON in acceptable time instead of 400.
- Cap breach beyond the new limit still returns the descriptive 400.

PR #506, opened 2026-07-11
feat(engine-postgis): events shape — EDR area queries for non-station event tables

### LINKED ISSUE #502: engine-postgis: events shape — EDR area/radius for event tables (lightning)
Implements the #113 design (sub-issue 2 of 3). Depends on #501 (Point domain type).

Fourth schema shape `[postgis.events]` for non-station event data — each row is an independent event with its own time + point geometry. Reference dataset: the `lightning` hypertable (`(id, time, the_geom, multiplicity, peak_current, cloud_indicator, ellipse_major)`, PK `(time, id)`, covering index `btree(time DESC, the_geom) INCLUDE (...)`), reachable via the same DSN/pool as the existing obs collection.

## Scope

- Config block per the sketch on #113: `table`, `time_col`, `time_col_tz`, `geom_col`, `id_col`, `default_datetime` (ISO 8601 duration, applied when the request has no `datetime` — never full history), `extent_bbox`, `[[events.parameters]]` (column/unit/label).
- EDR `area` + `radius` only: single SQL query per request (no fan-out, no interaction with the obs fanout semaphore), `WHERE time >= $1 AND time < $2 AND ST_Intersects/ST_DWithin`, `ORDER BY time DESC, id`, `LIMIT` bound to the remaining response-value budget (#498 machinery, `rows × selected_params`; breach → `QueryTooLarge` 400).
- `position` and `locations` → 400 with a message pointing at `radius` / explaining the collection has no stations. `data_queries` advertises area/radius only. No instances.
- Response: `CoverageCollection` of `Point` coverages (one per event, own `t`), parameters from the config blocks.
- Metadata: temporal extent via index-only `min(time)`/`max(time)`; spatial extent from `extent_bbox` config, never `ST_Extent`. Reuse existing ping/refresh/health machinery.

## Gotchas (verified against the live table)

- `time` is `timestamp without time zone` → require `time_col_tz`, reuse the bind-side `AT TIME ZONE` pattern (keeps the time index usable).
- `ellipse_major` is `numeric(3,1)`, currently rejected by the pg_type allowlist → cast event parameter columns in the emitted SELECT list (`::float8`) instead of widening the coercion table.
- SQL safety: every identifier through `quote_ident`, every value a `$N` bind (Critical Rule 8; CI-enforced).

## Acceptance

- Runnable example collection config AND an end-to-end server + curl smoke test against real data (unit tests alone miss integration bugs).
- `cargo test -p engine-postgis` integration test with a fixture (tests/integration.rs pattern).
- Budget breach path returns 400 naming the numbers and which dimension to narrow.

mrauhala:
Closes #502. Implements the #113 events-shape design (prerequisite #501 merged as #505). Restores lightning data for the EDR client after the GeoServer retirement.

## What

**Config** (`observations.shape = "events"` — a fourth shape rather than the separate `[postgis.events]` block sketched on #113; it reuses the whole observations validation/lowering machinery, and the semantics are identical):

- New fields `id_col` (ORDER BY tiebreak), `default_datetime` (window when the query has no `datetime` — an unqualified query never scans full history; default PT1H), `extent_bbox` (advertised spatial extent — never `ST_Extent`, a full-scan trap on a 1.7M-row hypertable). Each rejected on station shapes.
- A `[postgis.stations]` block, `station_fk_col`/`param_col`/`value_col`, `columns`/`tables`, or `locations_window` on an events collection are config-load errors.
- Events `source_key`s are identifier-validated at load and at engine resolve — they name SQL columns directly, with no columns/tables mapping to launder them through.

**Engine**:

- `EventsShape` + `LocationSource::None`; the exhaustive matches surfaced every station-path assumption at compile time.
- EDR **area only** (`supported_query_types()` → `data_queries` advertises just area); `position`/`location` return 400 with "use the area query".
- One SQL statement per request: time-range binds wrapped bind-side (`$1 AT TIME ZONE` — index/chunk-pruning preserved), `&&` bbox prefilter + exact `ST_Intersects`, `ORDER BY time DESC, id`, `LIMIT` bound to the response-value budget (#498) with a +1 sentinel. No fan-out; the fan-out semaphore is not involved and one pooled connection per request coexists with station collections on the shared pool.
- Response: `CoverageCollection` of `Point` coverages (one per event, 0-d scalar ranges — #501/#505). An empty window is a valid **empty collection**, not a 404 — no strikes in the area is an answer.
- Parameter columns cast `::double precision` in the SELECT, so `smallint` and `numeric(p,s)` (the lightning table's `ellipse_major`) work with zero config.
- Metadata: temporal extent via index-only `MIN`/`MAX`; spatial from config; ping/refresh/health machinery reused untouched.

## Tests

- ds-core: 9 TOML round-trip tests (accept + every rejection path).
- engine-postgis: shape lowering, resolve (LocationSource::None, window default/parse, bad source_key), `build_events_area` SQL/bind/limit/values-per-row assertions, station-builder rejection, and pure `assemble_event_coverages` (Point domain, scalar NdArray, null-stays-null).
- `cargo test -p ds-core -p engine-postgis` and `cargo clippy --all-targets -- -D warnings` clean.

## End-to-end smoke test (live data)

Ran the built server against the production lightning hypertable (1.76M rows) over an SSH tunnel, storm evening in progress:

- Collection metadata: `data_queries: [area]`, live temporal extent, config bbox, 4 parameters.
- Whole-Finland 6 h window: **HTTP 200, 11,051 Point coverages, 1.7 s** (debug build over tunnel), values decode correctly incl. the `numeric` column.
- No `datetime` → default window applied (16.8k strikes / 24 h); `parameter-name=peak_current` filters ranges.
- `position` → 400 "use the area query"; full-year whole-extent → 400 naming the 500k value budget.

Follow-ups: #503 (FeatureEngine/GeoJSON for events), #504 (optional WMS layer). EDR `radius` is not implemented — the query type doesn't exist in `EdrEngine`/api-edr at all yet; noted as a possible later addition since `area` covers the client.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -2149,6 +2149,17 @@ impl ServerConfig {
                     ))
                 })?;
                 validate_postgis(id, postgis)?;
+                // The events shape serves EDR only for now — its
+                // FeatureEngine surface is #503. Without this check an
+                // `apis = ["features"]` events collection would load
+                // cleanly and silently serve an empty station inventory.
+                if postgis.observations.shape == "events" {
+                    if let Some(api) = collection.apis.iter().find(|a| a.as_str() != "edr") {
+                        return Err(crate::error::DataServerError::Config(format!(
+                            "Collection '{id}': apis entry '{api}' is not supported for observations.shape 'events' (only \"edr\" until #503 implements event features)"
+                        )));
+                    }
+                }
             } else if collection.postgis.is_some() {
                 return Err(crate::error::DataServerError::Config(format!(
                     "Collection '{id}': [collections.postgis] is set but engine_type is '{}'",
@@ -3703,6 +3714,21 @@ unit = "1"
             .to_string();
         assert!(
             err.contains("id_col is only valid for observations.shape 'events'"),
+            "got: {err}"
+        );
+    }
+
+    #[test]
+    fn postgis_events_rejects_non_edr_apis() {
+        let tmp = TempDir::new().unwrap();
+        let toml =
+            lightning_events_toml().replace("apis = [\"edr\"]", "apis = [\"edr\", \"features\"]");
+        let path = write_config(tmp.path(), "config.toml", &toml);
+        let err = ServerConfig::from_file(path.to_str().unwrap())
+            .unwrap_err()
+            .to_string();
+        assert!(
+            err.contains("'features' is not supported for observations.shape 'events'"),
             "got: {err}"
         );
     }
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1026,6 +1026,20 @@ pub struct PostgisObservationsConfig {
     /// `per_parameter` shape: one entry per parameter → table mapping.
     #[serde(default)]
     pub tables: Vec<PostgisObservationTable>,
+    /// `events` shape: unique/tiebreak column for deterministic ordering
+    /// (`ORDER BY time DESC, id`). Required for `events`, forbidden otherwise.
+    #[serde(default)]
+    pub id_col: Option<String>,
+    /// `events` shape: window applied when a query carries no `datetime`
+    /// (ISO 8601 duration, e.g. `"PT1H"`; that is also the default). Event
+    /// tables are never scanned over full history by an unqualified query.
+    #[serde(default)]
+    pub default_datetime: Option<String>,
+    /// `events` shape: advertised spatial extent `[west, south, east, north]`
+    /// (CRS84). Static config — computing `ST_Extent` over a large event
+    /// table on every metadata refresh is a full-scan trap.
+    #[serde(default)]
+    pub extent_bbox: Option<[f64; 4]>,
 }
 
 impl PostgisObservationsConfig {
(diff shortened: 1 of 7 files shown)


mrauhala:
Addressed all four review findings in the follow-up commit: events `time_col_tz` now goes through `validate_tz_name` at config load; `extent_bbox` promoted to required for the events shape (with a rejection test); default-window constant unified via `pub(crate)` so the engine fallback can't drift; budget message is shape-aware ("events × parameters"). Full ds-core + engine-postgis suites and scoped clippy clean.

mrauhala:
Round-2 findings addressed: `decode_event_rows` decodes lon/lat as `Option<f64>` and skips degenerate-geometry rows (plain NULL geoms already can't pass the WHERE clause — documented in a comment); added engine-level tests (lazy pool, no DB needed) for area-only `supported_query_types` and the `query_location`/`query_position` 400s. 128 engine tests green, scoped clippy clean.

mrauhala:
Round-3 finding addressed: events collections now reject any `apis` entry other than `"edr"` at config load (message points at #503, which lifts the restriction when event features land), plus a rejection test.

PR #508, opened 2026-07-11
fix(render): key raster caches on the RESOLVED timestep, not the requested time

### LINKED ISSUE #507: WMS meta-tile cache poisoning: TIME requested before ingestion caches previous timestep under the new frame's key
Observed in production 2026-07-11 evening under storm load, multiple times: animation frames where a meta-tile-aligned region (e.g. the right part of the viewport) shows the PREVIOUS 5-min timestep while the rest of the frame is correct — two consecutive frames visibly "share" that region. Self-clears after hours (LRU), but recurs for every fresh frame under sustained load. Verified live during the incident; pixel-diff 3.5 h later showed the artifact aged out.

## Root cause (confirmed by code inspection)

Requested-TIME vs resolved-timestep mismatch in the WMS meta-tile path:

1. `api-wms/src/handlers.rs:226` uses the raw requested `TIME` — it goes unchanged into BOTH the meta-tile pixel-cache key prefix (`handlers.rs:382`) and the render closure (`get_raster_tile`, `handlers.rs:401`).
2. The geotiff engine snaps to latest-not-after-T (`engine-geotiff/src/lib.rs:1433-1446`, `catalog.entries.range(..=t).next_back()`) and returns a bare `RasterTile` — the handler never learns which timestep actually rendered.
3. So a `GetMap TIME=T` served while T's file is not yet in the catalog (30 s poll granularity + ingest lag) renders T−5 pixels and caches them under **T's** key (`render/src/metatile.rs:148-158,395`).
4. `ds-cache` `ByteBoundedCache` has no TTL and no invalidation API; reload preserves the cache. A hit never re-renders — the poison persists until LRU eviction.
5. Each covering tile resolves independently (`metatile.rs:354-405`), so only the tiles touched by the too-early request go stale — producing the partial, tile-aligned artifact. The meta-tile loop also uses plain `get`/`insert` (no single-flight), so a slow queued T−5 render can overwrite a correct T tile (last-writer-wins).

## Why it surfaced tonight

The trigger needs a client requesting `TIME=T` at ≈wall-clock T, before catalog visibility. Storm load widens the absent-window (ingest lag + 30 s poll) and multi-second semaphore queueing lets a single request's tile loop straddle the catalog swap — which is what makes the staleness *partial* and eye-catching rather than whole-frame.

## Fix (recommended: option 1)

1. **Key on the RESOLVED timestep.** Add `MapEngine::resolve_time(&self, Option<DateTime<Utc>>) -> Option<DateTime<Utc>>` (default identity); geotiff implements it with the same `range(..=t)` selection and `get_raster_tile` reuses it. api-wms resolves once per request and uses the resolved instant in the cache keys and the render call. A too-early request for T then caches its T−5 pixels under T−5's key (correct — it IS T−5 data, shared with real T−5 requests) and the first post-ingest request for T is a clean miss. Self-healing, no invalidation machinery, benign with reload and DIM_REFERENCE_TIME (forecast engines: identity). Files: `ds-core` (trait default), `engine-geotiff`, `api-wms/handlers.rs`.
2. Optional stopgap: reject/snap `TIME` newer than `RasterInfo.times.last()` at the API layer (mirrors existing dimension validation, `handlers.rs:176-201`).

Rejected: catalog-swap invalidation (needs new cache API + engine→global-cache layering violation); TTL (doesn't fix the key mismatch, hurts the hot newest frame).

## Repro

Unit-level (isolates the poison, `render/src/metatile.rs` tests): one `TilePixelCache`, key prefix `time=Some(T)`; render viewport A with a closure returning "old" pixels; render overlapping viewport B with a closure returning "new"; assert the overlap still serves "old" from cache.

End-to-end: local geotiff collection with only T−5 present → `GetMap TIME=T` (EPSG:3857) → drop T's file → wait one poll → overlapping `GetMap TIME=T` shows mixed timesteps.

Also check: the exact-bbox rendered cache and Maps/Tiles direct path share the same requested-time keying assumption — audit while fixing.

mrauhala:
Closes #507 — the storm-night partial one-step-behind animation regions.

## Root cause recap

`GetMap TIME=T` served while T's file is not yet ingested renders T−1 pixels (geotiff snaps latest-not-after; ODIM snaps nearest) but caches them under **T's** key in the no-TTL meta-tile + rendered caches. Once T lands, later requests for frame T serve the poisoned tiles as hits for the covered region and render fresh T elsewhere — a meta-tile-aligned region one timestep behind, persisting until LRU eviction and recurring for every fresh frame under load.

## Fix

**`MapEngine::resolve_time(time, reference_time)`** (new, default identity): the exact timestep the engine will render — the timestamp that must key any cache of the output.

- **api-wms / api-maps / api-tiles** resolve once per request, before any cache key is built. Bonus: the whole request is pinned to one timestep, so a catalog swap mid-render can no longer mix timesteps within a single response (the "straddle" variant seen under semaphore queueing).
- **geotiff**: selection factored into `Catalog::select_timestamp`; `get_raster_tile` and `resolve_time` share it — they cannot drift.
- **ODIM COMP**: `resolve_time` shares the existing `select_entry` (nearest).
- **ODIM PVOL site views**: selection factored into `select_volume_entry`, shared the same way.
- Engines not touched (grib/querydata/zarr/cap) keep the identity default — behavior unchanged; forecast run selection already requires an exact `DIM_REFERENCE_TIME` match.

Net semantics: a too-early request for T caches its T−1 pixels under **T−1's** key (correct — that *is* T−1 data, legitimately shared with real T−1 requests), and the first post-ingest request for T is a clean cache miss. Self-healing; no invalidation machinery, no TTL, benign with reload-preserves-cache.

## Verification (three levels)

1. **Unit**: `Catalog::select_timestamp` selection rules (exact / not-yet-ingested / pre-range / None / empty).
2. **Integration** (`api-wms/tests/integration.rs`): a time-snapping mock engine replays the incident — too-early `TIME=T` renders T−1; an explicit `TIME=T−1` for the same viewport is a pure cache hit (zero fresh engine calls — shared under the resolved key); after "ingestion", the same `TIME=T` re-renders every tile fresh instead of serving the pre-ingest tiles.
3. **End-to-end** (real server, real GeoTIFF catalog, real poll loop): `TIME=20:20` with only the 20:15 file present → 16 meta-tile misses; explicit `TIME=20:15` → served entirely from the rendered cache (proving the cross-frame key sharing); drop the 20:20 file, wait for the poll, repeat `TIME=20:20` → **misses 16→32, hits 0** — every tile re-rendered fresh. Pre-fix this final step served 16 stale hits.

Full six-crate test sweep + workspace `clippy -D warnings` clean.

## Notes

- The plain get/insert (no single-flight) last-writer-wins race in the meta-tile loop becomes benign with resolved keys: concurrent renders of the same key now always contain the same timestep's pixels. Single-flight remains a possible perf (not correctness) follow-up.
- `resolve_time` is snapshot-only (O(log n)), no STAC on-demand fetch — an empty catalog falls back to the requested time; the render then errors and caches nothing.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -188,6 +188,12 @@ gh issue create --title "..." --label "bug,priority: high" --milestone "v0.2"
 6. Wire it into the appropriate registries per the collection's `apis`.
 7. Obey the Performance & Concurrency rules — especially: spawn the poll loop
    on the background runtime, and never project per output pixel.
+   **If the engine snaps a requested time to an available timestep**
+   (latest-not-after, nearest, …) instead of exact-matching, it MUST
+   override `MapEngine::resolve_time` with the SAME selection logic
+   `get_raster_tile` uses (share one helper so they cannot drift) — the API
+   layers key the no-TTL rendered/meta-tile caches on it. Skipping this
+   reintroduces the #507 cache poisoning (stale animation frames).
 8. Ship a runnable (enabled) example collection config AND do an end-to-end
    server + curl smoke test against real data. Unit tests alone miss
    integration and unit-conversion bugs.
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -1010,6 +1010,11 @@ async fn render_map(
     // engines that materialise `RasterInfo` lazily.
     let raster_info = engine.raster_info();
     let time = validated.time.or_else(|| raster_info.times.last().copied());
+    // #507: snap to the exact timestep the engine will render before the
+    // cache key is built — a not-yet-ingested datetime must cache the
+    // previous timestep's pixels under the PREVIOUS timestep's key.
+    // Maps passes reference_time: None (latest run) throughout.
+    let time = engine.resolve_time(time, None);
 
     // Parameter selection precedence: ?parameter-name= wins over style.parameter.
     // Validate against the engine's advertised list when the query supplied one
--- a/crates/api-tiles/src/handlers.rs
+++ b/crates/api-tiles/src/handlers.rs
@@ -1385,6 +1385,11 @@ async fn render_tile(
     // the redundant call.
     let raster_info = engine.raster_info();
     let time = validated.time.or_else(|| raster_info.times.last().copied());
+    // #507: snap to the exact timestep the engine will render before the
+    // cache key is built — a not-yet-ingested datetime must cache the
+    // previous timestep's pixels under the PREVIOUS timestep's key.
+    // Tiles passes reference_time: None (latest run) throughout.
+    let time = engine.resolve_time(time, None);
 
     let tile_size = params::TILE_SIZE;
 
--- a/crates/api-wms/src/handlers.rs
+++ b/crates/api-wms/src/handlers.rs
@@ -225,6 +225,18 @@ pub async fn wms_handler(
             // the request falls through as `None` = the engine's own latest.
             let time = params.time.or_else(|| info.times.last().copied());
 
+            // #507: snap that instant to the exact timestep the engine will
+            // actually render, BEFORE any cache key is built. Engines that
+            // select latest-not-after (geotiff) would otherwise render T−1
+            // pixels for a not-yet-ingested TIME=T and cache them under T's
+            // key — permanently poisoning frame T for the covered tiles once
+            // T's file lands (the partial one-step-behind animation regions
+            // seen under storm load). Resolving here also pins one timestep
+            // for the whole request, so a catalog swap mid-render can no
+            // longer mix timesteps within a single response. Exact-match
+            // engines keep the identity default.
+            let time = engine.resolve_time(time, reference_time);
+
             // Build cache key
             let cache_key = CacheKey {
                 layer: params.layer.clone(),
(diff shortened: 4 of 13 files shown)


mrauhala:
Round-1 findings addressed: `resolve_time` overrides added for **grib** (via `resolve_run` + `find_step_for_time`, exact-then-nearest, last step on no-TIME), **querydata** (via `select_data` + `find_time_idx`) and **zarr** (via `nearest_time_idx`) — each sharing the exact selection its render path uses. GRIB's inherent `resolve_time` renamed `resolve_step` to remove the trait-method shadow. ODIM PVOL CELLS (`select_entry_and_quantity`) now calls the shared `select_volume_entry` instead of its inline copy. Root CLAUDE.md engine checklist documents the override contract. Four-engine test sweep + workspace clippy clean.

mrauhala:
Round-2 finding addressed: `resolve_step` now returns a `ResolvedStep { valid_time, file }` and the trait `resolve_time` calls it directly instead of re-deriving run+step selection — GRIB now has exactly one selection implementation like the other engines. engine-grib tests + scoped clippy clean.

Issue #113, opened 2026-04-20
engine-postgis: design 'events' shape for non-station event data (lightning, etc.)


mrauhala:
Follow-up to #99. Deferred past v1.

## Problem

Some observation data is not tied to fixed stations. Example: the \`lightning\` hypertable on nexus.meteo.fi — columns \`(id, time, the_geom, multiplicity, peak_current, cloud_indicator, ellipse_major)\`, no \`station_fk\`. Each row is an independent event with its own geometry.

The v1 DSL's three shapes (\`long\` / \`wide\` / \`per_parameter\`) all assume station-keyed observations. None fit event data.

## Proposal: fourth shape \`events\`

- Surfaces only via \`FeatureEngine\` (each row = one Feature with its own geom + properties)
- No station table required
- Config block names which columns become feature properties, pg_type → PropertyValue coercion reuses #107
- \`get_features(bbox, datetime, limit, offset)\` with keyset pagination on \`(time, id)\`
- No \`Engine\` trait impl for \`events\` shape — events are features, not coverages

## Open questions

- Do we need `[[events.parameters]]` metadata blocks, or does an event table advertise itself via column names?
- How does \`/collections\` describe an events collection (it has no parameters list in the EDR sense)?
- Does EDR try to route to it? If so, 400 with a message pointing at Features API?
- Interaction with `observations-as-time-aware-features` — that's a related but distinct problem (station history exploded into features).

## Not in scope for this issue

Implementation. This issue is design-only — produce a short addendum to the plan doc, then open implementation sub-issues.

mrauhala:
## Design: `events` shape (lightning)

Verified against the live `lightning` hypertable (2026-07-11): 1.76M rows total, ~38k strikes on a storm day, PK `(time, id)`, gist on `the_geom`, and a covering index `btree(time DESC, the_geom) INCLUDE (peak_current, cloud_indicator, ellipse_major, multiplicity)` — exactly the time+bbox access path this design needs. The table lives in the same `weather` DB the `fmi-obs` collection already uses, so no new DSN/pool/deployment surface.

**Operational driver:** the GeoServer that served the lightning WMS layer is being retired; the EDR-based client needs lightning from MeteoCore.

### Revision to the original proposal: EDR is in scope

[...]

PR #509, opened 2026-07-12
feat(engine-postgis): age-colored lightning WMS/Maps/Tiles layer for the events shape

### LINKED ISSUE #504: Lightning WMS/Maps/Tiles raster layer (age-colored symbol splat)

mrauhala:
Closes #504 — the final piece of the #113 events design (with #503 features remaining). Server-rendered lightning for WMS/Maps/Tiles clients, replacing the retired GeoServer layer.

## Design (resolving #504's open questions)

- **TIME semantics**: TIME = window END, floored to a 1-minute quantum; window length = the collection's existing `default_datetime` (PT1H for the prod config). The dimension advertises the most recent 6 h at 1-min cadence (≤360 capabilities entries, rebuilt per metadata refresh so `raster_info()` stays O(1)); older frames render via explicit TIME.
- **Symbol size vs zoom**: fixed 2 px-radius discs — classic marker-style display, constant across zoom. Pixel VALUE = strike age in minutes; the new `lightning_age` builtin colormap ramps fresh strikes near-white/yellow → orange → red → dark violet at the window edge. Newest strikes paint last and win overlaps (input sorted ascending).
- **MVT vs raster**: raster (this PR); MVT stays possible after #503.

## Architecture

- **Vector→raster per the CAP precedent**: one `world_to_fraction` per strike (per-vertex, never per output pixel — Critical Rule 5).
- **`spawn_blocking` DB bridge**: `get_raster_tile` runs inside `spawn_blocking` where the engine's `block_in_place` bridge panics (Critical Rule 7), so the window fetch drives the pool via `Handle::block_on` — the ODIM PVOL pixel-fetch pattern, documented in the module header.
- **One DB fetch per frame, not per tile**: the meta-tile loop calls `get_raster_tile` up to ~200×/viewport; a ds-cache byte-bounded single-flight LRU (`MC_LIGHTNING_STRIKE_CACHE_MB`, default 32, `lightning_strike_cache_*` metrics + Grafana panel targets) holds decoded whole-extent strike windows keyed `(collection, start, end)`. Strikes are immutable ⇒ settled windows never go stale; the live frame's staleness is bounded by the 1-min quantum. Truncation cap 200k strikes/window keeps the NEWEST (fetch is `ORDER BY time DESC`, reversed in memory for paint order).
- **#507 compliance**: `resolve_time` shares the exact window-end resolution the render uses (quantize / latest advertised step).
- **Wiring**: events collections may now list `wms`/`maps`/`tiles` (`features` still gated on #503); station shapes are explicitly rejected for `wms`/`maps` at config load and keep the MVT feature-tile path. Registration in the postgis admin arm mirrors the raster engines, including per-collection styles.

## Verification

- Unit: quantization, `build_wms_times` bounds/ordering/clamping, splat semantics (disc shape, age values, newest-wins, out-of-view skip), window SQL builder shape (bind order, DESC + LIMIT bind), config validation both directions (events+maps allowed, station+wms rejected).
- **Live smoke test** against the production lightning table over an SSH tunnel: GetCapabilities advertises the layer + 1-min TIME dimension; a storm-peak frame (2026-07-11T21:55Z) renders the three active cells with the age gradient exactly as intended; WMS + Maps + Tiles all 200; strike cache showed **2 misses / 30 hits** across two frames of meta-tile renders — the one-fetch-per-frame design working.
- 8 crate test suites + workspace `clippy -D warnings` clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -316,7 +316,7 @@ they were found. Critical Rules 5–7, 9 and 10 above are part of this set.
 | ODIM PVOL | `EdrEngine` + `MapEngine` + `VolumeEngine` (per-site views) + `FeatureEngine` (network engine) | EDR (position, locations, area, trajectory), WMS, Maps, Tiles, 3D Tiles, Features (site inventory) |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
 | Zarr | `EdrEngine` + `MapEngine` | EDR (position), WMS, Maps, Tiles; local + S3/HTTP |
-| PostGIS | `EdrEngine` + `FeatureEngine` | EDR (position, locations, area), Features |
+| PostGIS | `EdrEngine` + `FeatureEngine` + `MapEngine` (events shape only) | EDR (position, locations, area), Features; events shape: EDR (area) + WMS/Maps/Tiles (age-colored strike layer) |
 
 ## Config Format
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1221,6 +1221,7 @@ dependencies = [
  "arc-swap",
  "chrono",
  "deadpool-postgres",
+ "ds-cache",
  "ds-core",
  "futures",
  "postgis",
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -2149,16 +2149,33 @@ impl ServerConfig {
                     ))
                 })?;
                 validate_postgis(id, postgis)?;
-                // The events shape serves EDR only for now — its
-                // FeatureEngine surface is #503. Without this check an
-                // `apis = ["features"]` events collection would load
-                // cleanly and silently serve an empty station inventory.
+                // The events shape serves EDR + the raster map APIs (#504
+                // age-colored strike layer). Its FeatureEngine surface is
+                // #503 — without this check an `apis = ["features"]` events
+                // collection would load cleanly and silently serve an empty
+                // station inventory. Station shapes conversely have no
+                // raster rendering: wms/maps stay rejected for them via the
+                // engine_type allowlist gate in server admin wiring.
                 if postgis.observations.shape == "events" {
-                    if let Some(api) = collection.apis.iter().find(|a| a.as_str() != "edr") {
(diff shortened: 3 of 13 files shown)


mrauhala:
Round-1 findings addressed: `splat_strikes` pre-filters on lon/lat (two comparisons, 4× symbol margin) before projecting — the tile loop no longer costs tiles × strikes projections; a new test pins that edge strikes within the margin still paint their partial disc. The window cache key now folds in the qualified table + pool identity so a reload that repoints a collection id at a different table/DSN misses instead of serving stale windows. Maps tests + scoped clippy clean.

mrauhala:
Round-2 findings addressed: the lon/lat pre-filter is now scoped to Wgs84/WebMercator outputs (projected CRSs skip it — they never take the meta-tiled path, so the direct render's unfiltered scan is affordable and no fixed margin multiple is provably safe there); the module doc + README now state the accurate staleness bounds (explicit TIME = 1-min quantum; TIME-omitted = latest advertised step on the refresh cadence, since the API layers default to `times.last()` before `resolve_time` runs) with `metadata_refresh_secs = 60` in the example and explicit-TIME guidance for live clients. Maps tests + scoped clippy clean.

mrauhala:
Round-3 nit fixed: doc/README now state 361 inclusive steps for a full 6 h window, matching `build_wms_times` and its test. No code changes — this should be the converging pass.

Issue #522, opened 2026-07-20
engine-nowcast phase 1: derived-collection engine + WMS/Maps/Tiles serving


mrauhala:
Part of #519. Blocked by #520 (skill gate) and #521 (cache staleness).

- New `engine_type = "nowcast"` with `[collections.nowcast]` config (`source`, `horizon`, `step`, `history_frames`); config validation (unknown source, nowcast-of-nowcast rejected).
- Two-pass engine construction in `server/src/admin.rs`: base engines first, derived engines get an `Arc<dyn MapEngine>` clone of their source; `engine_type → supported_apis` allowlist entry; reload keeps working atomically.
- Poll loop on `poll_runtime()` watching source `raster_info().times`; generation = fetch last N native-grid frames from source → motion (from #520 core) → advect → store `RasterValues::U8` frames in an `ArcSwap`ped `BTreeMap` keyed by reference time.
- `MapEngine` impl serving stored frames through `OutputCrs`/`ProjectionGrid`; **`resolve_time` override sharing the same snap helper as `get_raster_tile`** (#507 contract, engine checklist item 7).
- Health: degraded until first generation; metrics: `nowcast_generation_seconds`, frame-count gauge, source-lag gauge (+ Grafana row or follow-up issue per repo rule).
- Runnable example collection config + end-to-end server + curl smoke test (WMS GetMap with future TIME, /preview animation across now) per engine checklist item 8.

Issue #519, opened 2026-07-20
epic: radar nowcasting — 0–2 h extrapolated composite layers (engine-nowcast)


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep as an epic; prioritize remaining children.

**Current assessment and challenge:** Nowcast raster/instances/cell intelligence shipped; numerical reflectivity EDR remains #523. Do not rank the original entire engine feature again.

The original roadmap below is historical context; shipped phases should not be counted as new work or as independent priority candidates. Remaining child issues retain their individual implementation and verification scope.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

## Why

"Will it rain on me in the next 30 minutes" is the highest-value consumer question, and almost no off-the-shelf OGC server answers it. MeteoCore is unusually close: cell tracking already computes per-cell motion (`ds_core::cells` `motion_ms` — its own doc comment calls it "the seed for a future motion/optical-flow field"), and the whole render pipeline treats TIME values in the future like any other timestep. An EDR position query on a nowcast collection **is** a minutecast API.

## What

A derived-collection engine `engine-nowcast` (`engine_type = "nowcast"`) that wraps an existing raster collection:

```toml
[[collections]]
id = "radar-nowcast-fi"
engine_type = "nowcast"
apis = ["edr", "wms", "maps", "tiles"]

[collections.nowcast]
source = "fmi-radar-composite"   # collection id of any raster engine
horizon = "PT2H"
step = "PT5M"                    # default: source cadence
history_frames = 3
```

- Wiring: second pass in `server/src/admin.rs` hands the nowcast engine an `Arc<dyn MapEngine>` of its source (reject unknown source, nowcast-of-nowcast). Reload rebuilds atomically as today. The pattern is reusable for future derived collections (accumulation, composite timelines).
- Poll loop on `poll_runtime()` watches the source's `raster_info().times`; each new frame triggers a generation.
- Motion field: block cross-correlation (pySTEPS-style Lucas–Kanade-lite, pure Rust, no new deps) between the last N frames at native grid, estimated only from blocks with echo (also keeps stationary clutter from anchoring the field), robust-filtered, interpolated + smoothed.
[...]

PR #525, opened 2026-07-20
feat(engine-nowcast): phase-0 motion + advection + skill core

### LINKED ISSUE #520: engine-nowcast phase 0: offline motion+advection skill spike (gate: beat persistence)
Part of #519.

Prove the extrapolation core works before any server code exists.

**Deliverables** (new crate `crates/engine-nowcast`, algorithm only — no engine traits yet):

- `motion`: block cross-correlation between two frames on a coarse grid (~32 px blocks), vectors estimated only from blocks with real echo, robust-filtered, filled + smoothed into a continuous block-level field with bilinear sampling.
- `advect`: semi-Lagrangian backward advection — integrate the departure trajectory in substeps, one nearest-neighbour value sample from the analysis frame (preserves sparse-echo sharpness; bilinear smears, see #456 lessons).
- `skill`: contingency table (POD / FAR / CSI) at configurable dBZ thresholds, nodata-aware.
- Unit tests on synthetic translating fields: recovered motion within ±1 px, advected CSI beats persistence.
- `examples/skill_spike.rs`: hindcast harness over a directory of real composite GeoTIFF frames (loads via `GeoTiffEngine` at native grid — same decode path Phase 1 will use). Estimate motion from frames (t−2, t−1)… advect to t, score vs the real frame t, print nowcast vs persistence per threshold.

**Fixtures**: `testdata/smhi-radar-geotiff-4326/` (3 × 5 min), `testdata/smhi-radar/` (3 × 5 min), `testdata/dmi-radar-geotiff-4326/` (3 × 10 min) — local, uncommitted.

**Gate**: nowcast CSI > persistence CSI at 20 dBZ on the fixture hindcasts. If it can't beat persistence at lead 1, the motion estimator is wrong and Phase 1 does not start.

mrauhala:
Phase 0 of the radar nowcasting epic #519. Closes #520.

## What

New crate `crates/engine-nowcast` — the pure extrapolation core, **zero dependencies**, no engine traits yet (those are phase 1, #522):

- **`motion`** — block cross-correlation (SAD) between consecutive frames: vectors only from blocks with real echo (stationary clutter and empty sky can't anchor the field), robust median/MAD outlier rejection, neighbour fill + box smoothing, bilinear per-pixel sampling. Two-stage (coarse+refine) search.
- **`advect`** — semi-Lagrangian backward advection: trajectory integrated in substeps through the field, then **one nearest-neighbour sample** of the analysis frame — deliberately not bilinear, so sparse far-range echoes survive (the #456 lesson). Inflow boundary becomes nodata; nodata never becomes echo.
- **`skill`** — nodata-aware contingency table with POD/FAR/CSI.
- **`examples/skill_spike`** — hindcast harness: loads a GeoTIFF composite sequence through the public `GeoTiffEngine` (the same decode path phase 1 will consume via `Arc<dyn MapEngine>`), estimates motion from each consecutive pair, extrapolates across all available leads, and prints nowcast-vs-persistence skill.

9 unit tests pin: translation recovery within 1 px, identity advection (bitwise), inflow-nodata behaviour, extrapolation-beats-persistence on synthetic motion, and exact contingency/score math.

## Gate result — PASS

Hindcast on a **25-frame live SMHI Sweden composite** (2026-07-20 20:50–22:50Z, 5-min cadence, real widespread precipitation; gain 0.4 / offset −30 / nodata 255 per SMHI's README):

| lead | CSI 20 dBZ nowcast | CSI 20 dBZ persistence | rel. gain |
|---|---|---|---|
| +5 min | **0.676** | 0.623 | +9% |
| +15 min | **0.556** | 0.449 | +24% |
| +30 min | **0.448** | 0.346 | +29% |
| +60 min | **0.339** | 0.273 | +24% |
| +115 min | **0.263** | 0.210 | +25% |

Same picture at 10 dBZ; at 35 dBZ the nowcast wins through ~+55 min, beyond which pure advection can't track convective core growth/decay (known limitation, phase-4 territory). The skill margin **grows with lead**, which is exactly the signature of real motion skill rather than noise.

Cost on a 1006×671 grid (release, M-series laptop): motion ~25 ms per generation, advection ~30 ms per lead frame — a full 24-frame generation lands well under 1 s, comfortably inside the phase-1 poll-loop budget.

## Findings for later phases

- `engine-geotiff` **silently drops** files it can't use during the catalog scan — the spike burned significant time on live SMHI files that are strip-organized (not tiled) and in SWEREF99; the catalog just came back empty with no diagnostic in the example (tracing warnings exist but need a subscriber). The strip-support gap is the old backlog idea; worth a visibility improvement regardless.
- Live-fixture recipe (also in the phase-0 issue): SMHI day-zip → `gdalwarp -srcnodata 255 -dstnodata 255 -t_srs EPSG:4326 -co TILED=YES -co COMPRESS=DEFLATE`. Forgetting `-srcnodata` clamps the out-of-coverage ring 255→254 = fake 71.6 dBZ everywhere, which inflates CSI to ~0.99 and invalidates the whole comparison.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1193,6 +1193,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "engine-nowcast"
+version = "0.1.0"
+dependencies = [
+ "chrono",
+ "ds-core",
+ "engine-geotiff",
+]
+
+[[package]]
 name = "engine-odim"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "crates/engine-postgis",
     "crates/engine-querydata",
     "crates/engine-grib",
+    "crates/engine-nowcast",
     "crates/engine-odim",
     "crates/engine-cap",
     "crates/engine-zarr",
--- a/crates/engine-nowcast/Cargo.toml
+++ b/crates/engine-nowcast/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "engine-nowcast"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+[dev-dependencies]
+# Phase-0 skill spike only (#520): the example loads fixture composites through
+# the public GeoTiffEngine — the same decode path phase 1 consumes via
+# Arc<dyn MapEngine>. The library itself stays dependency-free.
+ds-core = { path = "../core" }
+engine-geotiff = { path = "../engine-geotiff" }
+chrono = "0.4"
--- a/crates/engine-nowcast/examples/skill_spike.rs
+++ b/crates/engine-nowcast/examples/skill_spike.rs
@@ -1,0 +1,370 @@
+//! Phase-0 hindcast harness for the nowcasting epic (#519 / #520).
+//!
+//! Loads a directory of composite GeoTIFF frames through the public
+//! `GeoTiffEngine` (the same decode path phase 1 will consume via
+//! `Arc<dyn MapEngine>`), estimates motion from each consecutive frame pair,
+//! extrapolates, and scores the extrapolation against the frame that actually
+//! followed — next to a persistence baseline.
+//!
+//! Gate: nowcast CSI must beat persistence CSI at the gate threshold, lead 1.
+//!
+//! ```text
+//! cargo run --release -p engine-nowcast --example skill_spike -- \
+//!     --dir testdata/smhi-radar-geotiff-4326 \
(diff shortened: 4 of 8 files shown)


mrauhala:
Both review findings addressed in 18e49b3: `--gate-threshold` not present in `--thresholds` is now a hard error at arg parse (verified: exits with a message instead of silently gating on index 0), and an all-nodata frame prints `range n/a (all nodata)` instead of f32 sentinels. Re-verified the live-sequence run: gate still passes (lead-1 CSI 0.676 vs 0.623).

PR #526, opened 2026-07-21
fix(api): key rendered/meta-tile caches on the concrete latest run, not None

### LINKED ISSUE #521: API caches key on requested reference_time (None ⇒ latest) — stale frames across run boundaries; blocker for nowcast
Part of #519, but stands alone: the staleness is latent for GRIB/QueryData today.

**Mechanism** (`crates/api-wms/src/handlers.rs`, `CacheKey` construction): the rendered/meta-tile cache key stores the **requested** `reference_time` — `None` means "latest run at render time". The #507 fix pinned the *valid time* axis by resolving TIME through `MapEngine::resolve_time` before keying, but the reference axis is still keyed on the request value. When a new run supersedes the latest and re-covers the same valid times with different pixels, every tile cached under `(time, reference_time: None)` keeps serving the old run — the caches are no-TTL, so this never heals.

**Why it's mostly invisible today**: GRIB/QueryData runs arrive hours apart and overlapping valid times far from the run boundary rarely re-render from cache before their LRU eviction. It becomes acute with nowcasting (#519): every generation (~5 min) rewrites *all* future valid times, so a `None`-reference animation would permanently pin the first generation's frames.

**Fix sketch**: at the existing `resolve_time` call site in the handlers, when the request has no `reference_time` and the collection has runs (`RasterInfo.reference_times` non-empty), resolve `None` → the engine's latest reference time and use *that* in the cache key (and pin it for the whole request, mirroring what #508 did for TIME). Apply to WMS + Maps + Tiles key paths. Exact-match/no-run engines are unaffected (`reference_times` empty ⇒ key stays `None`).

mrauhala:
Closes #521. Prerequisite for nowcast phase 1 (#522, epic #519); also fixes a latent staleness for GRIB/QueryData at run boundaries today.

## The bug

The no-TTL rendered + meta-tile caches keyed the run axis on the **requested** `reference_time`, where `None` means "latest run at render time". When a newer run supersedes the latest and re-covers the same valid times with different pixels, every tile cached under `(time, reference_time: None)` keeps serving the old run's pixels — forever, since these caches have no TTL. It's the run-axis twin of the #507 valid-time poisoning that #508 fixed by resolving TIME before keying.

Mostly invisible today because NWP runs arrive hours apart; acute for nowcasting, where every ~5-min generation rewrites *all* future valid times.

## The fix

Resolve the run axis to a **concrete** reference time before any cache key is built:

- **WMS**: a run-less request (and an explicit pin of the current latest — previously normalised to `None`) now resolves to `info.reference_times.last()`. The cache-sharing property is preserved in the opposite direction: both request shapes produce the same concrete-run key, and a new run naturally rolls to fresh keys. The resolved value flows unchanged into the direct path, the meta-tile `TileKeyPrefix`, and `get_raster_tile`.
- **Maps/Tiles**: pin the same way (their `reference_time` query parameter remains a #337 Phase 4 follow-up) and now pass the pinned run to `get_raster_tile`, so a run swap mid-render can't mix runs within one response.
- Non-forecast collections (`reference_times` empty) keep `None` everywhere — no behavior change.

## Tests

- New `run_less_getmap_re_renders_when_a_new_run_lands` replays the incident shape (render → repeat is a pure hit → new run lands → same request must re-render every tile). **Verified failing on the unfixed handler** — `0 fresh renders where 16 required` — and passing with the fix.
- Two existing tests asserting the old None-normalization updated to the new contract (`getmap_default_reference_time_resolves_to_latest_run`, `getmap_explicit_latest_run_shares_key_with_default`).
- Full api-wms / api-maps / api-tiles suites green; clippy clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -194,6 +194,13 @@ gh issue create --title "..." --label "bug,priority: high" --milestone "v0.2"
    `get_raster_tile` uses (share one helper so they cannot drift) — the API
    layers key the no-TTL rendered/meta-tile caches on it. Skipping this
    reintroduces the #507 cache poisoning (stale animation frames).
+   **If the engine retains model runs** (non-empty
+   `RasterInfo.reference_times`), it MUST likewise override
+   `MapEngine::resolve_reference_time` with the SAME run selection
+   `get_raster_tile` uses (`None` ⇒ the concrete run it would render,
+   including any cross-run fallback) — the caches key the run axis on it
+   (#521). Skipping this freezes the first-rendered run's pixels when a
+   newer run re-covers the same valid times.
 8. Ship a runnable (enabled) example collection config AND do an end-to-end
    server + curl smoke test against real data. Unit tests alone miss
    integration and unit-conversion bugs.
--- a/crates/api-maps/src/handlers.rs
+++ b/crates/api-maps/src/handlers.rs
@@ -1010,11 +1010,18 @@ async fn render_map(
     // engines that materialise `RasterInfo` lazily.
     let raster_info = engine.raster_info();
     let time = validated.time.or_else(|| raster_info.times.last().copied());
+    // #521: pin the run axis to the engine's CURRENT latest run before the
+    // cache key is built. The Maps `reference_time` query parameter is still
+    // a follow-up (#337 Phase 4) — the handler always renders the latest run
+    // — but the no-TTL rendered cache must key on the concrete run stamp:
+    // keyed as `None`, the first-rendered run's pixels would keep serving
+    // after a newer run re-covers the same valid times (acute for nowcast
+    // generations, latent for NWP). Empty `reference_times` stays `None`.
+    let reference_time = raster_info.reference_times.last().copied();
     // #507: snap to the exact timestep the engine will render before the
     // cache key is built — a not-yet-ingested datetime must cache the
     // previous timestep's pixels under the PREVIOUS timestep's key.
-    // Maps passes reference_time: None (latest run) throughout.
-    let time = engine.resolve_time(time, None);
+    let time = engine.resolve_time(time, reference_time);
 
     // Parameter selection precedence: ?parameter-name= wins over style.parameter.
     // Validate against the engine's advertised list when the query supplied one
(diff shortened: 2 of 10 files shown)


mrauhala:
All three findings addressed in aa3a99e — and they were correct: the pin disabled GRIB's cross-run fallback (which, per `find_step_for_time`'s unbounded nearest-step snap, triggers for valid times *before* the newest run's reference — the animate-past-frames case). Rather than pinning `reference_times.last()` in the handlers, there's now a `MapEngine::resolve_reference_time` (identity default; run-retaining engines override with the same selection helper `get_raster_tile` uses — the #507 pattern applied to the run axis). GRIB returns the run `resolve_step` actually picks (fallback included, pinned by a new unit test); QueryData mirrors `instances::select_run`. The handlers key the caches on the engine's answer, so a fallback render keys the fallback run. Contract documented in the trait docs and the CLAUDE.md engine checklist.

mrauhala:
Round-3 findings: the Maps/Tiles coverage gap is closed in 1ea064b — both suites now have a forecast-capable run-swap mock replaying the #521 incident shape (cold render → pure cache hit → new run lands → must re-render). On the GRIB triple-snapshot note: acknowledged, accepted as-is — the three `catalog.load()` calls are inherent to `resolve_reference_time`/`resolve_time`/`get_raster_tile` being separate trait calls, and as you note the race fails safe (a vanished pin falls back to the requested time, the render errors, nothing is cached). Consolidating would need a combined resolve API across the trait boundary; not worth the surface change for a fails-safe microsecond race.

mrauhala:
Round-4 note addressed in ffbe3f4: selection now lives in a borrowing `select_run_step` shared by every caller (one authority preserved); the resolve-only paths (`resolve_time`/`resolve_reference_time`) read scalars off the borrow with no `StepFile` clone, and `resolve_step` keeps the clone only for the actual read paths. `ResolvedStep` slims to just the file.

mrauhala:
Round-5 finding confirmed as a real regression and fixed in 041a1a0: the explicit-latest→`None` normalization is reinstated *before* `resolve_reference_time`, so a client echoing the capabilities `default=` keeps GRIB's cross-run fallback, while the engine still returns the concrete run for the cache key. New `getmap_explicit_latest_pin_keeps_fallback_resolution` test uses a fallback-shaped mock to pin both halves — including that an explicit older-run pin shares the cache entry with the fallback-resolved request (same concrete key).

PR #529, opened 2026-07-21
perf(engine-nowcast): O(leads) trajectory integration — 6-12× faster generations

### LINKED ISSUE #528: engine-nowcast: generation is O(leads²) in trajectory integration — 67 s on prod for 24 leads
Part of #519. Observed at prod enablement of `fmi-radar-nowcast` (2.27 Mpx working grid, 24 leads): `nowcast_last_generation_duration_ms = 66659` on nexus. Harmless today (background runtime, once per 5 min) but it occupies one of the four poll-runtime workers for ~a quarter of every cadence interval, and scales badly with horizon.

**Root cause**: `advect::for_each_departure` integrates each lead's trajectory from scratch with `steps = lead × substeps`, so total work is `Σ i·substeps·pixels` — **quadratic in the lead count** (~2.7 G field samples for 24 leads at 2.27 Mpx).

**Fix**: compose displacement maps incrementally — derive lead *i*'s departure map from lead *i−1*'s plus one interval of integration (the standard semi-Lagrangian long-lead scheme), making generation O(leads). Value sampling stays a single nearest-neighbour read from the analysis frame, so output is unchanged in spirit (tiny trajectory-precision differences); keep the phase-0 skill harness as the regression gate.

Expected: ~12× fewer field samples for the 24-lead case → prod generation in the low seconds.

Refs: #522 (phase 1), prod metrics 2026-07-22.

mrauhala:
Closes #528. Unblocks raising `max_pixels` in the prod `fmi-radar-nowcast` config — the visible-resolution complaint: the 4 Mpx default puts the working grid at ~1 km vs the source's 250 m, and at the current 67 s/generation a 4× pixel increase would blow the 5-minute cadence.

## What

Generation advected every lead from scratch with integration steps proportional to the lead — `Σ i·substeps·pixels`, quadratic in the lead count (~2.7 G field samples for prod's 24 leads at 2.27 Mpx → 67 s on nexus).

`TrajectoryIntegrator` keeps one departure-point set per generation and extends it by one lead-delta per iteration — the standard long-lead semi-Lagrangian scheme. For a uniform lead schedule this walks the **identical** piecewise trajectory (same step size, same step sequence), pinned bitwise by the new `incremental_advances_match_one_shot_bitwise` test. `advect`/`advect_u8` stay as one-shot wrappers over the same core so the two paths cannot drift; the phase-0 skill harness is unchanged and still applies.

## Measured

Same debug-build smoke config (25-frame SMHI live sequence, 12 leads at 0.67 Mpx): **14,376 ms → 2,380 ms** — 6.0×, matching the `(leads+1)/2` theoretical factor. Prod (24 leads) predicts ~12×: 67 s → ~5 s.

## Follow-up once deployed

Set `max_pixels = 16_000_000` on `fmi-radar-nowcast` (nexus collections.d): one halving instead of two → ~500 m working grid (2481×3658), ~4× the pixel density of today's output. Generation at that size with this fix ≈ the ~20 s range on nexus — comfortably inside cadence. Memory: ~227 MB/generation × 2 retained ≈ 450 MB, within the box's headroom.

### DIFF
--- a/crates/engine-nowcast/src/advect.rs
+++ b/crates/engine-nowcast/src/advect.rs
@@ -11,59 +11,111 @@
 use crate::motion::MotionField;
 use crate::Grid;
 
-/// Walk every output pixel's backward trajectory and hand the caller the
-/// source-pixel index it departs from (`None` when the trajectory leaves the
-/// domain). Shared core for the `f32` and raw-`u8` advection variants so
-/// their trajectory math cannot drift apart.
-fn for_each_departure(
+/// Incrementally extended backward trajectories — one per output pixel.
+///
+/// A generation advects the same analysis frame to a *schedule* of leads.
+/// Re-integrating each lead's trajectory from scratch costs
+/// `Σ leadᵢ·substeps·pixels` — quadratic in the lead count (67 s per prod
+/// generation at 2.27 Mpx × 24 leads, #528). Extending the stored departure
+/// points by one lead-delta at a time walks the SAME piecewise trajectory
+/// (identical step density and step sequence when the schedule is uniform)
+/// in `Σ substeps·pixels` — linear.
+///
+/// Departure points keep integrating even outside the domain (the motion
+/// field's edge-clamped sample keeps them moving) exactly like the one-shot
+/// integration did; bounds are only checked when sampling a frame.
+pub struct TrajectoryIntegrator<'a> {
+    field: &'a MotionField,
     width: usize,
     height: usize,
-    field: &MotionField,
-    lead: f32,
-    substeps: usize,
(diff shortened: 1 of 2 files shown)


mrauhala:
Both round-1 findings addressed in 004585a: the equivalence claim is now stated precisely — bit-identical only when `delta × substeps` is integral (the default source-cadence config, which is what the bitwise test pins), while a fractional explicit `step` gets per-segment discretization that is at least as fine as the one-shot proportional share (equal-or-better integration accuracy, never coarser). New `incremental_fractional_delta_stays_within_sampling_tolerance` test covers the fractional case (3 × 0.6 vs one-shot 1.8) within nearest-sample tolerance.

Issue #524, opened 2026-07-20
engine-nowcast phase 3: motion smoothing across generations, PVOL cell-track fusion, dashboards


mrauhala:
Part of #519. Quality pass after phases 1–2 are serving.

- Temporal smoothing of the motion field across generations (blend with previous field) — kills frame-to-frame jitter in the animation.
- Optional lead-time blur/decay knob (off by default; pure Lagrangian persistence stays v1 behaviour).
- Fuse PVOL cell-track vectors (`ds_core::cells` `motion_ms` / `Track::speed_direction`) where sites overlap the composite, weighted into the block field.
- Grafana row for the nowcast metric families (repo rule: new /metrics family ⇒ dashboard update).
- Hindcast skill trend: log CSI-vs-persistence per generation so quality regressions are visible in ops, not just at review time.

mrauhala:
Scope pin for the Grafana row (flagged in #543 review): the dashboard work here covers **all eight** nowcast metric families — the six from phase 1 (`nowcast_generations_total`, `nowcast_generation_failures_total`, `nowcast_last_generation_duration_ms`, `nowcast_source_lag_seconds`, `nowcast_retained_generations`, `nowcast_frames`) plus the two realized-skill gauges from #542/#543 (`nowcast_lead1_csi_permille`, `nowcast_lead1_persistence_csi_permille` — plot as a pair; the gap between them is the signal, and a collapsing gap means motion/data regression).

mrauhala:
Closing — every item is either shipped or explicitly re-homed:

- **Temporal motion smoothing across generations** — shipped in PR #530 (multi-pair estimation + cross-generation EMA 0.7/0.4; killed the rubber-band artifact).
- **Lead-time blur/decay knob** — resolved by the #546 verdict: three classical growth/decay estimators failed the hindcast gate (re-confirmed post-matching-fix with production-parity tracking), so the rendered field stays pure Lagrangian persistence deliberately; the `growth_decay` flag exists but stays off. Rendered decay now rides the V2.5 L-CNN route on #541.
- **PVOL cell-track fusion into the block field** — deprioritized on #541 (the epic's evidence sweep found no operational precedent that beats a well-stabilized field estimator); the tracker-convergence work that would precede it is scoped in #551 (shared overlap-based split/merge lineage for `ds_core::cells` + `cells2d`).
[...]

PR #530, opened 2026-07-22
feat(engine-nowcast): motion stabilization — multi-pair estimation + cross-generation EMA

mrauhala:
Part 1 of #524 (leaves it open — PVOL cell-track fusion, Grafana row, and per-generation skill logging remain). Fixes the user-reported "stretchy elastic back-and-forth" animation artifact, worst on convective cells.

## Diagnosis

Two stacked effects:
1. **Between generations** (the back-and-forth): motion was estimated from a single frame pair, and convective growth/decay reads as spurious motion to a block matcher. Every 5-minute generation re-estimated from scratch, so the +1–2 h frames swung position each time the client re-pinned — classic nowcast rubber-banding (the reason pySTEPS defaults to multi-frame motion).
2. **Within a loop** (the stretching): at the prod 500 m working grid the coarsened motion blocks were ~25 km — a 3–5 km cell sits inside the bilinear gradient between block vectors and gets differentially advected, elongating with lead time.

## Fix

- **`estimate_motion_multi`**: every consecutive history pair contributes block measurements (scaled to the last interval's unit for mildly irregular cadences; degenerate pairs skipped), averaged per block, then ONE shared outlier/fill/smooth pass. The engine fetches up to `history_frames` (default 3) frames per generation, all pinned to one source run; a missing older frame degrades to fewer pairs instead of failing the generation.
- **Cross-generation EMA**: each new field blends with the previous generation's (`0.7` new for measured blocks, `0.4` for filled — inferred vectors lean harder on history; auto-skipped when the block grid changes). The blended field is stored on the generation as the next one's history — ~1.5 generations of memory, enough to damp pair noise without lagging a real wind shift by more than a couple of cadence intervals.
- **`TARGET_SEARCH_PX` 24 → 48**: the FMI ~500 m grid stays uncoarsened, halving effective motion-block size to ~16 km so small cells get a closer-fitting vector — affordable since #529 made generation cost linear in leads.

## Verification

- 23 tests: new multi-pair units (uniform-motion agreement with single-pair, degenerate-interval skipping), EMA blend units (per-block alpha math, dimension-mismatch no-op), and a 3-frame end-to-end extrapolation gate (IoU > 0.8 against the mock source's closed-form future). All prior gates unchanged.
- Generation cost on the smoke fixture: 2,516 ms vs 2,380 ms (~5% for the extra pair). Expect a similar marginal increase on prod's 27 s.
- The artifact itself is an animation-domain effect — final verification is the tutka loop after deploy; the between-generation oscillation should visibly damp within two generations of the deploy.

### DIFF
--- a/crates/engine-nowcast/CLAUDE.md
+++ b/crates/engine-nowcast/CLAUDE.md
@@ -44,9 +44,21 @@ reason `MapEngine::resolve_reference_time` exists (#521).
   (1 B/px; `advect_u8` moves bytes); anything else falls back to f32
   (4 B/px — the FMI S3 COG path lands here until #475-style typed paths).
 - Motion is estimated on a coarsened grid sized so the physical search
-  window (40 m/s × source interval) fits ~24 px, then the field is scaled
-  back — deliberate scale handling; do NOT rely on the pixel budget to do
-  this implicitly.
+  window (40 m/s × source interval) fits `TARGET_SEARCH_PX` (48 — keeps
+  the FMI 500 m grid uncoarsened), then the field is scaled back —
+  deliberate scale handling; do NOT rely on the pixel budget to do this
+  implicitly.
+- **Motion stabilization (#524 part 1)** — two mechanisms against the
+  rubber-band animation artifact (single-pair block matching re-reads
+  convective growth/decay as motion noise every generation):
+  1. multi-pair estimation: every consecutive history pair contributes
+     measurements (scaled to the last interval's unit), averaged per
+     block, then ONE shared outlier/fill/smooth pass;
+  2. temporal EMA with the previous generation's field
+     (`Generation.field`, weights `EMA_ALPHA_MEASURED = 0.7` /
+     `EMA_ALPHA_FILLED = 0.4`; auto-skipped when the block grid changes).
+  Weakening either brings the between-generation oscillation back —
+  verify with an animation loop, not stills.
 
 ## Memory sizing (retention multiplies!)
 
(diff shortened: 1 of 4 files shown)


mrauhala:
Finding addressed in 687d82b: `history_frames` is now capped at 8 (`MAX_HISTORY_FRAMES`), rejected fail-fast at construction with a test — each frame is one sequential blocking source fetch per generation (Critical Rule 9), and pair-averaging saturates after a few pairs anyway.

mrauhala:
Fair challenge — the smoke fixture indeed never exercised the coarsening path (its shift is ~3 px), so I measured at **exact prod scale**: the 3 newest FMI composite frames from prod, `max_pixels = 16M` (2481×3658 working grid, 24 leads, `history_frames = 3`), release build, this branch:

| configuration | generation (release, M-series Mac) |
|---|---|
| `TARGET_SEARCH_PX = 48` (this PR: factor 1, search 32, motion at 2481×3658) | **10,416 ms** |
| `TARGET_SEARCH_PX = 24` (old path: factor 2, search 16, motion at 1240×1829) | **10,435 ms** |

The constant's cost is in the noise (<0.2%): motion blocks are echo-gated and the stage is dominated by neither — generation time is fetch (3 × 36 Mpx source decode + resample) plus advection (24 × 9 Mpx). The only real addition from this PR is the third source fetch, so prod should land ~10–20% above its current 27 s (call it ~30–32 s), with `source_lag_secs` confirming post-deploy as you suggest.

On the 30 s `poll_interval_secs` margin: that interval is only the *check* cadence, not a deadline — `MissedTickBehavior::Skip` drops ticks *during* a generation and the next tick re-checks the source, so a generation longer than the poll interval costs nothing beyond worst-case staleness of `generation + poll_interval` (~60 s), against the 300 s source cadence. The actual ceiling is generation time approaching the source cadence, which is 10× away.

Issue #541, opened 2026-07-24
epic: nowcasting v2 — growth/decay, cell intelligence, object-based verification


mrauhala:
## Triage update — 2026-09-16

Review baseline: [main `524a1df`](https://github.com/mrauhala/meteocore/commit/524a1df934d8da97c34af943f43df892f97fb704).

**Disposition:** Keep as an epic; prioritize remaining children.

**Current assessment and challenge:** Several v2 phases shipped; experimental growth/decay remains default-off after failed gates. Rank evidence/quality gaps before further modelling scope.

The original roadmap below is historical context; shipped phases should not be counted as new work or as independent priority candidates. Remaining child issues retain their individual implementation and verification scope.

---

<details>
<summary>Original report / roadmap (historical context; current scope is above)</summary>

Successor to #519 (extrapolation nowcast, shipped and live in prod). Scope and ordering are grounded in a July 2026 literature sweep (2023–2026, primary sources); the design principles below each carry their evidence.

## What the evidence says

1. **Keep the classical advection core; augment it.** FMI's own deep-probabilistic model DEUCE (Harnist et al., GMD 2024, 10.5194/gmd-17-3839-2024) only roughly matches STEPS on CRPS, and its ensemble mean is *worse* than plain advection as a deterministic nowcast. Nothing generative is confirmed operational at a national met service: NowcastNet (Nature 2023) makes no deployment claim, Météo-France's DGMR experiment was a dormant 2024 demo repo, LDCast is research. Meanwhile NowcastNet's own "evolution network" is structurally advection + learned residuals — the field converged on our architecture.
2. **The proven quality jump is growth/decay on top of advection**, two flavors:
   - *Classical*: NowPrecip 2 (Sideris et al., QJRMS 2026, 10.1002/qj.70171 — operational at MeteoSwiss; verification companion 10.1002/qj.70181 benchmarked vs pysteps + ICON to 4 h). Its innovations — improved optical flow, **intensity-based growth/decay profiles**, temporal disaggregation — are explicitly "straightforward in a programming sense" and offered as transferable. No deep learning. CPU-only compatible.
[...]

PR #543, opened 2026-07-24
feat(engine-nowcast): object-based verification harness + per-generation skill logging

mrauhala:
Nowcasting v2 phase **V2.1** (#541). Closes #542. This is the measurement foundation: every later v2 quality phase (growth/decay, cell hazard ML, learned residuals) gates on the numbers this produces.

## What

Modeled on the object-based verification framework of Ritvanen et al. (GMD 18, 1851–1878, 2025 — the paper that quantified how advection-based nowcasts miss growth/decay):

- **`objects` module** (dependency-free, like the rest of the algorithm core): 2D cell segmentation on composite frames (35 dBZ default contour, 8-connected, min-area filter, exceedance-weighted centroids), forecast↔observed cell matching by O(n³) Hungarian assignment under a centroid-distance gate (20 km default), growing/decaying classification from the volume-proxy derivative, and object contingency scores (POD/FAR/CSI + mean centroid error) attributed to the creation-time class.
- **Harness**: `skill_spike` now prints the object table (overall / growing / decaying CSI by lead vs the persistence baseline + centroid error in km) alongside the existing pixel table; growth classes chain forward through observed-cell matches at each lead. Run on the 25-frame live SMHI sequence: centroid error 3.7 km at +5 min → ~15 km at +2 h, and object CSI ≈ persistence at short leads on stratiform rain — an honest discriminator that pixel CSI (which shows a comfortable nowcast win) glosses over. Exactly what the V2.3 growth/decay work needs to move.
- **Continuous prod skill telemetry**: each generation scores the previous generation's prediction for the fresh analysis frame (pixel CSI at `min_echo`) next to the persistence baseline — one log line + two gauges (`nowcast_lead1_csi_permille`, `nowcast_lead1_persistence_csi_permille`). A collapsing gap in prod = motion/data regression, visible in ops rather than at review time. Absorbs the #524 skill-logging leftover (Grafana row remains tracked there).

## Tests

9 new: segmentation (centroids, min-area), matching (gate respected, globally optimal pairing on the classic greedy-fails geometry), growth classification, score attribution (hits/misses/false alarms + per-class), and the engine-level gauge test (translation nowcast CSI ≥ persistence, > 0.9). Suites: engine-nowcast 21 lib + 9 integration, server green, clippy `-D warnings` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #545, opened 2026-07-24
feat(engine-nowcast): cell intelligence — severity rank, tracks, deviant-mover flag

### LINKED ISSUE #544: nowcasting v2.2: cell intelligence — 2D severity rank, tracks, deviant-mover flag (Features serving)
Part of #541 (phase V2.2), building on the V2.1 objects module (#543).

**Scope revision vs the epic text**: severity uses the *verified* 2D composite cells (area, max dBZ, volume proxy) — NOT VoxelGrid attributes: the voxel implementation is currently beta/non-verified and slow (owner's call, 2026-07-25). Echo-top/VIL enrichment upgrades the rank later, once voxels are verified.

Part 1 (this issue):
- Per-generation cell tracking in engine-nowcast: segment the analysis frame (35 dBZ default), chain-match against the previous generation's cells (anisotropic km matching from V2.1) → track id, age, velocity, volume trend.
- TRT-lite severity rank (documented heuristic v1): weak/moderate/severe/very-severe from max dBZ + area + volume proxy.
- Deviant-mover flag: cell track velocity vs the generation's motion field sampled at the cell — sustained residual (angle + speed gates) marks storms moving off the ambient flow (right-mover detector; also the future fusion signal).
- Serve as GeoJSON Point features (centroid + properties: rank, class, max dBZ, area km², velocity m/s + bearing, deviant flag, track age) via FeatureEngine on the nowcast collection (`apis += features`). Footprint polygons = follow-up.

Part 2 (separate issue when this lands): per-cell lightning join (flash counts + sigma-style jump flag per track footprint) — needs a small ds-core event-source trait + engine-postgis impl + second-pass wiring.

mrauhala:
Nowcasting v2 phase **V2.2 part 1** (#541). Closes #544.

Built on the verified 2D objects machinery from V2.1 — deliberately **not** on VoxelGrid attributes (beta/unverified/slow per owner's call); echo-top/VIL enrichment revisits once voxels are verified.

- **`cells2d`**: per-generation tracking of the analysis frame's 35 dBZ cells — anisotropic km matching, persistent track ids, age, velocity EMA; **TRT-lite severity** (max-dBZ steps 45/50/55 + area ≥ 50 km², documented heuristic that V2.4's tree model replaces); **deviant-mover flag** = sustained ≥ 5 m/s residual between track velocity and the ambient motion field sampled at the cell (the estimator-disagreement right-mover detector, also the future fusion signal).
- **`FeatureEngine` on the nowcast collection** (`apis += features`): tracked cells as GeoJSON Point features — `severity`, `max_dbz`, `area_km2`, `speed_ms`, `bearing_deg`, `deviant_mover`, `track_age`, `observed` — with bbox filtering and pagination.
- Lightning join (per-track flash counts + jump flag) is part 2: needs a ds-core event-source trait + engine-postgis impl + second-pass wiring.

Tests: severity monotonicity; a two-cell scenario where a small cell moves against the ambient flow of a dominant large cell — the counter-flow cell gets flagged (streak semantics), the with-flow cell doesn't, speeds/bearings verified; Features integration (id persistence across generations, properties, 404). 36 total green, clippy clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -323,7 +323,7 @@ they were found. Critical Rules 5–7, 9 and 10 above are part of this set.
 | ODIM PVOL | `EdrEngine` + `MapEngine` + `VolumeEngine` (per-site views) + `FeatureEngine` (network engine) | EDR (position, locations, area, trajectory), WMS, Maps, Tiles, 3D Tiles, Features (site inventory) |
 | QueryData | `EdrEngine` + `MapEngine` | EDR (position only), WMS, Maps, Tiles |
 | Zarr | `EdrEngine` + `MapEngine` | EDR (position), WMS, Maps, Tiles; local + S3/HTTP |
-| Nowcast | `MapEngine` (derived: wraps another collection's engine) | WMS, Maps, Tiles — motion-extrapolated frames with future TIME values (EDR + instances = phase 2, #523) |
+| Nowcast | `MapEngine` + `FeatureEngine` (derived: wraps another collection's engine) | WMS, Maps, Tiles — motion-extrapolated future frames; Features — tracked cell intelligence (severity, deviant movers, #544). EDR + instances = #523 |
 | PostGIS | `EdrEngine` + `FeatureEngine` + `MapEngine` (events shape only) | EDR (position, locations, area), Features; events shape: EDR (area) + WMS/Maps/Tiles (age-colored strike layer) |
 
 ## Config Format
--- a/crates/engine-nowcast/CLAUDE.md
+++ b/crates/engine-nowcast/CLAUDE.md
@@ -85,6 +85,18 @@ follow-up (full frames only for the latest generation) is scoped in #523.
   `nowcast_lead1_persistence_csi_permille`. A persistent gap collapse in
   prod = motion or data regression; check before blaming the client.
 
+## Cell intelligence (V2.2, #544)
+
+- `cells2d` tracks the analysis frame's 35 dBZ cells across generations
+  (anisotropic km matching): TRT-lite severity (max dBZ 45/50/55 steps +
+  area ≥ 50 km² — deliberately NOT VoxelGrid attributes while voxels are
+  beta/unverified), velocity EMA, and the deviant-mover flag = sustained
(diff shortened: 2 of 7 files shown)


mrauhala:
Both findings addressed in 3e05e50 (same root cause): velocities are now stored in km/s — track displacement divides by the actual previous-anchor→anchor span (also robust to skipped generations), field vectors divide by the source interval stored on the Generation, and the Features `speed_ms` no longer references the lead step at all.

mrauhala:
Round-2 findings addressed in 88202e2: `?datetime=` is honored (an interval excluding the latest anchor returns an empty page — engine-cap precedent), and `feature_count()` is overridden to read the snapshot length instead of materializing every feature on each collection-metadata request. Both pinned in the integration test.

mrauhala:
Round-3 finding addressed in 3b78639: track continuity now requires the previous generation's grid geometry to equal the new one (full `GridGeom` comparison, same pattern as the #543 skill guard) — on any source geometry change the track set resets and cells restart as newborns rather than matching reinterpreted centroids.

mrauhala:
Round-4 findings addressed in ae75e77: the matching gate scales linearly with the actual previous-anchor→anchor span (`TRACK_GATE_KM` per `TRACK_GATE_BASE_SECS`, floored at one base gate) — a skipped generation now widens the gate instead of breaking fast storms' tracks; and `geometry_change_resets_cell_tracks` drives the reset branch with a mutable-extent mock, pinning that tracks restart as newborns.

mrauhala:
Round-5 findings addressed in e7b0a79: `get_feature` locates the track by id and builds exactly one feature via the shared `cell_feature` helper (both paths share one construction, no full-page materialization — cells are a Vec snapshot, so it's an O(n) scan over dozens of tracks rather than a HashMap, matching the data shape); the server wiring test now runs the nowcast collection with `apis = [wms, features]` and asserts the features registry.

mrauhala:
data_version finding addressed in c69683b: overridden to return `generations_total` (cells rebuild per generation), matching the other FeatureEngine impls — the stale-MVT-ETag trap is removed outright rather than documented.

PR #547, opened 2026-07-25
feat(engine-nowcast): growth/decay profile mechanism (off by default — scene-wide gate failed)

### LINKED ISSUE #546: nowcasting v2.3: intensity-based growth/decay profiles (NowPrecip-2-inspired)
Part of #541 (phase V2.3) — the "cells never die" fix, gated by the V2.1 object harness.

Method (NowPrecip 2 lineage — Sideris et al. QJRMS 2026, classical, CPU-only):
- **Measure**: per generation, advect the previous source frame one interval along the motion field and compare with the actual anchor frame *in the Lagrangian frame*, binned by intensity band (~2 dBZ bins above min_echo) → mean tendency (dBZ/interval) per band. EMA the profile across generations to damp scene noise.
- **Apply**: when building lead k, adjust each advected value by its band's tendency × k, damped with an e-folding time (~45 min) so long leads relax toward pure advection instead of running away; clamped to a sane range; nodata untouched. U8 path applies in raw-byte space via gain.
- **Config**: `growth_decay = false` default (flip in prod only after the gate passes).
- **Gate (V2.1 harness)**: growing/decaying object POD and pixel CSI-by-lead must improve or hold vs pure advection on the SMHI live sequence + an FMI sequence; the skill_spike gains a `--growth-decay` toggle to run the A/B.

mrauhala:
Part of #541 phase V2.3 / #546 — **does not close #546**: the mechanism lands, the quality gate did not pass, and the flag defaults off.

## What landed

`tendency::GrowthProfile` — the NowPrecip-2-lineage classical method: per-intensity-band tendencies measured in the Lagrangian frame (previous frame advected one interval vs the actual anchor), sparse-band fill from nearest measured neighbours + band-centre interpolation at apply time (measurement keys by the previous value, application by the current one — adjacent-band holes would silently disable the profile), cross-generation EMA, e-folding damped integration along leads. Wired behind `[collections.nowcast] growth_decay` (**default `false`**); measured every generation even when off, so enabling starts from a warmed profile. `skill_spike` gains `--growth-decay` for A/B gating.

## Honest gate result (the V2.1 harness doing its job)

SMHI 25-frame stratiform hindcast, pixel CSI at 20 dBZ vs **pure advection**:

| lead | advection | GD (45-min e-fold) | GD (15-min e-fold) |
|---|---|---|---|
| +5 min | 0.676 | 0.651 | 0.654 |
| +30 min | 0.448 | 0.322 | 0.401 |
| +60 min | 0.339 | 0.118 | 0.305 |
| +115 min | 0.263 | 0.029 | 0.240 |

Scene-wide profiles lose at **every** lead/threshold: the scene-mean tendency in widespread stratiform is mildly negative, and integrating it drains the entire field below threshold rather than decaying the cells that are actually dying. This matches why NowPrecip 2 embeds its profiles in a *localization* architecture (regional profiles), not scene-wide ones. Gate status is documented on `EFOLD_INTERVALS` and in #546, which stays open for the iteration: localized/per-region profiles, growth-class-conditioned application, or the learned Lagrangian residual (V2.5).

41 tests green (new: measurement/e-fold math, sparse-fill + clamps, EMA, and an engine-level fading-source test proving application works end to end); clippy clean.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -215,6 +215,11 @@ pub struct NowcastConfig {
     /// field.
     #[serde(default = "default_nowcast_min_echo")]
     pub min_echo: f64,
+    /// Apply intensity-based growth/decay profiles along the leads (#546,
+    /// NowPrecip-2 lineage). Default off — flip per collection once the
+    /// object-harness gate has passed for that source.
+    #[serde(default)]
+    pub growth_decay: bool,
 }
 
 fn default_nowcast_horizon() -> String {
--- a/crates/engine-nowcast/examples/skill_spike.rs
+++ b/crates/engine-nowcast/examples/skill_spike.rs
@@ -26,6 +26,7 @@ use engine_nowcast::objects::{
     ObjectScores, PixelScale,
 };
 use engine_nowcast::skill::{score, Contingency};
+use engine_nowcast::tendency::GrowthProfile;
 use engine_nowcast::{advect::advect, Grid};
 
 /// Keep the working grid at most this many pixels (halve dims until it fits).
@@ -46,6 +47,7 @@ struct Args {
     object_threshold: f32,
     min_area: usize,
     gate_km: f64,
+    growth_decay: bool,
 }
 
 fn parse_args() -> Result<Args, String> {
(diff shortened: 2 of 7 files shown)


mrauhala:
Finding addressed in e182adb: the collision guard steps away from the nodata byte in whichever direction exists (0 → 1, otherwise −1), so a hypothetical `nodata = 0` source can no longer alias adjusted samples into missing data.

mrauhala:
Doc mismatch fixed in 29a86f4: the comment now states the intent precisely — the floor at `base − BAND_WIDTH` is deliberate, so decay out of the lowest band registers as ~one band's loss rather than zero.

PR #548, opened 2026-07-25
feat(engine-nowcast): per-cell growth/decay tendencies (v2.3 iteration 1, draft)

mrauhala:
Iteration 1 on #546 (draft until the convective gate runs). Full findings trail on the issue.

**Where the design landed after two dead ends**: scene-wide profiles drain the whole field (gate failure in #547); class-conditioned per-band profiles silently no-op (single-digit samples per band×class on real cells). The estimator now lives at the **track level**: each cell carries an EMA'd mean-exceedance trend measured from its own matched blobs — robust to pixel misalignment — and application advects the cell **label map** along the lead trajectories so each cell's footprint evolves by its own tendency, e-fold damped. Background and newborn cells stay pure advection. Cell tracking moved inside `generate()` so labels exist at measurement time.

**Gate status**:
- SMHI stratiform: 20 dBZ CSI **identical to pure advection at every lead** — the drain failure mode is structurally gone. 35 dBZ shows a marginal loss (≤0.009 CSI) in the harness variant, which lacks the engine's track EMA (single-pair noise on tiny cells).
- Outstanding before un-drafting: harness parity with the engine's EMA'd tendencies, and the positive gate — decaying-cell POD / 35 dBZ CSI win on a **convective FMI sequence** (needs ~20 FMI frames fetched from nexus, ~55 MB).

`growth_decay` remains default-off. 41 tests green (fading-source test now exercises the two-generation tendency path), clippy clean.


mrauhala:
Added to this draft (client-visible prod bug): motion-compensated first-guess matching in `advance_tracks` — raw nearest matching paired fast movers with different upstream cells, producing the against-flow velocities observed on the map; and `CELL_MIN_AREA_PX` 5→10 to cut weak-speck churn. Client guide delivered separately (analysis-instant-only semantics, default filter tier, age gates for arrows).

mrauhala:
All review findings addressed + one feature added in 73305a4: (1) label overflow now falls back to advection (no 255 collision); (2) two-hypothesis matching — motion-compensated pass, then a raw-position pass on leftovers, so counter-flow cells outside the compensated gate still match (the deviant-detection regression flagged); (3) ClassedProfiles/GrowthProfile dead code removed, gate verdict documented in tendency.rs; (4) `growing` surfaced as `volume_trend` property; (5) stale doc fixed. NEW: **cell-snapshot history** — 48 retained snapshots (~4 h), `?datetime=` selects the newest snapshot in the interval so animating clients get the exact per-frame cell situation; temporal_extent advertises the span.

mrauhala:
Round addressed in e04a999: label-cap comment no longer contradicts itself (single statement: labels 1..=254 carry tendencies, >254 falls back to advection), and the crate CLAUDE.md now documents the Features history contract (48 snapshots ≈ 4 h, ?datetime= = newest snapshot in interval, temporal extent = retained span). Also on the record: the #546 verdict was re-tested with the fixed production tracker chained in skill_spike — tendencies still lose at 35 dBZ every lead and decayPOD drops (details on #546).

mrauhala:
Round addressed in c1b58a3: added `counter_flow_cell_matches_via_raw_position_pass` — a stationary cell under a 30 px/interval flow whose compensated hypothesis exits the 20 km gate; asserts the pass-2 raw-position rescue carries id/age and that velocity comes from original centroids (~zero, not the flow).

mrauhala:
Round addressed in e12d70a: (1) `label_map_is_one_based_compacted_and_drops_specks` pins the label-map contract directly (per-pixel labels on a 2-blob + dropped-speck grid, label areas == blob areas, creation order); (2) crate CLAUDE.md Gotchas now records that `growth_decay = true` adds a second full-grid sample per lead (~2× sampling cost vs #528's budget) — relevant only if the flag ever flips on, which the #546 verdict says it shouldn't.

PR #550, opened 2026-07-25
feat(nowcast): per-cell lightning join — flash rate + Schultz jump flag

### LINKED ISSUE #549: nowcasting v2.2 part 2: per-cell lightning join (flash rate + jump flag)
Part 2 of the v2.2 cell-intelligence phase (#544 shipped part 1; parent epic #541).

Join the live lightning events DB onto tracked nowcast cells, per generation:

- **ds-core `events` module**: framework-free `EventPoint {time, lon, lat}` + `EventSource` trait (`recent_events(start, end, limit)`), so engine-nowcast never depends on engine-postgis.
- **engine-postgis**: implement `EventSource` for the `events` shape, reusing the #504 `build_events_window` SQL (time+coords, newest-first, bounded) on the existing sync bridge — one bounded query per generation from the background poll runtime.
- **server wiring**: an `event_sources` registry (postgis events-shape collections); `[collections.nowcast] lightning_source = "<collection id>"` attaches one to a nowcast engine in the second pass (missing id ⇒ collection fails at load, not silently).
- **engine-nowcast**: strike→cell binning via the existing label map (strike lon/lat → grid px → label; unlabeled strikes fall back to nearest centroid within a km gate), per-track flash count + flash rate, and a Schultz-style 2σ lightning-jump flag over the track's recent rate history. Surfaced as `flash_count` / `flash_rate_per_min` / `lightning_jump` Feature properties (omitted when no source is configured).
- Event-source failure degrades gracefully: warn + skip the join for that generation, never fail the generation.

ProbSevere v3 (WAF 2024) found per-cell flash rate the top severe-wind predictor; this lands the plumbing and the classical jump flag ahead of the V2.4 tree model.

mrauhala:
Closes #549 (V2.2 part 2, epic #541).

## What

Tracked nowcast cells gain live lightning intelligence: per-cell `flash_count`, `flash_rate_per_min`, and a Schultz-style 2σ `lightning_jump` flag, joined from an engine-postgis events-shape collection (the lightning strike DB) once per generation.

## How

- **ds-core `events` module** — `EventPoint {time, lon, lat}` + `EventSource` trait: half-open `(start, end]` window, ascending, newest-kept truncation, one bounded call per consumer cycle. Framework-free; engine-nowcast never depends on engine-postgis.
- **engine-postgis** — `impl EventSource for PostgisEngine` (events shape only), reusing the #504 `build_events_window` SQL and the engine's sync bridge. The ds-storage calling rules apply and hold: the only caller is the nowcast poll loop on the multi-thread background runtime.
- **server** — `event_sources` registry populated from events-shape postgis collections; the nowcast second pass resolves `[collections.nowcast] lightning_source = "<id>"` from it. Missing/non-events id ⇒ the nowcast collection FAILS at load (no silent no-flash serving).
- **engine-nowcast** — strikes project to the working grid and bin by the analysis label map (`segment_cells_labeled` contract); unlabeled strikes join the nearest cell centroid within `LIGHTNING_JOIN_RADIUS_KM = 10` (anisotropic km); beyond that they're dropped. Rates feed a per-track 6-generation history; jump = rate > baseline mean + 2σ, with the published 10 fl/min Schultz floor (documented for re-tuning against Nordic storm data). Source errors degrade to a warning + null flash fields for that generation — the generation itself never fails.

## Property semantics

Absent = no source wired ("not measured"). Null = join skipped this generation (source error). `0` / `false` = measured and quiet.

## Testing

- Unit: label-then-radius attribution with an out-of-gate strike dropped; jump detector needs ≥2 baseline rates, respects the absolute floor, fires on a burst over a quiet baseline, history bounded.
- Integration: mock event source dropping a 30-strike burst on the moving disc — properties present with correct values across two generations; a source-less engine emits no flash properties at all.
- `cargo test --workspace`: 69 suites green. fmt + clippy `-D warnings` clean on ds-core, engine-nowcast, engine-postgis, server.

## Prod enable (after deploy)

Add to `fmi-radar-nowcast.toml`: `lightning_source = "<lightning collection id>"` under `[nowcast]`. Live smoke deferred to prod (the strike DB only exists there) — same pattern as v2.2 part 1.

https://claude.ai/code/session_01Edrv5Cf4SosH4YmM9tBnEJ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -403,6 +403,8 @@ apis = ["wms", "maps", "tiles"]
 source = "radar"        # collection id to extrapolate
 horizon = "PT2H"        # how far into the future (default PT2H)
 # step = "PT5M"         # timestep spacing (default: source cadence)
+# lightning_source = "lightning"  # events-shape postgis collection: per-cell
+                                  # flash_count/rate + lightning_jump (#549)
 
 [collections.wms]
 colormap = "radar_dbz"
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -220,6 +220,13 @@ pub struct NowcastConfig {
     /// object-harness gate has passed for that source.
     #[serde(default)]
     pub growth_decay: bool,
+    /// Collection id of a point-event collection (engine-postgis `events`
+    /// shape, e.g. lightning strikes) to join onto tracked cells as
+    /// per-cell flash counts/rates and a lightning-jump flag (#549). Must
+    /// exist in the same config; a missing or non-event collection fails
+    /// this collection at load.
+    #[serde(default)]
+    pub lightning_source: Option<String>,
 }
 
 fn default_nowcast_horizon() -> String {
--- a/crates/core/src/events.rs
+++ b/crates/core/src/events.rs
@@ -1,0 +1,41 @@
+//! Point-event sources (#549): lightning strikes and similar timestamped
+//! point observations, joinable onto other engines' domain objects (e.g.
+//! nowcast cell tracks). Framework-free like the rest of ds-core — the
+//! trait is the seam that lets engine-nowcast consume engine-postgis
+//! events without depending on it.
+
+use chrono::{DateTime, Utc};
+
+use crate::error::DataServerError;
+
+/// One point event in WGS84.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct EventPoint {
+    pub time: DateTime<Utc>,
+    pub lon: f64,
+    pub lat: f64,
+}
+
+/// A source of recent point events.
+///
+/// Contract:
+/// - `recent_events` returns every event in the HALF-OPEN window
+///   `(start, end]` across the source's whole extent, ascending by time,
+///   capped at `limit` — when capped, the NEWEST events are kept (the cap
+///   is a safety valve; callers treat `len() == limit` as possible
+///   truncation and may log it).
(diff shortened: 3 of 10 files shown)


mrauhala:
All four findings addressed in 45f1eca: (1) `lightning_jump` is now null when the join was skipped (flash_count doubles as the joined-this-generation marker — unknown ≠ false); (2)+(3) both garbled multi-line strings fixed (a python-side line-continuation had eaten the Rust backslashes); (4) `lightning_source_error_degrades_to_null_fields` covers the Err arm end to end — generation completes, track persists, all three flash fields read null.

mrauhala:
Perf note addressed in 7f93ee3 as documentation: the brute-force fallback is a sized decision — worst real Nordic window ≈ 10⁴ strikes × ~150 cells = ~10⁶ distance checks once per 5-min generation on the background runtime (ms); even MAX_JOIN_STRIKES × the 255-track label ceiling stays in the tens of ms. A spatial index earns its complexity only if either bound grows by orders of magnitude — the doc on apply_lightning now says so.

mrauhala:
Coverage gap closed in 7ab7910: `nowcast_missing_lightning_source_fails_that_collection` exercises the real `load_collections` path — the collection fails with an error naming `lightning_source`, stays out of the registries, and the base collection is unaffected. A station-shape postgis id fails through the same lookup (it never enters the event-source registry), so both halves of "missing or non-events" are covered by the registry miss.

PR #553, opened 2026-07-26
fix(engine-nowcast): speed-based track gates + velocity clamp — kills 200 km/h phantom cells

mrauhala:
## Bug (client-reported, verified live)

Cells on the prod Features layer report 150–216 km/h motion while their markers visibly move a fraction of that, with bearings tens of degrees off. Live two-snapshot diff via the `?datetime=` history (00:35→00:40Z, 2026-07-26) quantified it:

- track 1663: reported **21.2 m/s @ 184°**, actual marker displacement **0.97 km → 3.2 m/s @ 255°** (6.6× too fast, 71° off) — stale EMA memory of an earlier bad displacement;
- track 1643: a **16.5 km single-interval jump** (55 m/s implied) — an association mismatch fully legal under the flat gate;
- healthy tracks: median reported/implied ratio **0.97** — the velocity math itself is correct, the associations and their EMA legacy are the problem;
- at 00:50Z, 5 live tracks over 30 m/s, topping at 182 km/h.

## Root cause

`TRACK_GATE_KM = 20` per 5 min admits matches implying **66 m/s**; the motion-compensated pass grants that same 20 km ON TOP of the ambient displacement; and one bad displacement pollutes the 0.5-EMA for generations (50→25→12.5%).

## Fix (aligned with `ds_core::cells::track_cells`)

- **Speed-based gates**: raw pass = `MAX_CELL_SPEED_MS` (30 m/s) × elapsed + 3 km jitter base (12 km @ 5 min); compensated pass = 10 m/s residual + base (6 km) — the prediction has already absorbed advection, so only a small residual is legitimate. Counter-flow rescue via the raw pass is unchanged (test updated with the new numbers, still green).
- **Velocity clamp before the EMA fold**: displacements past the physical cap (merge/split centroid shifts) fold in magnitude-clamped, direction preserved — genuinely fast storms saturate at a plausible 30 m/s ceiling instead of either reading 200 km/h or never updating.

New regression test `gates_and_clamp_bound_implied_track_velocity` pins both: a 50 m/s implied association is rejected (blob born fresh), a 33 m/s legal match folds in clamped with bearing intact. All 45 engine-nowcast tests green; fmt + clippy clean.

Expected client-visible effect: no reported speeds above ~30 m/s, bearings recover within a generation of a footprint jump, slightly more track breaks in chaotic multicell clusters (correct behavior — those were false continuations).

https://claude.ai/code/session_01Edrv5Cf4SosH4YmM9tBnEJ

### DIFF
--- a/crates/engine-nowcast/src/cells2d.rs
+++ b/crates/engine-nowcast/src/cells2d.rs
@@ -21,12 +21,21 @@ pub const CELL_THRESHOLD_DBZ: f32 = 35.0;
 /// grid — below that, 35 dBZ specks churn between generations and flood
 /// the Features layer with unmatched one-generation "cells".
 pub const CELL_MIN_AREA_PX: usize = 10;
-/// Matching gate (km) for track continuity between generations, per
-/// [`TRACK_GATE_BASE_SECS`] of elapsed time — the gate scales linearly
-/// with the actual span (a skipped generation doubles the distance a
-/// storm legitimately covers), never below one base gate.
-pub const TRACK_GATE_KM: f32 = 20.0;
-pub const TRACK_GATE_BASE_SECS: f32 = 300.0;
+/// Speed-based matching gates (like `ds_core::cells::track_cells`):
+/// `gate_km = speed × elapsed + BASE_GATE_KM`, so the implied velocity of
+/// ANY accepted match is physically bounded. The old flat 20 km / 5 min
+/// gate allowed 66 m/s implied cell speeds — gate-edge mismatches read as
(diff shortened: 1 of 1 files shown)


mrauhala:
Addressed in f479d5d: `MAX_CELL_SPEED_MS` raised to the cited 35 m/s bow-echo bound (raw gate 13.5 km @ 5 min; clamp ceiling 126 km/h — still physical). Regression test updated: 50 m/s implied still rejected, 40 m/s legal match clamps to 35 with bearing intact.

